PR #8, opened 2015-01-12
added solution and projects for compilation with mono

xplicit:
cs.sln bond solution actively uses msbuild features which do not implemented in mono, so the original solution falis to build using mono. I added simply solution (without files generation steps) with two targets "Release and Debug", which can be compiled using xbuild or monodevelop. This solution makes only libraries, it does not allow to build tests or samples.
 
Tested on Ubuntu linux 12.04 with monodevelop 5.5 and mono 3.10.0 (current stable versions from official Xamarin repository)

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ variable, e.g.:
 
 ### Windows
 
-[![Build Status](https://ci.appveyor.com/api/projects/status/github/Microsoft/bond?svg=true&passingText=C%23%20passing&pendingText=C%23%20pending&failingText=C%23%20failing)](https://ci.appveyor.com/project/sapek/bond/branch/master)
+[![Build Status](https://ci.appveyor.com/api/projects/status/github/Microsoft/bond?svg=true&branch=master&passingText=C%23%20passing&pendingText=C%23%20pending&failingText=C%23%20failing)](https://ci.appveyor.com/project/sapek/bond/branch/master)
 
 Pre-built C# binaries signed by Microsoft are available from [NuGet](https://www.nuget.org/packages/Bond.CSharp/).
 


sapek:
Committing generated files is not good. Maintenance would be very error prone. I don't know if we can build the Bond compiler from cs.sln on Mono (even on Windows it is a bit tricky), but at the minimum we should run codegen for `bond.bond` and `bond_const.bond` as part of Bond.dll project. Does build system on Mono support running external executables?

What is the testing story on Mono? Is there an equivalent of MSTest?

sapek:
Also this is minor, but could you please follow [this](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) for commit messages? E.g. "Add Mono solution and projects".

xplicit:
What do you think about this idea: have generated files in the solution, and call gbc on corresponding *.bond in pre-build step. In this case we can build solution even if gbc is not installed. If it's installed we got updates on generated files. Pre-build steps with calling external command should work with xbuild.    

NUnit has great support in mono and monodevelop.

cs/src/core/Bond.mono.csproj
```diff
@@ -0,0 +1,121 @@
...
+  <PropertyGroup>
+    <ProjectGuid>{43CBBA9B-C4BC-4E64-8733-7B72562D2E91}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>Bond</RootNamespace>
+    <AssemblyName>Bond</AssemblyName>
+    <BondRedistributable>true</BondRedistributable>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
```
sapek: If Mono is equivalent of .Net 4.5 we should add:

    <DefineConstants>$(DefineConstants);NET45</DefineConstants>

There is some conditional code (I think at the moment it is only aggressive inlining attribute on methods).

sapek:
No, we should not commit generated files. Besides the maintenance problems (optional codegen doesn't really solve that) it is important that we don't send a messages to the users that committing generated files is OK. It is not and it can lead to very hard to troubleshoot bugs. Codegen should always runs as a build step.

I should have thought about Mono when I was deciding on unit test framework. I suspect that converting MSTest to NUnit won't be a huge deal, but it would have been easier to do that from the start...

What do you think about sharing project files between Windows and Mono and handling the differences in the imported `Portable.Internal.props`? Is that feasible?

xplicit:
The idea of having generated files in the project is a quick start of using bond. If I had generated files in `samples`project, I would have being able to look into the sources and understand how to properly use bond attributes on data types and reuse this technique in my own project without any additional steps. Also .NET developer have to install haskell environment, only for getting two generated files with constants and some base .NET types. That's not an easy sometimes :) (I spent a couple of hours for looking how to update and install required versions of GCC, boost, cmake on my Ubuntu 12.04 machine). Of course, building of the libraries from sources will not be necessary when NuGet packages will be available, so this can be avoided.
To simplify the process of starting with bond I would add generated files with optional regeneration (if gbc available) to the Bond project and to Samples project. In this case devs could make quick build of the solution and use the Bond libraries in their projects. In the Test project files need to be generated every time to be sure that tests are passed with the latest versions of gbc, so it should require gbc. 

As for tests transforming MSTest to NUnit should be an easy, for example this post describes the steps http://www.florian-rappl.de/News/Page/275/convert-mstest-to-nunit

Of course if it's possible, it could be great to use the same projects for monodevelop/xbuild and  VS/msbuild. But I don't know all the limitations of xbuild in mono, sharing may work or won't work. It's need to be tested. For example, in current cs.sln monodevelop can't read build targets for the projects (while it can read them for the whole solution). I did not look thoroughly into the proj files, but think they are defined in some common .target or .props file which should be imported by xbuild, but does not. It can be fixed if import does not work because of using some special $() macros. But if the cause of this behaviour is something other, this would be difficult to create some universal projects.

About current pull request: should I make changes (define constant, change commit message, remove generated files, add pre-build step with gbc) or should not do this before import of *.props be researched?

sapek:
I would like to avoid code duplication between Mono and Windows projects as much as possible. I pulled your fork and looking at the diffs there is a lot in common but it also clear that the Windows project require some re-factoring to accommodate sharing (the obvious one is not assuming case-insensitive file system). Let me take a look at this and I will get back to you.

sapek:
OK, let's go with separate projects for Mono, but without generated files checked in. For gbc location we should use `BOND_COMPILER_PATH` variable/property (this is what we use on [bond.cs.props](https://github.com/Microsoft/bond/blob/master/cs/build/bond.cs.props#L9)) and when it is not set use relative path for the default location where gbc is built when you execute default build instructions.

xplicit:
I'll try to reuse the projects. With some luck this can be managed (I'm almost get current Attributes project to workable state with its imports. What is left: resolve or workaround monodevelop NullRef exception while project configuration is loading).
Also Xamarin released mono 3.12 few days ago with a MSBuild bug fixed https://bugzilla.xamarin.com/show_bug.cgi?id=22981 This can help with files autogenerating.

sapek:
That's great. My attempt of reusing was to move list of files to a separate .props file that was imported in actual project. That works for build but you can't see/edit list of files in Visual Studio which was a deal breaker.

xplicit:
I've spent some time to find a workarounds for build files, I did not resolve all issues yet, just want to put here workarounds  that I've found/what need to be changed to make projects be compilable. If I get some free time to resolve all the issues, I'll make a fully-functional pull request. If not then someone will be able to use these results to complete the work.  

I've tested the build with command `xbuild /p:Configuration=Release cs.sln` using mono 3.12 on Ubuntu Linux platform

1. 
`<DocumentationFile>$(TargetDir)\$(TargetName).xml</DocumentationFile>` 
`$(TargetDir)` does not work in `DocumentationFile`  as expected (not working only in DocumentationFile, in other places in works), xbuild throws an exception. As a workaround I've used `$(OutDir)\$(TargetName).xml` here
2. `<FileToCopy Include="$(TargetDir)\$(TargetName).pdb" />` xbuild generates *.mdb files, so I have to replace this string to `<FileToCopy Include="$(TargetPath).mdb" />`
3. Different Uppecasing/lowercasing in directories names/project files includes. What should be used: rename `properties` folder  to `Properties` or change the folder name in *.csproj file from `Properties' to `properties`?
4. I don't know yet what is a proper way to detect xbuild in csproj files and make a switch in includes. For now I use  `Condition = " '$(OS)' == 'Unix' "`

These changes, I've written above allow to build Attributes project without changes of solution/project files. 

Then I've stuck with `<BondCompileCs>` target. It does not work with xbuild without significant changes. What is not working:

1. `<BondCodegen Include="$(BOND_INCLUDE_PATH)\bond\core\bond_const.bond" />` in Bond.csproj. When Includes are writing in *.tmp file, `$(BOND_INCLUDE_PATH)` is not evaluating, so I have to replace it to `..\..\..\cpp\inc\`
2. Seems that xbuild can't work with ItemDefinitionGroup. (There is a pull request in mono about it https://github.com/mono/mono/pull/1484) I have to remove all references to `Options` property https://github.com/Microsoft/bond/blob/master/cs/build/Bond.CSharp.targets#L38
otherwise xbuild throws an exception.
3. Then I've run into the issue, which could not resolve yet, and don't know what is its actual reason. I've got an xbuild error 	`: error : Error building target BondCompileCs: Object reference not set to an instance of an object`

  		Error building target BondCompileCs: System.NullReferenceException: Object reference not set to an instance of an object
		  at Microsoft.Build.BuildEngine.Project.GetMetadataBatched (System.String itemName, System.String metadataName) [0x00008] in /tmp/buildd/mono-3.12.0/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/Project.cs:1389 
		  at Microsoft.Build.BuildEngine.MetadataReference.ConvertToString (Microsoft.Build.BuildEngine.Project project, ExpressionOptions options) [0x00000] in /tmp/buildd/mono-3.12.0/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/MetadataReference.cs:74 
		  at Microsoft.Build.BuildEngine.ExpressionCollection.ConvertToString (Microsoft.Build.BuildEngine.Project project, ExpressionOptions options) [0x00047] in /tmp/buildd/mono-3.12.0/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/ExpressionCollection.cs:147 
		  at Microsoft.Build.BuildEngine.ExpressionCollection.ConvertToNonArray (Microsoft.Build.BuildEngine.Project project, System.Type type, ExpressionOptions options) [0x00000] in /tmp/buildd/mono-3.12.0/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/ExpressionCollection.cs:98 
		  at Microsoft.Build.BuildEngine.ExpressionCollection.ConvertTo (Microsoft.Build.BuildEngine.Project project, System.Type type, ExpressionOptions options) [0x00051] in /tmp/buildd/mono-3.12.0/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/ExpressionCollection.cs:86 
		  at Microsoft.Build.BuildEngine.Expression.ConvertTo (Microsoft.Build.BuildEngine.Project project, System.Type type, ExpressionOptions options) [0x00000] in /tmp/buildd/mono-3.12.0/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/Expression.cs:633 
		  at Microsoft.Build.BuildEngine.BuildItem.AddMetadata (System.String name, System.String value) [0x0002c] in /tmp/buildd/mono-3.12.0/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/BuildItem.cs:273 
		  at Microsoft.Build.BuildEngine.BuildItem.Evaluate (Microsoft.Build.BuildEngine.Project project, Boolean evaluatedTo) [0x0005d] in /tmp/buildd/mono-3.12.0/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/BuildItem.cs:301 
		  at Microsoft.Build.BuildEngine.BuildTaskItem.Execute () [0x00007] in /tmp/buildd/mono-3.12.0/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/BuildTaskItem.cs:64 
		  at Microsoft.Build.BuildEngine.TaskBatchingImpl.Execute (IBuildTask buildTask, TaskExecutionMode taskExecutionMode) [0x00028] in /tmp/buildd/mono-3.12.0/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/TaskBatchingImpl.cs:104 
		  at Microsoft.Build.BuildEngine.TaskBatchingImpl.Build (IBuildTask buildTask, TaskExecutionMode taskExecutionMode, System.Boolean& executeOnErrors) [0x00020] in /tmp/buildd/mono-3.12.0/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/TaskBatchingImpl.cs:52 
		  at Microsoft.Build.BuildEngine.TargetBatchingImpl.RunTargetWithBucket (System.Collections.Generic.Dictionary`2 bucket, Microsoft.Build.BuildEngine.Target target, System.Boolean& executeOnErrors) [0x000c8] in /tmp/buildd/mono-3.12.0/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/TargetBatchingImpl.cs:140 
		  at Microsoft.Build.BuildEngine.TargetBatchingImpl.Run (Microsoft.Build.BuildEngine.Target target, System.Boolean& executeOnErrors) [0x00060] in /tmp/buildd/mono-3.12.0/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/TargetBatchingImpl.cs:85 
		  at Microsoft.Build.BuildEngine.TargetBatchingImpl.Build (Microsoft.Build.BuildEngine.Target target, System.Boolean& executeOnErrors) [0x00016] in /tmp/buildd/mono-3.12.0/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/TargetBatchingImpl.cs:64 
		  at Microsoft.Build.BuildEngine.Target.DoBuild (System.Boolean& executeOnErrors) [0x00017] in /tmp/buildd/mono-3.12.0/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/Target.cs:287

xplicit:
@sapek I think, I found workarounds for most of the xbuild issues, to make a new pull request I'd like to know what to do with case-insensitive files: change the filename case in the `include` attribute of *.csproj file or rename (uppercase) the directory which do not match the case in the *.csproj?

sapek:
We should change the names in the project, not rename directories.

PR #19, opened 2015-01-22
Make Bond compilable with mono xbuild

xplicit:
This is the first step to make the bond solution compilable with xbuild
This commit allows to build Bond libraries and some bond examples. There
are some limitations due to the reason xbuild does not support all MSBuild features.

What is not supported:
- NET 4.0 profile. xbuild builds only .NET 4.5 profile, there are compile errors when
  it tries to build .NET 4.0 target after building .NET 4.5. The issue that xbuild does
  not pass codegenerated files to mcs when build for .NET 4.0. Need more investigation
  why this happens.
- Building tests. They are using VisualStudio test libraries which is not supported by mono.
  It will work when tests will be converted to NUnit tests
- There are some compile errors in examples related to using LinkedList<double>. Need
  more investigation is this class supported by mono and if yes why there are errors.


xplicit:
I use current solution. And here is the command I use to test the build:

`xbuild /p:Configuration=Release cs.sln`

xplicit:
I'll change the name of Bond.JSON back. But in this case I have to change `InternalVisibleTo["Bond.JSON",..` is it appropriate solution?

sapek:
You are right. I missed that. Please do change that.

xplicit:
I changed back assembly name to Bond.JSON.

PR #21, opened 2015-01-27
Add CI builds for C# solution

xplicit:
Allows to build the csharp solution in Travis CI. Tested only for linux platform, because
Travis CI requires to do manual request to enable CI on multiple platforms. 
Also, beta-support of C# is enabled only for linux in Travis CI, so it may be need to do 
additional install of mono on os x platform in `before_install` step.


xplicit:
I don't know is it mandatory to run builds on OSX, I don't have ability to test osx builds, because osx is disabled in my travis account.
Also I changed `make --jobs 2 check` to `make check`, because I often get the clang compiling job be killed due to out of memory on travis machine.

xplicit:
Without `make --jobs 2 check` it timeouts.

sapek:
I build C++ project on OS X in travis because there is no Ubuntu 12.04 package for Boost 1.54 or newer. I was planning to add C++ build once travis adds support for Ubuntu 14.04 (they said they are working on it). I think it would be fine for now to have C++ build on OS X and Mono C# build on Linux.

xplicit:
@sapek I added installation of Boost 1.55 on Ubuntu 12.04 to .travis.yml. It works in general. I have issues with `make check` command on travis linux machine. It timeouts if I use `make check` command or be killed by out of memory if I use `make --jobs 2 check`.  If use `make` it works.
Also now I am waiting the build result for os x

You can look at the results:
linux build: https://travis-ci.org/Microsoft/bond/jobs/48547741
os x build (currently building): https://travis-ci.org/Microsoft/bond/jobs/48547743

xplicit:
Also I tried to use GCC 4.7 instead of clang, but

sapek:
Clang 3.6 is the fastest, that might work.

sapek:
If you want you can leave the `make check` OS X only for now and just enable Mono build in this PR. I will revisit the C++ build (you can leave the installation script part, just disable calling `make check` on Linux). C++ build with all unit tests take a long time on a VM and I might have to look at optimizing it in general to avoid Travis timeouts.

xplicit:
Did it already (`make check` for OS X and `make` for linux). I'll wait when the current OS X build ends and commit the change.

xplicit:
OS X build is passed (if don't mind about MSTest compile errors)!

xplicit:
I rebased commits and put them into one commit. Now it builds cs.sln both on linux and OS X, on OS X it also runs C++ unit-tests. 
One thing is left: need to convert MSTest to NUnit and the solution will be built without errors and C# unit-tests will be able to run. In this case the build will be passed without exceptions.

sapek:
We need to disable the test projects on Mono until we convert them to NUnit or the CI build will constantly show as broken, right?

xplicit:
@sapek Yes, you're right. I converted tests to unit in this commit https://github.com/xplicit/bond/commit/37b82ed307771246a9a9520e116d902a147be497

After conversion .NET 140 tests passed, one test is failed (it seems, that NUnit can't compare empty array segments). 

    Assert.IsEqual(new ArraySegment<byte>(), new ArraySegment<byte>()); //throws exception

On mono 46 tests are failed (something with unsafe IO, I did not look inside them yet). 

Also, I want to try to add stable version of NUnit (at first attempt I added NUnit 3.0 alpha package, because it has got different libraries for different frameworks, while v2.6 has got only one library for all targets).

To run Unit tests from VS you need to install NUnit Test Adapter (Tools/Extensions and Updates)

xplicit:
Here new pull request with the same thing: https://github.com/Microsoft/bond/pull/36

PR #23, opened 2015-01-29
Convert MSTest tests to NUnit tests

xplicit:


xplicit:
There was not enough time to run all tests. I'll try to optimize test running to speed up the process.

sapek:
@xplicit I ran the test from your branch locally and in my VS the NUnit tests take the same amount of time to run as MSTests before. We were already close to the limit on AppVeyor so maybe it was just a fluke. I kicked off another build on the PR.

sapek:
@xplicit On AppVeyor the NUnit tests run significantly slower (comparing the longest running tests in last successful build and PR build it seems to be 4x slower).

xplicit:
@sapek I have a couple of ideas how to speed up NUnit runs, I want to try them. When I get the results, I'll write them here

sapek:
@xplicit I was wondering that maybe NUnit test runner tries to execute tests from all assemblies that are passed on the command line concurrently and on CI VM which has only one virtual processor this end up taking more time than executing them serially? Maybe try call the runner for each test assembly on it own?

xplicit:
@sapek I tried several options (by the way did not try to run files one by one), but could not isolate issue and find workaround yet :( They should run the tests one by one (at least at my local machine nunit-console works this way).  That's a very strange issue for me, because I have different behaviour of test runs for nunit-console on my machine and appveyor machine. I am still trying to find the solution.

sapek:
@xplicit Another possibility to test is that the problem is in the custom AppVeyor logger. Try running it w/o it. We would loose tests reporting but that is not such a big deal because the result is still in the console output.

sapek:
@xplicit Separately from working with @CharliePoole on figuring out why nunit-console is so slow running Bond tests, would it be an option to run the tests in appveyor using nulite? Or is it not available as an executable anywhere?

xplicit:
It's possible to run nunitlite with tests, the package is available on nuget. It's easy to use it, what is needed just to add new console project and reference NUnitLite nuget package. 
  
One thing, that current version of nunitlite on nuget has two limitations compared to nunit-console:
- NUnitLite can run only one test dll at once, so it's needed to split the test command on the several commands.
- NUnitLite loads assembly by their assembly name, not by filename. This requires one additional step of coping all testing assemblies (or nunitlite exe file with referenced assemblies) to one folder and run it from there.

The last thing is already fixed in master branch of NUnit, but recent NUnitLite package has not been uploaded to nuget yet. I wait new version of NUnitLite on nuget (hope to see it soon) or resolving the issue with long running of nunit-console - but the last one is not an easy thing (even to detect what is the cause of the issue). 
New version of NUnitLite would protect the test script from the noise like `copy path/to/test/dlls somefolder` commands, but it's possible even now to create the test script with friendly output of the test results to the console and creating tests artifacts on appveyor.

sapek:
Specifying each test assembly is not a big deal. However having to add a command line tool for NUnitLite is a bigger problem, that pollutes the repository, not just AppVeyor config. @CharliePoole, could NUnit ship with a console app for NUnitLite?

CharliePoole:
@sapek The concept of NUnitLite is that your test assembly is a console app and references both nunit.framework and nunitlite. Your Main calls the text-based UI in nunitlite and nunitlite executes the tests in the calling assembly. There's no "runner" in the sense you mean.

If you use nuget to add nunitlite to the project, an appropriate main will be added. If not, it's easy enough to create manually following the info in the readme.

The notion of loading an assembly by name is a remnant of an older version of nunitlite. In the current code, it also works by file name. However, we haven't yet decided whether to make nunitlite a full-fledged runner that loads files and runs tests or to leave it as just a library used by self-executing tests, which is what it is right now.

[Note that NUnitLite runs on platforms where there is no such a thing as a console. I've tailored my explanation here to running under desktop .NET or mono.]

xplicit:
@sapek I have a look into NUnitLite once more and can say that it's possible to get an exe file without creating additional projects in a bond solution.  It needs to change appveyor script file only. This could be used as a temporary solution while nunit-console will be fixed for slow execution on appveyor.

sapek:
I consolidated the unit tests in @40f3888 which should give us [a little breathing](https://ci.appveyor.com/project/sapek/bond) room in appveyor CI (after switching to open source plan build was already timing out if the cached items expired and had to be rebuilt). If you can use the nunitlite w/o adding a project that would work.

xplicit:
I have found a resolution for long-running tests in nunit-console on appveyor, now I'm checking the build. If all is OK, I'll make new PR

xplicit:
@sapek Here are new PR https://github.com/Microsoft/bond/pull/34

NUnit tests now have the similar execution time as MSTest.

CharliePoole:
Looking forward to it!

xplicit:
When I tried to compile nunitlite.exe with csc, I've noticed that compiled with csc version executes slow as nunit-console on appveyor.  What was interested  that msbuild - compiled works much faster on appveyor. Investigating the command line I've found a parameter which increases performance a lot. It was `/platform:anycpu32bitpreferred`. I tried to switch `nunit-console` in bond project to `nunit-console-x86` (or `nunit-console.exe --x86` for version 3.0) and it worked with near the same speed as nunitlite on appveyor.

CharliePoole:
Have any of the previously reported results been from versions of NUnit you
built yourself? That's a potential difference we should probably explore.

sapek:
Merged these changes in #35. Thanks.

PR #25, opened 2015-01-31
Remove possible underflow in string buffer encoder

dnmiller:
Found an issue with possible underflow in functions ReadStringData and WriteStringData.


dnmiller:
I should add, platforms with non-standard size types.

PR #27, opened 2015-02-03
Fix clone of MemoryStream on Mono

xplicit:
Mono MemoryStream internal implementation differs from .NET implementation
It does not have `InternalGetOriginAndLength` method. So the cloning
method changed to use mono implementation of MemoryStream
when the method `InternalGetOriginAndLength` is not found

This commit fixes `NullReferenceException` when running unit tests on mono


xplicit:
@sapek Done.

PR #29, opened 2015-02-06
Add support for paths with quotes in C# MSBuild targets

chwarr:
This pull request
* adds support for import directories with spaces in them
* adds an example using paths with spaces
* adds an example of using import

I can split the new examples into different pull request(s) if you want. I can also remove one or both of them. Let me know what you think.


sapek:
I don't think we need two examples, and I don't think we need example with spaces for that matter. From user perspective there is no difference. If anything, spaces should be used in a unit tests.

examples/cs/core/import/schema.bond
```diff
@@ -0,0 +1,9 @@
+﻿import "common.bond"
```
sapek: It would be good to illustrate import with partial path.

chwarr:
Split the import example into its [own pr](https://github.com/Microsoft/bond/issues/30) and added a unit test. Removed the spaces example; you're right about the two examples being basically the same.

There's a really simple test in serialization that references the type from the .bond file with the spaces. I did this so that future changes would be less likely to break the files with spaces, as they would otherwise only be part of compilation, but never referenced in code. Let me know if there's a better place for this test--different file, its own file, different test directory.

Also, rebased atop the recent changes in master.

chwarr:
Wrong button. :-)

chwarr:
The AppVeyor failure is because I accidentally clicked the "Close and comment" button. It detected that the pr had been closed, so it abandoned the build.

PR #32, opened 2015-02-07
Replace dynamic exception specifications with portable boost config specifier

dnmiller:
This PR replaces the dynamic exception specifications that are deprecated in newer versions of the C++ standard with a macro that's configured by boost. 

This should hopefully preserve the optimizations in boost::variant, though I can't confirm first-hand if that's the case.


sapek:
Thanks. I will verify that the performance is not affected.
Please limit the commit title to 50 or so characters (e.g. `Replace throw() with noexcept`). Otherwise Git truncates it many contexts. Longer description can be in the rest of the message.

dnmiller:
Commit message updated

PR #34, opened 2015-02-11
Convert unit tests from MSTest to NUnit

xplicit:
Resolved an issue with tests long run under nunit-console on appveyor.


xplicit:
The time of running tests is roughly 100 seconds per each framework, so both .NET 4.5 and .NET 4.0 tests are executed less than 5 minutes.

sapek:
Cool. Could you squash into one commit?

xplicit:
All in one commit https://github.com/Microsoft/bond/pull/35

PR #36, opened 2015-02-12
Add building cs.sln and running unit tests in Travis.CI

xplicit:


xplicit:
This is updated version of previous pull request. Maybe cabal sandbox should be cached too (like it's done on AppVeyor) to improve build time (I'm not familiar with Haskell and do not know how to do it properly).

Here are two workers:
OS X worker: build cpp solution, run cpp unit tests, build cs solution.
Linux (Ubuntu 12.04) worker:  build cpp solution, build cs solution, run cs unit tests.

PR #37, opened 2015-02-16
Warning cleanup

qnikst:
This is a set of commits that fixes most of the warnings generated when code is compiled with `-Wall`.
Only unfixed warnings are the ones that appear because of `RecordWildCards` extension usage. 
After the fix build output is more sane and it's easier to reason about the code and not miss real warning.

Commit is not squashed as it seems it may be easier to review it that way.


compiler/Bond/Template/Cpp/Types_cpp.hs
```diff
@@ -27,7 +28,7 @@ types_cpp cpp file imports declarations = ("_types.cpp", [lt|
         if null declParams then CPP.schemaMetadata cpp s else mempty
 
     -- global variables for enum name/value conversions
-    statics e@Enum {..} = [lt|
+    statics   Enum {..} = [lt|
```
sapek: No need to add the white space.

compiler/Bond/Template/Cpp/Types_h.hs
```diff
@@ -237,6 +241,7 @@ namespace std
             allocInitValue i f d = i f d
             keyType (BT_Set key) = cppType key
             keyType (BT_Map key _) = cppType key
+            keyType _            = error "keyType: impossible happened."
```
sapek: The added white space doesn't really align definitions.

compiler/Bond/Template/CustomMapping.hs
```diff
@@ -30,11 +31,18 @@ data NamespaceMapping = NamespaceMapping
     , toNamespace :: QualifiedName
     }
 
+
```
sapek: How about:

    type Parser a = Parsec SourceName () a

    whitespace :: Parser String
    -- etc

Then we don't need to import `Data.Monad.Identity`

compiler/Bond/Template/CustomMapping.hs
```diff
@@ -42,9 +50,9 @@ integer = decimal <$> many1 digit <?> "decimal number"
 -- parse alias mapping specification from the command line --using flags
 -- e.g.: --using="OrderedSet=SortedSet<{0}>"
 parseAliasMapping :: [String] -> IO [AliasMapping]
-parseAliasMapping = mapM parseAliasMapping
+parseAliasMapping = mapM parseAliasMapping'
```
sapek: Maybe better names would be:

    parseAliasMappings = mapM parseAliasMapping

?

compiler/Bond/Template/CustomMapping.hs
```diff
@@ -56,9 +64,9 @@ parseAliasMapping = mapM parseAliasMapping
 -- parse namespace mapping specification from the command line --namespace flags
 -- e.g.: --namespace="bond="
 parseNamespaceMapping :: [String] -> IO [NamespaceMapping]
-parseNamespaceMapping = mapM parseNamespaceMapping
```
sapek: Maybe `parseNamespaceMappings = mapM parseNamespaceMapping`?

compiler/Bond/Template/TypeMapping.hs
```diff
@@ -158,6 +170,7 @@ infixr 6 <<>
 pureText :: ToText a => a -> TypeNameBuilder
 pureText = pure . toText
 
+commaSepTypeNames :: [Type] -> Reader Context Builder
```
sapek: -> TypeNameBuilder

compiler/Bond/Template/TypeMapping.hs
```diff
@@ -248,23 +262,27 @@ cppType (BT_UserDefined a@Alias {..} args) = aliasTypeName a args
 cppType (BT_UserDefined decl args) = declQualifiedTypeName decl <<>> (angles <$> commaSepTypeNames args)
 
 -- C++ type mapping with custom allocator
+cppTypeCustomAlloc :: Builder -> Type -> Reader Context Builder
```
sapek: -> TypeNameBuilder

compiler/Bond/Template/TypeMapping.hs
```diff
@@ -248,23 +262,27 @@ cppType (BT_UserDefined a@Alias {..} args) = aliasTypeName a args
...
 cppTypeCustomAlloc alloc (BT_List element) = "std::list<" <>> elementTypeName element <<>> ", " <>> allocator alloc element <<> ">"
 cppTypeCustomAlloc alloc (BT_Nullable element) | structType element = "bond::nullable<" <>> elementTypeName element <<> ", " <> alloc <> ">"
-cppTypeCustomAlloc alloc (BT_Nullable element) = "bond::nullable<" <>> elementTypeName element <<> ">"
+cppTypeCustomAlloc _lloc (BT_Nullable element) = "bond::nullable<" <>> elementTypeName element <<> ">"
 cppTypeCustomAlloc alloc (BT_Vector element) = "std::vector<" <>> elementTypeName element <<>> ", " <>> allocator alloc element <<> ">"
 cppTypeCustomAlloc alloc (BT_Set element) = "std::set<" <>> elementTypeName element <<>> comparer element <<>> allocator alloc element <<> ">"
 cppTypeCustomAlloc alloc (BT_Map key value) = "std::map<" <>> elementTypeName key <<>> ", " <>> elementTypeName value <<>> comparer key <<>> pairAllocator alloc key value <<> ">"
 cppTypeCustomAlloc _ t = cppType t
 
+comparer :: Type -> Reader Context Builder
```
sapek: -> TypeNameBuilder

compiler/Bond/Template/TypeMapping.hs
```diff
@@ -248,23 +262,27 @@ cppType (BT_UserDefined a@Alias {..} args) = aliasTypeName a args
...
 comparer t = ", std::less<" <>> elementTypeName t <<> ">, "
 
+allocator :: Builder -> Type -> Reader Context Builder
```
sapek: -> TypeNameBuilder

compiler/Bond/Template/TypeMapping.hs
```diff
@@ -248,23 +262,27 @@ cppType (BT_UserDefined a@Alias {..} args) = aliasTypeName a args
...
 cppTypeCustomAlloc _ t = cppType t
 
+comparer :: Type -> Reader Context Builder
 comparer t = ", std::less<" <>> elementTypeName t <<> ">, "
 
+allocator :: Builder -> Type -> Reader Context Builder 
 allocator alloc element =
     "typename " <>> alloc <>> "::rebind<" <>> elementTypeName element <<> ">::other"
 
+pairAllocator :: Builder -> Type -> Type -> Reader Context Builder
```
sapek: -> TypeNameBuilder

compiler/Bond/Parser.hs
```diff
@@ -88,6 +89,7 @@ declaration = do
     updateSymbols decl <?> "declaration"
     return decl
 
+updateSymbols :: Monad m => Declaration -> ParsecT s Symbols m ()
```
sapek: `updateSymbols :: Declaration -> Parser ()`?

compiler/Bond/Parser.hs
```diff
@@ -204,14 +208,17 @@ struct = do
...
+        case findDuplicates fields' of
+            [] -> return fields'
             Field {..}:_ -> fail $ "Duplicate definition of the field with ordinal " ++ show fieldOrdinal
       where
         findDuplicates xs = deleteFirstsBy ordinal xs (nubBy ordinal xs)
         ordinal = (==) `on` fieldOrdinal
 
+manySortedBy :: (a -> a -> Ordering)
+             -> ParsecT s u m a
+            -> ParsecT s u m [a]
```
sapek: manySortedBy :: (a -> a -> Ordering) -> Parser a -> Parser [a] 

the function is more generic but we don't export it and only use with Bond parser.

compiler/Bond/Template/Cpp/Types_h.hs
sapek: `allocatorCtor/keyType`

compiler/Bond/Template/Cs/Util.hs
```diff
@@ -107,8 +118,9 @@ defaultValue cs Field {fieldDefault = (Just def), ..} = explicitDefault def
         explicitDefault (DefaultInteger x) = Just [lt|#{x}|]
         explicitDefault (DefaultFloat x) = Just $ floatLiteral fieldType x
             where
-                floatLiteral BT_Float x = [lt|#{x}F|]
-                floatLiteral BT_Double x = [lt|#{x}|]
+                floatLiteral BT_Float y = [lt|#{y}F|]
+                floatLiteral BT_Double y = [lt|#{y}|]
+                floatLiteral _         _ = error "floatLiteral: impossible happened."
```
sapek: `defaultValue\floatLiteral`

sapek:
Sorry for the delay. I was traveling. today I finally had time to review the changes. I added a few comments. 

As for commits for merge, I think it should be either a single commit or two:

    Fix warnings in compiler
    Write correct signature for template functions

(for consistency, please don't period at the end of commit title)

qnikst:
@sapek thanks for reply! I'll address all your comments and questions later today (in 8-12hrs, I suppose) and force push changes to this PR.

qnikst:
> The TemplateBuilder alias (although I would call it CodegenTemplate) seems to be only useful in Main.hs, to declare the codeGen function type nicely. Using it in the actual template definitions actually obscures information. Wouldn't it be better to use the types of the parameters there? E.g.:

> enum_h :: Context -> String -> [Import] -> [Declaration] -> (String, L.Text)
> enum_h cpp _file _imports declarations = 

I'm ok with ether remove type alias entirely or use it in common part of the template definitions. As you mentioned that it obscures information, then I prefer to remove it.

qnikst:
> Another related thing: if we export Context explicitly we probably should rename it to MappingContext and get rid of newMappingContext which doesn't add any value when Context is exported. What do you think?

I think that `newMappingContext` make sense only in case if it's smart constructor currently it's easier to export `Context`, if you have plans to make it smart constructor as one of the short term goals then there is no need in such change. Otherwise renaming Context to MappingContext and exporting constructor is better (I think I'll add this change as a separate commit)

compiler/Bond/Template/CustomMapping.hs
qnikst: Yes of cause, I mentioned that I can use  simple `Parser` type-synonym only after this change.

compiler/Bond/Template/CustomMapping.hs
qnikst: I really don't like such name shadowing, basically for 2 reasons:
1. it introduces a warning that can refer to a real error
2. it looks a bit complicated and forces user to look at the context as `foo = mapM foo` it looks like some scary thing with infinite type :).

I'd use `go` or worker or just inlined it as:

```haskell
parseAliasMapping = mapM (\s -> either (fail.show) return (parse aliasMapping s s)
```

Not sure what solution will require less brain cycles from the person who reads it..

qnikst:
I've addressed all warnings and force pushed squashed patches.

sapek:
I didn't notice before that the first commit added one version of template functions signatures and the second changed them. To avoid this unnecessary history I squashed those two commits and merged the PR manually. 

Thanks for contributing.

PR #46, opened 2015-03-31
Haskell support for Bond

rblaze:
It is a little bit strange to have bond compiler written in haskell and not able to generate haskell code. There, I fixed it. It is not yet equal to C++ and C# versions, but can handle most simple cases.

This patch adds "gbc haskell" mode, which behaves just like any other compiler mode, and runtime library with support for Simple, Compact and Fast protocols. Test suite successfully runs on compat data.

For now only compile-time schemas are supported, no runtime schemas yet. Mutually recursive structures also not yet supported.

Unfortunately, I'm not familiar with cmake, so here is build and test instructions:
cd haskell/runtime
cabal sandbox init
cabal sandbox add-source ../../compiler
cabal install gbc
cabal exec gbc -- haskell -o generated ../../test/compat/schemas/*.bond
cabal configure
cabal build
cabal test


rblaze:
Messed up with squash, will resubmit.

PR #47, opened 2015-03-31
haskell support

rblaze:
It is a little bit strange to have bond compiler written in haskell and not able to generate haskell code. There, I fixed it. It is not yet equal to C++ and C# versions, but can handle most simple cases.

This patch adds "gbc haskell" mode, which behaves just like any other compiler mode, and runtime library with support for Simple, Compact and Fast protocols. Test suite successfully runs on compat data.

For now only compile-time schemas are supported, no runtime schemas yet. Mutually recursive structures also not yet supported.

Runtime library is build by cmake but not installed anywhere: I still to figure out how to install haskell libraries without cabal.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,7 @@ include (PythonTest)
 add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} -C ${CMAKE_CFG_INTDIR} --output-on-failure)
 
 add_subdirectory (compiler)
+add_subdirectory (haskell/runtime)
 add_subdirectory (cpp)
 add_python_subdirectory (python)
 add_subdirectory (examples)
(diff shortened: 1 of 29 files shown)


sapek:
Thanks!
This will probably take some time to chew on (I think we may start a feature branch for the work). In the meantime, could I ask you to sign [Contribution License Agreement](https://cla.microsoft.com/)? 

Also, for consistency, we are trying to follow [this convention](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) for commit messages.

rblaze:
I find it hilarious to sign CLA in the name of my employer, which happens to be same Microsoft I'm contributing to. Anyway, done.

sapek:
You are right, CLA is not needed if you are a MSFT employee. That wasn't clear though.

rblaze:
Travis linux builds are now working. OS X build fails because of missing apt-key even before source code is checked out.
I wonder what to do with AppVeyour: haskell-src-exts fails build because it's memory hungry, it takes about 1.5G on my machine and ghc is limited to 1/10th of it.

rblaze:
Starting new request with squashed commits.

PR #49, opened 2015-04-07
Adds title and fixes typo on Why Bond? page

soundslocke:
Adds page title as "Why Bond?" and fixes typo "transform" to "transforms"


sapek:
I merged it as a quick fix but it should also be corrected in the [source file](https://github.com/Microsoft/bond/blob/master/doc/src/why_bond.md).

BTW, in case you or someone else wanted to build the documentation, it is converted from markdown source to HTML using [Pandoc](http://johnmacfarlane.net/pandoc/). If Pandoc is installed then the make files will have additional target `documentation`.

PR #53, opened 2015-04-10
Fix Bond.CSharp.targets to support BondOptions

ranrav:
A bug in the condition for the _BondCodegenWithDefaultOptions items was
causing a separate instance of gbc.exe to be launched per BondCodeGen
item when BondOptions wasn't empty.


sapek:
This change breaks Linux/Mono build. Maybe the alternative of not adding global option to BondCodegen options would be better?

ranrav:
I couldn't find a build log to see what fails.  Did you understand what didn't work?

Sent from my Windows Phone

sapek:
https://travis-ci.org/Microsoft/bond/jobs/58008885

PR #57, opened 2015-04-22
Update README to suggest Chocolatey

JayBazuzi:


sapek:
Applied manually with 1c9e96f.

PR #58, opened 2015-04-22
README link to Chocolatey

JayBazuzi:


sapek:
Looks like you got GitHub confused by adding another PR for the same file and it can't merge it automatically. Maybe it will be easier for me to just apply you change to README myself, but if you want to fix your PR my suggestion is to close this new PR and amend the commit on the original one. BTW, commit message should be in imperative, like the original one.

sapek:
Applied manually with 1c9e96f.

PR #61, opened 2015-05-06
Add Python3 Support

alfpark:


cmake/Config.cmake
```diff
@@ -19,8 +19,21 @@ if (WIN32)
             "${CMAKE_CURRENT_SOURCE_DIR}/cs/test/compat/bin/retail")
 endif()
 
-find_package (PythonLibs 2.7)
-find_package (PythonInterp 2.7)
+# set python version
+if (NOT DEFINED BOND_PYTHON_VERSION)
```
sapek: `FindPython.cmake` defines `Python_ADDITIONAL_VERSIONS` variable. Could we use it to control user preferred version instead of defining a Bond specific variable?

cmake/Config.cmake
```diff
@@ -28,7 +41,27 @@ find_package (Boost 1.53.0
         thread
         system
         unit_test_framework
-        python)
+        ${_BOOSTPYLIB})
+
+# if _BOOSTPYLIB was not found for non-windows platforms,
+# try searching for built-from-source defaults
```
sapek: Does the naming convention not apply to built-from-source for windows?

python/inc/bond/python/converters.h
```diff
@@ -303,26 +307,51 @@ struct rvalue_set_container_from_python
...
+#endif
     }
 
     static void extract(bond::blob& dst, const boost::python::object& src)
     {
+#if PY_VERSION_HEX >= 0x03000000
+        boost::shared_ptr<void> hold_convertible_ref_count(
+            (void*)0,
+            boost::python::converter::shared_ptr_deleter(
+              boost::python::handle<>(boost::python::borrowed(src.ptr()))));
```
sapek: indentation

python/inc/bond/python/converters.h
```diff
@@ -406,8 +435,10 @@ void register_builtin_convereters()
             true
             >();
 
+#if PY_VERSION_HEX < 0x03000000
```
sapek: This probably can be left in Py3, it doesn't cost anything.

cmake/Config.cmake
alfpark: Fixed per suggestion.

cmake/Config.cmake
alfpark: Yes, you're correct. Fixed.

cmake/Config.cmake
```diff
@@ -19,8 +19,24 @@ if (WIN32)
             "${CMAKE_CURRENT_SOURCE_DIR}/cs/test/compat/bin/retail")
 endif()
 
-find_package (PythonLibs 2.7)
-find_package (PythonInterp 2.7)
+# find python interpreter and library
+# to specify a different version, invoke cmake with:
+# -DPython_ADDITIONAL_VERSIONS=Major.Minor
+find_package (PythonInterp ${Python_ADDITIONAL_VERSIONS})
```
sapek: I think you don't need to/shouldn't specify `${Python_ADDITIONAL_VERSIONS}` explicitly. It is added by `find_package` to the front of the list of versions to consider. I *think* we might still need to specify 2.7 as minimum version though.

alfpark: I tried to follow the cmake recommendation to not use any version at all, but did not work (at least on Ubuntu 14.04). Result matrix with the header corresponding to values in the `find_package(PythonInterp` **header value**`)`:

 | Blank | 2.7 | `${Python_ADDITIONAL_VERSIONS}` |
|---|---|---|---|---|
`cmake` | py27 | py27 | py27  |   |   |
`cmake -DPython_ADDITIONAL_VERSIONS=2.7` | py27 | py27 | py27 |   |   |
`cmake -DPython_ADDITIONAL_VERSIONS=3.4` | interp=py27, pylib=not found, boostpylib=py27 | interp=py27, pylib=not found, boostpylib=py27 | py34  |   |   |

alfpark: After fiddling with some of the options for cmake, I removed the tag altogether and this appears to be the proper way to configure python2 vs python3:
- Python2: `cmake`
- Python3: `cmake -DPYTHON_EXECUTABLE=/usr/bin/python3`

For any versions that cmake has no implicit built-in support, then the user can specify `Python_ADDITIONAL_VERSIONS`

cmake/Config.cmake
```diff
@@ -28,7 +47,27 @@ find_package (Boost 1.53.0
...
+        OPTIONAL_COMPONENTS
+            date_time
+            thread
+            system
+            unit_test_framework
+            ${_BOOSTPYLIB})
+endif()
+# remap boost python vars
+set(Boost_PYTHON_FOUND ${Boost_${_BOOSTPYLIB}_FOUND})
+set(Boost_PYTHON_LIBRARY ${Boost_${_BOOSTPYLIB}_FOUND})
```
sapek: This breaks `add_python_subdirectory` macro defined in `PythonTest.cmake` and as the result the Python tests/examples are not included in the build. It should be:

    set(Boost_PYTHON_FOUND Boost_${_BOOSTPYLIB}_FOUND)
    set(Boost_PYTHON_LIBRARY Boost_${_BOOSTPYLIB}_FOUND)

cmake/Config.cmake
```diff
@@ -28,7 +47,27 @@ find_package (Boost 1.53.0
...
+    find_package (Boost 1.53.0
+        OPTIONAL_COMPONENTS
+            date_time
+            thread
+            system
+            unit_test_framework
+            ${_BOOSTPYLIB})
+endif()
+# remap boost python vars
+set(Boost_PYTHON_FOUND ${Boost_${_BOOSTPYLIB}_FOUND})
```
sapek: This breaks add_python_subdirectory macro defined in PythonTest.cmake and as the result the Python tests/examples are not included in the build. It should be:

set(Boost_PYTHON_FOUND Boost_${_BOOSTPYLIB}_FOUND)

python/inc/bond/python/converters.h
```diff
@@ -181,12 +181,16 @@ extend_map(Map& map, const boost::python::object& obj)
...
 
     BOOST_ASSERT(PyDict_Check(obj.ptr()));
-    
+
     dict dict(obj);
-        
+
     BOOST_FOREACH(object k,
         std::make_pair(
+#if PY_VERSION_HEX >= 0x03000000
```
sapek: This doesn't compile under Visual Studio 2013. `BOOST_FOREACH` is a macro and using preprocessor directives within macro arguments has an undefined behavior.

I'd suggest moving the conditional code outside of `BOOST_FOREACH`:

~~~
const auto keys = 
#if PY_VERSION_HEX >= 0x03000000
            dict.keys();
#else
            dict.iterkeys();
#endif 
~~~

cmake/Config.cmake
sapek: This doesn't work, Python examples/unit tests are not built (tested with cmake 3.0.2 and Python 2.7). 

First of all the second `set` has a clear bug, it should not use `_FOUND` variable. But there are also other subtleties of cmake (`_LIBRARY` variable can have multiple values) and what looks to me like bug in cmake (the first `set` should work but doesn't). Here's a version that seems more robust and I verified it in the above environment:

```
if (Boost_${_BOOSTPYLIB}_FOUND)
    set (Boost_PYTHON_FOUND ON)
    set (Boost_PYTHON_LIBRARY "${Boost_${_BOOSTPYLIB}_LIBRARY}")
endif() 
```

Can you please check if that also works on Linux and Python 3?

python/inc/bond/python/converters.h
jasonzio: Create a reference to the one you want, not a copy.
#if PY_VERSION_HEX >= 0x03000000
const auto & keys = dict.keys();
#else
const auto & keys = dict.iterkeys();
#endif

cmake/Config.cmake
alfpark: Good catch. Your suggestion still did not work, at least under Linux. The issue, it appears, is that `_BOOSTPYLIB` is lowercase, while the dynamic variable is uppercase. I've tested the cmake on both Python2 and Python3 under Ubuntu 14.04 and both build full python and unit tests.

python/inc/bond/python/converters.h
sapek: These method return `object` by value so we can't take a reference.

cmake/Config.cmake
```diff
@@ -19,16 +19,36 @@ if (WIN32)
...
+# -DPYTHON_LIBRARY=/path/to/libpython.so
+# -DBoost_PYTHON_LIBRARY=/path/to/libboost-python.so
+# and optionally with:
+# -DPython_ADDITIONAL_VERSIONS=Major.Minor
+# if your python version is not implicitly supported by cmake
 find_package (PythonInterp 2.7)
+find_package (PythonLibs 2.7)
+
+# to modify boost python auto-detection for your system, modify
+# "python" below to "pythonlibname" in libboost<pythonlibname>.so
```
sapek: It won't work properly with library named anything other than `python` because the makefiles use `Boost_PYTHON_LIBRARY/FOUND` variables. I'd just remove this comment. The comment above describes how to use different library/version of Python.

cmake/Config.cmake
```diff
@@ -19,16 +19,36 @@ if (WIN32)
...
+if (NOT EXISTS ${Boost_PYTHON_LIBRARY})
```
sapek: I'd reverse this to be:

```
if (Boost_PYTHON_LIBRARY)
    set (Boost_PYTHON_FOUND ON)
else() 
    find_package (Boost 1.53.0
        OPTIONAL_COMPONENTS
         python)
endif()
```

cmake/Config.cmake
```diff
@@ -19,16 +19,36 @@ if (WIN32)
...
 find_package (Boost 1.53.0
     OPTIONAL_COMPONENTS
         date_time
         thread
         system
-        unit_test_framework
-        python)
+        unit_test_framework)
+
+message(STATUS "Boost Python Found: ${Boost_PYTHON_FOUND}")
```
sapek: Please remove this one, the message with Boost_PYTHON_LIBRARY should be enough.

PR #65, opened 2015-06-23
Update bond.runtime.csharp.nuspec

Tuatan:
Removed unnecessary dependencies on Newtonsoft.JSON


sapek:
`Bond.JSON.dll` does depend on `Newtonsoft.JSON`.

Tuatan:
My mistake, I have been looking through Bond C# assemblies using dotPeek trying to minimize dependencies, and JetBrains' tool didn't show those.

Thanks Adam.

PR #66, opened 2015-06-29
Remove gcc warnings

oliora:
Eliminate GCC warnings, mostly about unknown pragma-s.

### DIFF
--- a/cpp/inc/bond/core/apply.h
+++ b/cpp/inc/bond/core/apply.h
@@ -3,8 +3,12 @@
 
 #pragma once
 
+#include "config.h"
+
+#if defined(BOOST_MSVC)
 #pragma warning(push)
 #pragma warning(disable: 4702)
+#endif
 
 #include "bonded.h"
 #include "bonded_void.h"
@@ -78,4 +82,6 @@ Apply(const Transform& transform, const T& value)
 
 } // namespace bond
 
+#if defined(BOOST_MSVC)
 #pragma warning(pop)
+#endif
--- a/cpp/inc/bond/core/bonded.h
+++ b/cpp/inc/bond/core/bonded.h
@@ -10,6 +10,7 @@
 #include "detail/double_pass.h"
 #include "detail/marshaled_bonded.h"
 
+
 namespace bond
 {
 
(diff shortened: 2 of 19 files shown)


sapek:
These warning don't show in our build. Do we need to some additional flags in the makefiles? It would be good to catch regressions in the future.

oliora:
@sapek on Linux we build with `-Wall` option.

sapek:
Does everything build cleanly with `-Wall` with this PR? If yes, could you please add `-Wall` to the makefiles? If the whole tree builds cleanly then you can add it to [Config.cmake](https://github.com/Microsoft/bond/blob/9692499b14381ef30e3159a1b8699079f4f74843/cmake/Config.cmake#L48). If there are still warnings in unit tests (possible) then we could at least enable `-Wall` for the core libraries by adding the following to [cpp/CMakeLists.txt](https://github.com/Microsoft/bond/blob/master/cpp/CMakeLists.txt):

    cxx_target_compile_options (GNU bond PRIVATE -Wall)
    cxx_target_compile_options (GNU bond_apply PRIVATE -Wall)

I also don't think that `BOOST_MSVC` adds any value over `_MSC_VER`. I would suggest replacing:

    #include "config.h"
    #if defined(BOOST_MSVC)

with

    #ifdef _MSC_VER

oliora:
I'll try to run local build with ``-Wall`` and also with ``-Wpedantic`` (just in case).

The reason why I used ``BOOST_MSVC`` instead of ``_MSC_VER`` is that according to Boost, some other vendors (Metrowerks, for example) also define ``_MSC_VER``. Please check comment after ``#elif defined _MSC_VER`` in http://www.boost.org/doc/libs/1_58_0/boost/config/select_compiler_config.hpp
I'm not sure is this still valid case and do such compilers need to be supported but to be on a safe side, I'm relying on ``BOOST_MSVC``. If you still prefer ``_MSC_VER``, I'll change it of course.

sapek:
With compilers pretending to be MSVC it could go either way: they might have the same warnings and implement MSVC pragmas or not. Until we have someone with such a compiler to verify I would keep it simple and use `_MSC_VER`.

sapek:
Resolved with 2abd39faa54206f8c0c26ae7dfdee6731edd870c.

PR #87, opened 2015-10-01
Update to Newtonsoft.Json 7.0.1

ianitor:
Hello everybody,

we have updated bond to Newtonsoft.Json 7.0.1. It gets more and more difficult to remain on 6.x especially when several libraries (with dependency Newtonsoft.Json 7.x or 6x) are used and have to be supported with several executables.

### DIFF
--- a/cs/src/json/JSON.csproj
+++ b/cs/src/json/JSON.csproj
@@ -27,10 +27,10 @@
       <HintPath>..\attributes\$(OutputPath)\Bond.Attributes.dll</HintPath>
     </Reference>
     <Reference Include="Newtonsoft.Json" Condition="'$(TargetFrameworkVersion)' == 'v4.5'">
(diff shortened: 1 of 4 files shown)


sapek:
Won't that create a problem for those who need to use 6.x? Is 7.x 100% backward compatible?

ianitor:
It is still possible to use 6.x version by redirecting assembly version as far no new features of 7.x are used. But the referred possibility should be an upgrade to 7.x.
We have tested the modifications by comparing the output between 6.x and 7.x. with the result that the output is exactly the same.

sapek:
Is it possible to redirect assembly to to 7.x instead when needed?

ianitor:
I don't understand the question.

sapek:
You said that with reference to 7.x one could use assembly redirect to use 6.5. Can't you do the redirect to 7.x now instead of changing the reference?

ianitor:
It is more complex, because we deliver our services as libraries (*.dll).

A redirect is only possible at executables, but this executables are developed by customers.

sapek:
Fixed with e362cc0306c994a3d0fdb46d7ce2da78645002ee

PR #93, opened 2015-11-04
Support for unsafe pointer based InputBuffer and OutputBuffer

### LINKED ISSUE #92: Support for unsafe pointer based InputBuffer and OutputBuffer
Currently Bond is missing a facility to deal with unmanaged memory as source and/or destination of the data. From what I see on the code implementing an InputPtrBuffer and OutputPtrBuffer would be straightforward. Is that the proper approach?

Would you accept such a contribution if possible? (I have a CoreCLR CLA already signed).

redknightlois:
The InputPtrBuffer and OutputPtrBuffer will not have the ability to grow because they do not have ownership of the memory involved.

Added an optimized memory copy implementation that it's even faster than BlockCopy, System.Buffer (.Net 4.6) and looping for small values. That’s 4,8 Gb/sec against 3,6 Gb/sec (fastest in System.Buffer) in Bulk Copy mode in the tested environment.
https://gist.github.com/redknightlois/149fdd3479cbd074f115

I haven't implemented an optimized ReadBytes(). That would require some internal changes that I don't think I can do on my own without further discussion. 

Fixes #92


cs/test/core/Util.cs
```diff
@@ -271,6 +273,22 @@ public static ArraySegment<byte> SerializeUnsafeCB<T>(T obj)
             return output.Data;
         }
 
+        public static ArraySegment<byte> SerializeUnsafeCBWithPtr<T>(T obj)
```
sapek: Wouldn't it be better if the `XxxWithPtr` methods accepted/returned `byte*` rather that convert to/from `ArraySegment<byte>`? You would need to add a `ptrRoundtrip` test delegate but that is small.

cs/src/io/unsafe/MemoryHelper.cs
```diff
@@ -0,0 +1,127 @@
...
+using System.Runtime.InteropServices;
+using System.Security;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Bond.IO.Unsafe
+{
+    internal unsafe class MemoryHelper
+    {
+        [DllImport("msvcrt.dll", EntryPoint = "memcpy", CallingConvention = CallingConvention.Cdecl, CharSet = CharSet.Unicode, SetLastError = false)]
```
sapek: I would expect this to fail OS X test but it [https://travis-ci.org/Microsoft/bond/jobs/89251589](succeeded). Are we missing a test case? Are we missing a test case?

We need to keep the code cross-platform.

cs/src/io/unsafe/InputPtrBuffer.cs
```diff
@@ -0,0 +1,269 @@
...
+namespace Bond.IO.Unsafe
+{
+    using System;
+    using System.Diagnostics;
+    using System.IO;
+    using System.Text;
+    using Bond.Protocols;
+
+    /// <summary>
+    /// Implements IInputStream on top of memory buffer
```
sapek: unmanaged memory buffer

cs/test/core/Util.cs
redknightlois: Depends. If we want to be memory conservative in the tests yes. However, there is little gain and more complexity on the test side. After all we are treating an array as its ptr representation (ownership should not be transfered to the OutputPtrBuffer).

cs/src/io/unsafe/MemoryHelper.cs
redknightlois: True, that wont work on other than Windows. Any idea how we do a native memcpy on the other platforms supported? With 4.6 System.Buffer does expose it, but in 4.5.x it doesnt exist.

redknightlois:
Travis build failed because there is a file that has been moved permanently. Is that something to fix in this PR?

cs/test/core/Util.cs
sapek: When you see a test case using `prtRoundtrip` you *know* you are testing with unmanaged memory because the type enforces that. If you see `memoryRoundtrip` than you are not sure until you look at the implementation of the Serialize/Deserialize delegate.

sapek:
I'm looking the Travis failures. Don't worry about it...

cs/test/core/Util.cs
redknightlois: Fair enough. :)

cs/src/io/unsafe/MemoryHelper.cs
```diff
@@ -0,0 +1,127 @@
...
+                case 1:
+                    *dest = *src;
+                    return;
+                case 0:
+                    return;
+                default:
+                    break;
+            }
+
+            if (n <= 512)
```
sapek: Comment say 4096 but here is is 512.

redknightlois: Nice catch, I changed that to be more conservative (performance wise) on certain CPUs a looooong time ago.

cs/src/io/unsafe/MemoryHelper.cs
sapek: Do you think it would be worth to have a test just for this function, hitting all the cases of buffer length? We already have a unit test project for internal functions.

redknightlois: The missing test case because of your latter comment should be 512 bytes of output.

redknightlois: It is very tested on our end, but because of OSX implementations and potentially CoreCLR down the line. Yes, I think we absolutely have to add such a test case. It is a very easy to test function :)

cs/src/io/unsafe/InputPtrBuffer.cs
```diff
@@ -0,0 +1,269 @@
...
+            position += sizeof(double);
+            return result;
+        }
+
+        /// <summary>
+        /// Read an array of bytes verbatim
+        /// </summary>
+        /// <param name="count">Number of bytes to read</param>
+        /// <exception cref="EndOfStreamException"/>
+        public ArraySegment<byte> ReadBytes(int count)
```
redknightlois: While you are at it... We could decide what to do with this... The most performance oriented implementation should expose byte* and the parsers should be able to deal with that instead.

sapek: Yes, this is sub-optimal, making `byte*` slower than `byte[]` in this case. But this isn't that simple to fix because this is dictated by the type used to represent `blob` in generated classes and it percolates throughout, not just `IInputStream` but also protocol interfaces. And I'm not sure if it would even be desirable to allow representing `blob` as a `byte*` to the source buffer, lifetime management would be very error prone.

redknightlois: We can always use some internal buffer pool to deal with that to achieve similar speed at the expense of a bit more complexity though.

redknightlois: We still wont be able to side-step the memcpy but at least we can avoid the allocations.

cs/src/io/unsafe/MemoryHelper.cs
redknightlois: Is there any way to know if you are building on OSX and or Windows? Is the same nuget package built for all of them or there are different builds for each?

The memcpy for linux and osx can be defined as: [DllImport("libc", EntryPoint = "memcpy")] in those platforms.

sapek: The same binaries can be used on all platforms. Let me ping some .Net experts to see what options we have here...

cs/test/core/Util.cs
redknightlois: Looking into this and the main issue is now you cannot test compatibility roundtrip between unsafe and safe ptr without introducing all the combinations. Which sounds a bit too complex when the current can deal with it at the expense of having to enter into the delegate to understand what it is happening. What do you think? It is worth the complexity?

sapek: This is a good point. I don't think we need all combination but we should add another rountrip delegate which serializes to `ArraySegment<byte>` and deserializes using ` byte*`. This shouldn't add much complexity.

redknightlois: That would ensure Managed Serialize - Unmanaged Deserialize compatibility but not the other way around. :)

sapek: Isn't compatibility we are considering here transitive though?

redknightlois:
```
Managed Serialize -> Unmanaged deserialize
Unmanaged Serialize -> Unmanaged deserialize
```
Does not imply:
```
Unmanaged Serialize -> Managed Deserialize
```

It is OK for me, if you are convinced enough there is no posibility it won't introduce problems down the line. Just wanted to note that would be required for completeness.

sapek: But we also have:
```
Managed Serialize -> Managed Deserialize
Unmanaged Serialize -> Unmanaged Deserialize
```
Which I *think* should cover it. Although I will also extend the compatibility test, which covers more than current version roundtrip, to use all the streams.

cs/src/io/unsafe/MemoryHelper.cs
sapek: It seems like there isn't a good option to support `memcpy` in a cross-platform way. It would require building platform specific assemblies.

I would suggest to ditch `memcpy` and use `System.Buffer.MemoryCopy` conditionally on 4.6 and otherwise just the loop in C#. We don't yet build a separate version for .NET 4.6 but we can add that and then the `MemoryCopy` will light up.

sapek: Another option that was suggested is to use `Marshal.Copy` since what we really need here is copy from unmanaged memory into `byte[]`. 

Also for completeness: I have learned that Mono does have a DllMap feature allowing to remap `DllImport` via configuration. This would actually make it possible to import `memcpy` from different libraries on different platforms. But that is a rather unwieldy thing to maintain so I don't think we should do it here.

redknightlois:
Applied the changes we talked about.

sapek:
One of the tests needs a bigger output buffer.

cs/src/io/unsafe/InputPtrBuffer.cs
```diff
@@ -0,0 +1,267 @@
...
+        public uint ReadUInt32()
+        {
+            if (position > end - sizeof(uint))
+            {
+                EndOfStream(sizeof(uint));
+            }
+            uint result = buffer[position++];
+            result |= ((uint)buffer[position++]) << 8;
+            result |= ((uint)buffer[position++]) << 16;
+            result |= ((uint)buffer[position++]) << 24;
```
sapek: Shouldn't that be:
```
var result = *((uint*)(buffer + position));
position += sizeof(uint);
```

cs/src/io/unsafe/OutputPtrBuffer.cs
```diff
@@ -0,0 +1,243 @@
...
+            var b = data;
+            b[i++] = (byte)value;
+            b[i++] = (byte)(value >> 8);
+            b[i++] = (byte)(value >> 16);
+            b[i++] = (byte)(value >> 24);
+            b[i++] = (byte)(value >> 32);
+            b[i++] = (byte)(value >> 40);
+            b[i++] = (byte)(value >> 48);
+            b[i++] = (byte)(value >> 56);
+            position = i;
```
sapek: Why not:
```
*((ulong*) (data + position)) = value;
position += sizeof(ulong);
```

The same for other little-endian, fixed-size ints and for `float`/`double`

cs/src/io/unsafe/IntegerHelper.cs
```diff
@@ -0,0 +1,244 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Bond.IO.Unsafe
+{
+    using System;
+    using System.Runtime.CompilerServices;
```
sapek: unneeded/unused using declarations

cs/src/io/unsafe/InputPtrBuffer.cs
```diff
@@ -0,0 +1,267 @@
...
+            }
+
+            sbyte* ptr = (sbyte*) (buffer + position);
+            var result = new string(ptr, 0, size, encoding); // TODO: Check this
+
+            position += size;
+            return result;
+        }
+
+        private ulong DecodeVarUInt64Checked()
```
sapek: We don't use explicit `private` in the codebase.

cs/src/io/unsafe/InputPtrBuffer.cs
```diff
@@ -0,0 +1,267 @@
...
+
+    /// <summary>
+    /// Implements IInputStream on top of unmanaged memory buffer
+    /// </summary>
+    public unsafe sealed class InputPtrBuffer : IInputStream, ICloneable<InputPtrBuffer>
+    {
+        readonly int offset;
+        internal byte* buffer;
+        internal int end;
+        internal int position;
```
sapek: Could be:
```
readonly byte* buffer;
readonly int end;
int position;
```

cs/src/io/unsafe/OutputPtrBuffer.cs
```diff
@@ -0,0 +1,243 @@
...
+
+    /// <summary>
+    /// Implements IOutputStream on top of unmanaged memory buffer
+    /// </summary>
+    public sealed unsafe class OutputPtrBuffer : IOutputStream
+    {
+        internal byte* data;
+        internal int position;
+        internal int length;
+
```
sapek: Could be:
```
readonly byte* data;
readonly int length;
int position;
```

cs/src/io/unsafe/OutputPtrBuffer.cs
redknightlois: That was a verbatim copy, didnt need it to modify to make it work. But you are right we can be even more efficient there.

sapek: Yes, that was a copy from `Safe.OutputBuffer`. `Unsafe.OutputBuffer` already uses pointer. The unmanged version should as well.

redknightlois:
Ok, that should be all.

redknightlois:
Any idea why the build has failed?

sapek:
Haskell package conflicts. Fixed with 55cf387e2e8afdc274e72a8ad84b3d8f98ac97f3.

sapek:
Should we use `IntPtr` instead of `byte*` in the APIs? This would allow using the unmanaged buffers from safe code.

Here are proposed changes: https://github.com/sapek/bond/commits/redknightlois-master

I also made some changes to tests:
- I rewrote the int limits test (67e2db49fbf2). Now it is not necessary to update it with this change.
- Moved the code to test interop between managed and unmanaged into [roundtrip delegate](https://github.com/sapek/bond/blob/eed40b931c00c81f560ce3ef23d6451f749a121c/cs/test/core/Util.cs#L611-L618).
- Removed unsafe code from unit test.

If these changes look good I can include them in the merge.

There is one final thing: the names of the buffer classes. I'm not sure if `PtrBuffer` in the name conveys what we want here. It sounds like *buffer of pointers*. 

Some other names I considered: 
 - `UnmanagedInputBuffer` (self-explanatory but .NET APIs don't seem to use the term `Unmanaged`)
 - `InputPtr` (complements/consistent with `InputBuffer` and `InputStream` naming, however I'm not sure if it won't create an expectation of something similar to `IntPtr`)

ranrav:
To me InputPointer sounds like a better name because A) .NET conventions don't like abbreviations in public names and B) InputPointer is more likely to be interpreted as "a pointer serving as an input" and not as "a pointer pointing at an input".

Sent from my Windows Phone

redknightlois:
@sapek The only problem I see with IntPtr (which should exist as a constructor too) is that it requires an explicit cast.

```csharp
unsafe class Program
{
    public static void UsingIntPtr(IntPtr ptr)
    { }

    static void Main(string[] args)
    {              
        var array = new byte[1000];
        fixed (byte* ptr = array)
            UsingIntPtr((IntPtr)ptr);
    }
}
```
From someone that deals with unmanaged byte arrays everyday I can say that the code gets pretty ugly pretty fast. I don't expect the JIT to allocate in that boxing but I am not that sure about that either.

@ranrav Not really true, Ptr is a pretty common abbreviation used through the framework. ```IntPtr``` is just one of those.

sapek:
@redknightlois, I pushed the latest proposal to https://github.com/sapek/bond/commits/redknightlois-master

I that looks good I will squash the commits and merge.

sapek:
@redknightlois, did you have a chance to look at my proposed changed?

redknightlois:
Yes. I am actually using those as code without trouble while waiting for it to be available in the nuget package. LGTM

sapek:
Manually merged with 01b3587f1a15a16cef9c10bd4f7296a75a17f2ed. I will publish a new Nuget package soon.
Thanks for contributing.

redknightlois:
Just for the purpose of completeness. These are the results using Bond and the pointer buffers and a very small metaprogramming core. Important to say that the code being replaced was quite optimized. 

bond-raw

PR #102, opened 2015-11-13
Removed all private modifiers as per review comment

redknightlois:
There were many private modifiers, reason why I didnt notice the code avoided it's use. I just went ahead a remove all of them in this PR. 

Removed all private modifiers as per review comment at https://github.com/Microsoft/bond/pull/93


sapek:
Reworded the commit message to follow length line limits and merged from command line. Thx for fixing this.

PR #105, opened 2015-11-26
Support for serialization and deserialization of IReadOnly... interfaces in c#

lukakms:
Changed logic of DeserializerTransform in case of collection interfaces which can't be directly written to.
If a collection is supposed to be initialized, an object of schemaType will be created first and that object will be used for deserialization.
The object will then be assigned to the property with non-writable interface.

### DIFF
--- a/cs/src/core/expressions/DeserializerTransform.cs
+++ b/cs/src/core/expressions/DeserializerTransform.cs
@@ -4,6 +4,7 @@
 namespace Bond.Expressions
 {
     using System;
+    using System.Collections;
     using System.Collections.Generic;
     using System.Linq;
     using System.Linq.Expressions;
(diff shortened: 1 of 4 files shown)


sapek:
A few notes about this PR:
- Please build/run all configurations before submitting. All builds failed.
- There should be tests for nested containers with read-only collection.
- Please follow http://chris.beams.io/posts/git-commit/ for commit message. Of course you can do this at the end with rebase. 
- Finally: it might be possible to implement this with a much simpler change. It seems like handling needed for all read-only collections is the same: deserialize into a temp variable and assign to the field. Please see if something similar to the way [struct fields are handled](https://github.com/Microsoft/bond/blob/57662ca5f4c035295b5a0d11977aae5bfcfc7093/cs/src/core/expressions/DeserializerTransform.cs#L395-L401) wouldn't be sufficient here.

chwarr:
We, the Bond team, haven't heard from you about this pull request in a while. If you want to continue with this, can you address @sapek's comments from November and re-open this PR? Thanks.

PR #116, opened 2015-12-15
Support Serializers without nested struct inlining

ranrav:
Add support to create a Serializer without inlining of nested types.
Serialization would be slightly slower, but compilation time can be significantly decreased for complex type graphs.


sapek:
Merged with cb95fdb3e1e10c3e4cae2f2d55e2b116041010a8.

PR #120, opened 2016-01-13
Add rudimentary Visual Studio Code syntax highlighting

robrodi:
It's fairly simplistic, but should be a decent start.
Based on the [VS Code Sample][1] + bond's Sublime Text Plugin.

[1]:https://code.visualstudio.com/docs/customization/colorizer


tools/syntax/VsCode/syntaxes/bond.tmLanguage
```diff
@@ -0,0 +1,168 @@
...
+            <dict>
+                <key>1</key>
+                <dict>
+                    <key>name</key>
+                    <string>keyword.other.namespace.source.bond</string>
+                </dict>
+                <key>2</key>
+                <dict>
+                    <key>name</key>
+                    <string>storage.type.bond</string>
```
chwarr: The second capture corresponds to either the cpp or the csharp part, if I'm reading this correctly. Does "storage.type.bond" have a special meaning here? Why "storage"? The cpp/csharp part is the language-specific namespace, so if we have flexibility on the naming, we should reflect that.

tools/syntax/VsCode/syntaxes/bond.tmLanguage
```diff
@@ -0,0 +1,168 @@
...
+                </dict>
+                <key>3</key>
+                <dict>
+                    <key>name</key>
+                    <string>entity.name.type.namespace.source.bond</string>
+                </dict>
+            </dict>
+            <key>match</key>
+            <string>^(namespace)\s+(cpp|csharp) (.+)$</string>
```
chwarr: The `namespace foo` statement does not have to appear at the beginning of the line. Is leading/trailing whitespace already stripped?

chwarr: `namespace foo` is valid. One does not need to specify `cpp` or `csharp`. I think that the second capture needs to be made optional.

tools/syntax/VsCode/syntaxes/bond.tmLanguage
```diff
@@ -0,0 +1,168 @@
...
+            <key>name</key>
+            <string>meta.namespace.source.cs</string>
```
chwarr: `.bond` instead of `.cs`?

tools/syntax/VsCode/syntaxes/bond.tmLanguage
sapek: Specifying language like `cpp` or `csharp` in a namespace declaration is not just not needed, it is very strongly discouraged. It is a deprecated feature only present to support some legacy schemas. If we need to highlight this part, could we choose some bright red color by default?

tools/syntax/VsCode/syntaxes/bond.tmLanguage
chwarr: 'constant.language.language.bond' instead of 'storage.type.bond'?

I'm using the [TextMate Language Grammars](http://manual.macromates.com/en/language_grammars) page as my reference. Happy to be pointed to more up-to-date/VS Code-specific references if they apply.

sapek:
Note that build break is caused by cabal dependency conflicts caused by introduction of a new major version of `transformers` package. This [PR](https://github.com/ekmett/exceptions/pull/51) would fix it but we might have to work around until `exceptions` package is updated.

tools/syntax/VsCode/syntaxes/bond.tmLanguage
```diff
@@ -0,0 +1,168 @@
...
+            <dict>
+                <key>1</key>
+                <dict>
+                    <key>name</key>
+                    <string>keyword.other.enum.source.bond</string>
+                </dict>
+                <key>2</key>
+                <dict>
+                    <key>name</key>
+                    <string>storage.type.source.enum.bond</string>
```
chwarr: `entity.name.type.enum.bond`?

tools/syntax/VsCode/syntaxes/bond.tmLanguage
```diff
@@ -0,0 +1,168 @@
...
+                    <key>captures</key>
+                    <dict>
+                        <key>1</key>
+                        <dict>
+                            <key>name</key>
+                            <string>entity.name.bond</string>
+                        </dict>
+                    </dict>
+                    <key>match</key>
+                    <string>^\s*([^\s]*)\s</string>
```
chwarr: How will this handle an enum with a value, like

    enum Foo {
        Bar = 17,
    }

tools/syntax/VsCode/syntaxes/bond.tmLanguage
```diff
@@ -0,0 +1,168 @@
...
+                    <key>name</key>
+                    <string>entity.name.type.enum.value.bond</string>
+                </dict>
+            </array>
+        </dict>
+        <dict>
+            <key>begin</key>
+            <string>(struct)\s+(.+)</string>
```
chwarr: What about bases?
`struct Foo : Bar`

tools/syntax/VsCode/syntaxes/bond.tmLanguage
```diff
@@ -0,0 +1,168 @@
...
+            <key>captures</key>
+            <dict>
+                <key>1</key>
+                <dict>
+                    <key>name</key>
+                    <string>keyword.other.struct.source.bond</string>
```
chwarr: Here and elsewhere in the file where we name elements, do we need the `source` part? That looks like it is only for the filenames to match.

tools/syntax/VsCode/syntaxes/bond.tmLanguage
```diff
@@ -0,0 +1,168 @@
...
+                    <key>match</key>
+                    <string>//.+</string>
+                    <key>name</key>
+                    <string>comment.line.double-slash.bond</string>
+                </dict>
+                <dict>
+                    <key>match</key>
+                    <string>"[^"]+"</string>
+                    <key>name</key>
+                    <string>string.quoted.double.source.cs</string>
```
chwarr: `.bond`?

tools/syntax/VsCode/syntaxes/bond.tmLanguage
```diff
@@ -0,0 +1,168 @@
...
+                    <key>match</key>
+                    <string>\b((0(x|X)[0-9a-fA-F]*)|(([0-9]+\.?[0-9]*)|(\.[0-9]+))((e|E)(\+|-)?[0-9]+)?)(L|l|UL|ul|u|U|F|f|ll|LL|ull|ULL)?\b</string>
+                    <key>name</key>
+                    <string>constant.numeric.bond</string>
+                </dict>
+                <dict>
+                    <key>match</key>
+                    <string>\b(true|false|nothing)\b</string>
+                    <key>name</key>
+                    <string>constant.language.source</string>
```
chwarr: `.bond`?

tools/syntax/VsCode/syntaxes/bond.tmLanguage
```diff
@@ -0,0 +1,168 @@
...
+    <dict>
+        <key>commonStuff</key>
+        <dict>
+            <key>patterns</key>
+            <array>
+                <dict>
+                    <key>match</key>
+                    <string>u?int(8|16|32|64)|double|bool|wstring|map|set|vector|nullable</string>
+                    <key>name</key>
+                    <string>constant.language.source</string>
```
chwarr: Can we consistently use `.bond` at the end of the element names?

chwarr: `storage.type.bond`?

tools/syntax/VsCode/syntaxes/bond.tmLanguage
```diff
@@ -0,0 +1,168 @@
...
+    </array>
+    <key>repository</key>
+    <dict>
+        <key>commonStuff</key>
+        <dict>
+            <key>patterns</key>
+            <array>
+                <dict>
+                    <key>match</key>
+                    <string>u?int(8|16|32|64)|double|bool|wstring|map|set|vector|nullable</string>
```
chwarr: Missing string, list, float, void. Possibly want to include blob, bonded

chwarr:
Thanks for the PR. I spent some time looking over the changes--and learning about the TextMate syntax, so please let me know if my comments don't make sense because of something I've missed in the TextMate syntax.

tools/syntax/VsCode/syntaxes/bond.tmLanguage
robrodi: I think we (halo) are the legacy partner that uses it :+1:

tools/syntax/VsCode/syntaxes/bond.tmLanguage
robrodi: yeah, i copied and pasted a ton of their grammars, but it seemed to work.  Changed to `constant.language.language.bond` per your suggestion

tools/syntax/VsCode/syntaxes/bond.tmLanguage
robrodi: Just tested it out with that exact example, and it highlights correctly.

tools/syntax/VsCode/syntaxes/bond.tmLanguage
robrodi: Can I add this in a future drop?  It's currently all colored the same (which isn't great, but not _terrible_)

chwarr: Makes sense.

tools/syntax/VsCode/syntaxes/bond.tmLanguage
robrodi: C# had that as a keyword rather than a type.  Happy to change it if you feel strongly about it.

chwarr: `storage.type.enum.bond` seems fine too.

tools/syntax/VsCode/syntaxes/bond.tmLanguage
robrodi: I don't remember why that was there.  So i'll cut it :)

tools/syntax/VsCode/syntaxes/bond.tmLanguage
robrodi: fixed as part of removing `source` from a million places :+1:

chwarr:
OK. Build issues should be fixed now. Can you rebase and update this PR?

chwarr:
Hmmm... The rebase went all wonky and is now including commits that I wouldn't have expected. Can you try that again? (The history that we want to see is mering of a branch with one commit.)

I was able to get this fixed up locally by
```
git log
git fetch origin # or whatever Microsoft/bond is for you
git checkout CodeSupport
git reset --hard origin/master
git cherry-pick 7fe3243e745d64fd1bc6170f5814a4e737233212 # or whatever the commit with the VS Code changes ends up being
```

Issue #111, opened 2015-12-10
Field ordinals should be rejected if they are larger than the largest allowed ordinal


chwarr:
gbc is parsing field ordinals larger than 65535 and then converting them to unsigned 16-bit integers. This results in unexpected field ordinals and collisions.

Given the following .bond file:

    namespace fieldIdWrapping;
    
    struct SameFieldIds
    {
        0: int32 FieldWithId0;
        65536: string FieldWithId0AsWell;
    }

when I run `.\gbc.exe c++ .\duplicate_fields.bond`, I get the following error:

    Error parsing .\duplicate_fields.bond: ".\duplicate_fields.bond" (line 8, column 1):
    unexpected end of input
    Duplicate definition of the field with ordinal 0

Instead of using wrap-around semantics, this should fail code generation with an error about a field ordinal being too large. Bonus points for including the struct and field name in the error message.

Version info:

    > .\gbc.exe -V
    Bond Compiler 0.4.0.1, (C) Microsoft

chwarr:
The largest allowed ordinal is 65535. From [bond.bond](https://github.com/Microsoft/bond/blob/ec7e0eb39ba95e827d979a1dc49419f22e820e05/cpp/inc/bond/core/bond.bond#L85):

```
// Schema representing a field definition
struct FieldDef
{
    // Field metadata
    0: Metadata metadata;           
    
    // Field ordinal
    1: uint16 id;

    // Field type definition
    2: TypeDef type;
}
```

PR #124, opened 2016-01-21
Add check for field ordinals, should be less than max value of Word16

lalo:
Fix for issue #111 

Sample bond file:
```
namespace fieldIdWrapping;

struct SameFieldIds
{
    0: int32 FieldWithId0;
    65538: string FieldWithId0AsWell;
}
```
Generates error:
```
Error parsing bb.bond: "bb.bond" (line 6, column 10):
unexpected ":"
expecting digit
Field ordinal out of range 65538
```

First pull request to this project and I haven't done haskell work in a while so hopefully I'm not messing up with the style or any other considerations that I didn't consider.

Thanks

Eduardo

### DIFF
--- a/compiler/src/Language/Bond/Parser.hs
+++ b/compiler/src/Language/Bond/Parser.hs
@@ -24,6 +24,7 @@ module Language.Bond.Parser
 import Data.Ord
 import Data.List
 import Data.Function
+import Data.Word
 import Control.Applicative
 import Control.Monad.Reader
 import Prelude
@@ -247,7 +248,13 @@ manySortedBy = manyAccum . insertBy
 field :: Parser Field
 field = makeField <$> attributes <*> ordinal <*> modifier <*> ftype <*> identifier <*> optional default_
   where
-    ordinal = (fromIntegral <$> integer) <* colon <?> "field ordinal"
+    ordinal = word16 <* colon <?> "field ordinal"
+      where
+        word16 = do
+            i <- integer
+            if i <= toInteger (maxBound :: Word16) && i >= toInteger (minBound :: Word16)
+                then return (fromInteger i)
+                else fail "Field ordinal must be within the range 0-65535"
     modifier = option Optional
                     (keyword "optional" *> pure Optional
                  <|> keyword "required" *> pure Required


compiler/src/Language/Bond/Parser.hs
```diff
@@ -247,7 +248,11 @@ manySortedBy = manyAccum . insertBy
 field :: Parser Field
 field = makeField <$> attributes <*> ordinal <*> modifier <*> ftype <*> identifier <*> optional default_
   where
-    ordinal = (fromIntegral <$> integer) <* colon <?> "field ordinal"
+    ordinal = fromIntegral <$> (bounded integer) <* colon <?> "field ordinal"
```
sapek: Instead of `(bounded integer)`, how about making it a `word16` parser?

As for style, please indent lines within `do`. Also commit message title should be no more than about 52 chars (makes for much nicer output from git on command line). Thx.

lalo:
Cool @sapek, added the parser for word16 with the latest change

sapek:
Maybe it would be better to keep the parser within `field` and have a more specific error message? We don't need the parser anywhere else and likely won't.

sapek:
Also please squash the commits.

lalo:
Gotcha, I'll refactor it inside the field parser within a where clause. Is this specific enough error message "Field ordinal out of range 65538"?

chwarr:
As a hint to the user, we want to say something like "Field ordinal 65538 is outside of the allowable range of 0-65537, inclusive."

So, how about this?

```
fail $ "Field ordinal " ++ show i ++ " is outside of the allowable range of " ++
          show (minBound :: Word16) ++ "-" ++ show (maxBound :: Word16) ++ ", inclusive."
```

@sapek, would it make sense to use `printf` from `Text.Printf` to format long strings like this, or is the formulation with `++` good enough. Something like:

```
word16 :: Parser Word16
word16 = do
    i <- integer
    if i <= toInteger (maxBound :: Word16) then
            return (fromInteger i)
        else
            fail msg
            where msg :: String
                  msg = printf "Field ordinal %d outside of range %d-%d, inclusive"
                               i (minBound :: Word16) (maxBound :: Word16)
```

sapek:
I wouldn't over-complicate this. A simple string literal "Field ordinal must be within the range 0-65537" should be enough. The error contains line number, which is what is important for the user to fix it, not the invalid ordinal value.

chwarr:
Right, `fail` already reports the line number. That's going to be good enough.

PR #127, opened 2016-01-29
Use scoped syntax to refer to BondDataType, ProtocolType, and Modifier

jdubrule:
Use scoped syntax for BondDataType, Modifier, and ProtocolType, which are defined in bond.bond and bond_const.bond.
This does not change Bond to generated 'enum class ___' enums, nor does it require that you use the namespace when using the enums.

This puts Bond a step towards generating 'enum class' definitions as an option.  Plus, we've locally converted Bond over to generating 'enum class' definitions, and this will reduce the differences in our code from the stock bond.

### DIFF
--- a/cpp/inc/bond/core/detail/cmdargs.h
+++ b/cpp/inc/bond/core/detail/cmdargs.h
@@ -99,7 +99,7 @@ public:
 
     bool IsOptional() const
     {
-        return metadata.modifier == bond::Optional;
+        return metadata.modifier == bond::Modifier::Optional;
     }
 
 private:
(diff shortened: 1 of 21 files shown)


chwarr:
Hmmm... Our enum headers and all the various C++ compilers aren't playing well together...

sapek:
There is a [test](https://github.com/Microsoft/bond/blob/3f4d62497649512f78dafd6bcc55c79ae1f5abaa/cpp/test/core/basic_tests.cpp#L261-L265) for this functionality. We should look what is different in the PR...

chwarr:
Closing to re-open to trigger the CLA bot again.

chwarr:
Hello, CLA bot, I know that you don't actually read these messages, but please check agian.

chwarr:
@jdubrule, if you want to continue with this, can you look into the failing builds and re-open this PR? Thanks.

Issue #96, opened 2015-11-06
NotImplementedException when trying to deserialize complex type from json


SergeyKanzhelev:
I want to deserialize complex bond. Exception got thrown when I'm using ```SimpleJsonWriter```. Is there some examples on json deserialization of complex types? Here is a schema I'm trying:

``` json
struct InnerStruct
{
    10: optional string baseType;
}
struct Record
{
    0: string Name;
    1: vector<double> Constants;
    2: optional bonded<InnerStruct> data;
}
```

Code:

``` csharp
var src = new Record
{
    Name = "FooBar",
    Constants = { 3.14, 6.28 },
    data = new Bonded<InnerStruct>(new InnerStruct() { baseType = "a" })
};

var jsonString = new StringBuilder();
var jsonWriter = new SimpleJsonWriter(new StringWriter(jsonString));

Serialize.To(jsonWriter, src);
jsonWriter.Flush();

var reader = new SimpleJsonReader(new StringReader(jsonString.ToString()));
src = Deserialize<Record>.From(reader);
```

Output:

```
{"Name":"FooBar","Constants":[3.14,6.28],"data":{"baseType":"a"}}

Unhandled Exception: System.TypeInitializationException: The type initializer for 'Cache`1' threw an exception. ---> System.NotImplementedException: The method or operation is not implemented.
   at Bond.Expressions.Pull.PullParser`1.Bonded(ValueHandler handler)
   at Bond.Expressions.DeserializerTransform`1.Value(IParser parser, Expression var, Expression valueType, Type schemaType, Boolean initialize)
   at Bond.Expressions.DeserializerTransform`1.CheckedValue(IParser parser, Expression var, Expression valueType, Type schemaType, Boolean initialize)
   at Bond.Expressions.DeserializerTransform`1.<>c__DisplayClass11.<>c__DisplayClass15.<Struct>b__c(IParser fieldParser, Expression fieldType)
   at Bond.Expressions.Field.Bond.Expressions.IField.Value(IParser parser, Expression valueType)
   at Bond.Expressions.Json.SimpleJsonParser`1.ProcessField(ParameterExpressionrequiredFields, IEnumerable`1 transforms)
   at Bond.Expressions.Json.SimpleJsonParser`1.<>c__DisplayClass9.<CreateStateMachine>b__8(Expression state)
   at Bond.Expressions.Pull.StateTransition.<>c__DisplayClass2.<.ctor>b__0(Expression s)
[...]

PR #130, opened 2016-02-16
Provide serialization and deserialization of objects with only a RuntimeSchema

csdahlberg:
This includes the addition of RuntimeSerializer and RuntimeDeserializer, which use a new RuntimeObject type (instead of actual .NET types) to represent bond structs. That allows bond data to be serialized/deserialized using just a RuntimeSchema, without a corresponding .NET type.

Unit tests were added to verify that RuntimeDeserializer deserializes values that are the same as Deserializer, and that RuntimeSerializer serializes the exact same bytes as Serializer.

### DIFF
--- a/cs/src/core/RuntimeBonded.cs
+++ b/cs/src/core/RuntimeBonded.cs
@@ -1,0 +1,57 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Bond
+{
+    using System;
+
+    /// <summary>
+    /// Interface representing bonded payload of unknown type
+    /// </summary>
+    public interface IRuntimeBonded
+    {
+        /// <summary>
+        /// Deserialize an object from the IRuntimeBonded instance
+        /// </summary>
+        /// <returns>Deserialized object</returns>
+        object Deserialize();
+    }
+
+    /// <summary>
+    /// Interface representing the schema type bonded&lt;T>
+    /// </summary>
+    /// <typeparam name="T">Type representing a Bond schema</typeparam>
+    public interface IRuntimeBonded<out T> : IRuntimeBonded
+    {
+        /// <summary>
+        /// Deserialize an object of type T from the IRuntimeBonded&lt;T> instance
+        /// </summary>
+        /// <returns>Deserialized object</returns>
+        new T Deserialize();
+    }
+
+    public class RuntimeBonded<T> : IRuntimeBonded<T>
+    {
+        private readonly Func<T> valueDeserializer;
+
+        public RuntimeBonded(T value)
+            : this(() => value)
+        {
+        }
+
+        public RuntimeBonded(Func<T> valueDeserializer)
+        {
+            this.valueDeserializer = valueDeserializer;
+        }
+
+        object IRuntimeBonded.Deserialize()
+        {
+            return Deserialize();
+        }
+
+        public T Deserialize()
+        {
+            return valueDeserializer();
+        }
+    }
+}
--- a/cs/src/core/RuntimeDeserializer.cs
+++ b/cs/src/core/RuntimeDeserializer.cs
@@ -1,0 +1,671 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Bond
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+    using Bond.IO.Safe;
+    using Bond.Protocols;
+
+    /// <summary>
+    /// Deserializes <see cref="RuntimeObject"/>s from a protocol reader using only a runtime schema.
+    /// </summary>
+    public static class RuntimeDeserialize
+    {
+        /// <summary>
+        /// Deserializes a <see cref="RuntimeObject"/> from a tagged protocol reader using only a runtime schema.
+        /// </summary>
+        /// <param name="reader">The tagged protocol reader to read the object data from.</param>
+        /// <param name="schema">The schema to use when reading the object data.</param>
+        /// <returns>A <see cref="RuntimeObject"/> created from <paramref name="reader"/> and <paramref name="schema"/>.</returns>
+        public static RuntimeObject From(IClonableTaggedProtocolReader reader, RuntimeSchema schema)
+        {
+            return new RuntimeDeserializer(schema).Deserialize(reader);
+        }
+
+        /// <summary>
+        /// Deserializes a <see cref="RuntimeObject"/> from an untagged protocol reader using only a runtime schema.
+        /// </summary>
+        /// <param name="reader">The untagged protocol reader to read the object data from.</param>
+        /// <param name="schema">The schema to use when reading the object data.</param>
+        /// <returns>A <see cref="RuntimeObject"/> created from <paramref name="reader"/> and <paramref name="schema"/>.</returns>
+        public static RuntimeObject From(IClonableUntaggedProtocolReader reader, RuntimeSchema schema)
+        {
+            return new RuntimeDeserializer(schema).Deserialize(reader);
+        }
+    }
+
+    /// <summary>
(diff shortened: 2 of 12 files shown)


cs/src/core/RuntimeBonded.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// </summary>
+    public interface IRuntimeBonded
+    {
+        /// <summary>
+        /// Deserialize an object from the IRuntimeBonded instance
+        /// </summary>
+        /// <returns>Deserialized object</returns>
+        object Deserialize();
+    }
+
```
sapek: Why not `IBonded` and `Deserialize<object>`?

cs/src/core/RuntimeDeserializer.cs
```diff
@@ -0,0 +1,671 @@
...
+        public static RuntimeObject From(IClonableUntaggedProtocolReader reader, RuntimeSchema schema)
+        {
+            return new RuntimeDeserializer(schema).Deserialize(reader);
+        }
+    }
+
+    /// <summary>
+    /// Deserializes <see cref="RuntimeObject"/>s from a protocol reader using only a <see cref="RuntimeSchema"/>.
+    /// </summary>
+    public class RuntimeDeserializer
```
sapek: This implementation goes against the grain of Bond architecture. A better way would be to implement the RuntimeDeserializer as a [transform](https://github.com/Microsoft/bond/blob/1da75c2ffa91d8a20a22ee2cb9714aef28217872/cs/src/core/expressions/IParser.cs#L67-L70). This way all the parsing code would be reused and the deserializer would automatically work for all current and future Bond protocols (this doesn't work for JSON/Xml). It would also benefit from all the performance optimizations in existing the parser implementations.

cs/src/core/RuntimeSerializer.cs
```diff
@@ -0,0 +1,246 @@
...
+        public static void To(IProtocolWriter writer, RuntimeSchema schema, RuntimeObject obj)
+        {
+            new RuntimeSerializer(schema).Serialize(writer, obj);
+        }
+    }
+
+    /// <summary>
+    /// Serializes <see cref="RuntimeObject"/>s to a protocol writer using only a runtime schema.
+    /// </summary>
+    public class RuntimeSerializer
```
sapek: A better way would be to implement [IParser](https://github.com/Microsoft/bond/blob/1da75c2ffa91d8a20a22ee2cb9714aef28217872/cs/src/core/expressions/IParser.cs#L114-L118) for `RuntimeObject`. It would be conceptually similar to [`ObjectParser`](https://github.com/Microsoft/bond/blob/1da75c2ffa91d8a20a22ee2cb9714aef28217872/cs/src/core/expressions/ObjectParser.cs#L14).

cs/src/core/RuntimeBonded.cs
csdahlberg: I thought that reusing the existing IBonded interfaces would have made it too easy to lose track of whether a bonded object is a "normal" one (i.e. a .NET class) or an instance of RuntimeObject. The method signatures would also make it would seem like a bonded RuntimeObject could be passed into the "normal" Serializer (or vice-versa), which would never work.

Making Deserialize generic seemed much less helpful than it is with the "normal" Deserializer, since the likely use case for all of this is in reflection-like scenarios where you don't know what type is being deserialized (and all calls would use "object" as the generic type argument, anyway).

cs/src/core/RuntimeDeserializer.cs
csdahlberg: I agree that this does not follow how existing Bond deserialization works, but that was out of necessity. In existing code, transforms are created/compiled once on a per-.NET-type basis, but we will often not have corresponding .NET types for the schema we are deserializing. A transform could be created for the provided schema every time a RuntimeDeserializer is created, but that seemed unreasonable.

I do not see an obvious solution for making this work for for all current and future Bond protocols. Explicit support for JSON and XML could be added, but that would require either moving RuntimeSerializer and RuntimeDeserializer to a new project (so it can reference the JSON project) or creating something like RuntimeJsonSerializer and RuntimeJsonDeserializer in the JSON project.

sapek: The fact that existing deserializer is instantiated per C# type is a choice, not necessity. We do this for every C# type because deserializer *is* different for every type.

There is nothing preventing you from having just one instance per protocol. For example that's how [Transcode](https://github.com/Microsoft/bond/blob/1da75c2ffa91d8a20a22ee2cb9714aef28217872/cs/src/core/Transcoder.cs#L10-L13) API works - there is one instance per a pair of protocols. It still reuses the existing parsers.

BTW, it might be good to have an option to generate different deserializer per `RuntimeSchema`. There is potential to get much better performance this way because you can process the `SchemaDef` at expression generation time rather than runtime.

Support for all protocols would work the same way it works for all other transforms. Note for example that `Deserializer` is fully implemented in `Bond.dll`, it knows nothing about `Bond.JSON.dll` and you can [use it to deserialize](https://github.com/Microsoft/bond/blob/1da75c2ffa91d8a20a22ee2cb9714aef28217872/examples/cs/core/simple_json/program.cs#L28) from JSON.

If you don't want to go full-blown transform-based deserializer, there might be another approach that is much simpler: implement it as `IProtocolWriter` and use transcoding. Since your `RuntimeObject` is simply a dictionary from string to object, and in [WriteFieldBegin](https://github.com/Microsoft/bond/blob/1da75c2ffa91d8a20a22ee2cb9714aef28217872/cs/src/core/protocols/IProtocolWriter.cs#L46) you get the `Metadata` which contains field name, you should be able to implement a state-full writer which puts all the fields into a tree of dictionaries. This way you also avoid duplicating all the parsing code. Have you considered/tried this?

BigWang:
It's good to have this feature. 
I am trying to use RuntimeSchema for JSON/XML data. However, SimpleJsonReader does not implement IClonableTaggedProtocolReader. Is there any comment for this?

cs/src/core/RuntimeDeserializer.cs
csdahlberg: This is my first time working with expression trees, and figuring out how everything works has been slow for me (not having source code to debug hurts!).

I agree that a more general solution like those you describe would be more appropriate (especially with there already being interest expressed in JSON and XML). With me not being very proficient with expressions and likely having higher-priority work to do during the day, I'm not sure how long it'll take me to change my implementation to use expressions (or maybe IProtocolWriter).

I opted not to cache deserializers on a per-schema basis primarily because SchemaDef objects are mutable.

sapek: I'd like to clarify my comment about an implementation based on `IProtocolWriter`. I don't think that would be a good general purpose implementation to contribute to Bond. But it could be a relatively simple solution for people who need similar functionality until we have a full, general purpose API in Bond.

As for implementation based on Bond parser/transform architecture, I concede that there is some learning curve and we actually need to provide more documentation/examples on how to do that.

csdahlberg: In that case, I don't think it'd be worth it for me to go down the IProtocolWriter path.

I could easily tweak the Runtime* types and their methods to be similar to existing types (using generic type parameters for readers/writers instead of specifying actual types), and use reflection along with the code in my pull request as an initial implementation to handle all current protocols. A full expression-based implementation could be created later on without any breaking interface changes.

What do you think?

sapek: I don't think this is a good idea.

It would mean *a lot* of code duplication. The `RuntimeDeserializer` and `RuntimeSerializer` already are 50% bigger than the completely generic `DeserializerTransform` and `SerializerTransform`.

Using reflection would be bad for performance, which has been very important for Bond. We've been very careful to limit use of reflection to the code that executes during initialization, once per type.

chwarr:
@BigWang, that sounds like issue #96 to me.

chwarr:
@csdahlberg, we, the Bond team, haven't heard from you about this pull request in a while. We don't feel like this approach is the right way to solve this problem. A transform or parser-based approach will support more protocols, have less code duplication, and likely have better performance. I'm going to close this PR for now. Feel free to open it again if there's any progress.

Thanks!

Issue #123, opened 2016-01-19
Field names should be rejected if they duplicate another field name


chwarr:
The following .bond file should fail `gbc` code generation, as it contains two fields with the same name.

```
namespace duplicate_field_names;

struct ShouldBeRejected
{
    0: bool DuplicateFieldName;
    1: bool DuplicateFieldName;
}
```

However, Bond Compiler 0.4.0.1 accepts this .bond file and generates C++ and C# code that does not compile.

Here's an excerpt of the generated C# code.

```
namespace duplicate_field_names
{
    using System.Collections.Generic;

    [global::Bond.Schema]
    [System.CodeDom.Compiler.GeneratedCode("gbc", "0.4.0.1")]
    public partial class ShouldBeRejected
    {
        [global::Bond.Id(0)]
        public bool DuplicateFieldName { get; set; }

        [global::Bond.Id(1)]
        public bool DuplicateFieldName { get; set; }

        public ShouldBeRejected()
            : this("duplicate_field_names.ShouldBeRejected", "ShouldBeRejected")
        {}

        protected ShouldBeRejected(string fullName, string name)
        {

        }
    }
} // duplicate_field_names
```

lalo:
For this one I was thinking of this simple change in Parser.h line 235:

```
    unique p = do
        fields' <- p
        case findDuplicates fields' fieldOrdinal ++ findDuplicates fields' fieldName of
            [] -> return fields'
            Field {..}:_ -> fail $ "Duplicate definition of the field with ordinal " ++ show fieldOrdinal ++
                " and field name " ++ show fieldName
      where
        findDuplicates xs ft = deleteFirstsBy (ordinal ft) xs (nubBy (ordinal ft) xs)
        ordinal ft = (==) `on` ft
```

The bad thing is that the error message will display both values as possible duplications when only one is duplicated. If we want a more specific error message for each case I can refactor it in some other way that allows this.

chwarr:
If we change the message to "Duplicate definition of the field with either ordinal 0 or field name foo", the text will cover both cases.

PR #132, opened 2016-02-24
Add check for duplicate field names

lalo:
Fix for issue #123 

Example bond file

```
namespace fieldIdWrapping;

struct SameFieldIds
{
    0: int32 FieldWithId0;
    2: string FieldWithId0;
}
```

Error displayed:

```
Error parsing test.bond: "test.bond" (line 7, column 2):
unexpected end of input
Duplicate definition of the field with ordinal 2 and name "FieldWithId0"
```


lalo:
Had to close to fix issue with line endings

lalo:
Apparently there are tests that have duplicate field names. I should update the tests after making sure that this is an appropriate fix.

chwarr:
Yes, changing `m_enum2` to something else is the right fix. You might have to regenerate the golden master test and include  as well. After building, something like `dist\build\gbc-test\gbc-test --accept` from the `compiler` directory should do it.

sapek:
Given that this is function to find duplicate *fields* calling it `whichField` would be confusing imo.

chwarr:
extractor? selector?

Maybe just combine it into the predicate?

lalo:
maybe rename to accessor?

PR #145, opened 2016-04-07
Add Implicit Conversion between Bond.GUID and System.Guid

EduardOstrovsky:
In order to make Bond.GUIDs first class citizens in .NET I figured that an implicit conversion between them and System.Guid would be helpful for developers. This way they can continue to use System.Guid throughout their system and not deal with the four property struct.

This will additionally help us move over to Bond v4 from our custom version of v3.


cs/test/core/Core.csproj
```diff
@@ -1,116 +1,118 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
chwarr: Looks like the line endings got messed up in this commit or some.

cs/test/core/Core.csproj
```diff
@@ -1,116 +1,118 @@
...
+    <Compile Include="$(IntermediateOutputPath)\Containers_types.cs" Condition="False" />
+    <Compile Include="$(IntermediateOutputPath)\Aliases_types.cs" Condition="False" />
+    <Compile Include="$(IntermediateOutputPath)\ReadOnly_types.cs" Condition="False" />
+    <Compile Include="$(IntermediateOutputPath)\NamespaceConflict_types.cs" Condition="False" />
+    <Compile Include="$(IntermediateOutputPath)\NamespaceConflictBond_types.cs" Condition="False" />
+    <Compile Include="$(IntermediateOutputPath)\Bond File With Spaces_types.cs" Condition="False" />
+    <!-- End Resharper Workaround -->
+  </ItemGroup>
+  <ItemGroup>
+    <Reference Include="Microsoft.VisualStudio.QualityTools.UnitTestFramework, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
```
chwarr: This addition is unexpected. Was it intentional? Can we change the tests to just use NUnit?

cs/src/core/Bond.cs
```diff
@@ -3,7 +3,27 @@
...
 {
-    using System.Collections.Generic;
+    using System;
+    using System.Collections.Generic;
+
+    public partial class GUID
+    {
+        public static implicit operator Guid(GUID bondGuid)
+        {
+            return new Guid((int)bondGuid.Data1, (short)bondGuid.Data2, (short)bondGuid.Data3, BitConverter.GetBytes((long)bondGuid.Data4));
```
chwarr: Minor: please split parameters onto their own lines, as this line is a bit long.

chwarr: Can we do the bit manipulation on `Data4` ourselves to avoid the `byte[]` allocation from `BitConverter.GetByte`?

Something like this (though, watch out for endianess issues; I might have the conversion backward):
```
(short)bondGuid.Data3,
(byte)bondGuid.Data4,
(byte)bondGuid.Data4 >> 8,
...);
```

cs/src/core/Bond.cs
```diff
@@ -3,7 +3,27 @@
...
+        }
+
+        public static implicit operator GUID(Guid systemGuid)
+        {
+            var bytes = systemGuid.ToByteArray();
+            var bondGuid = new GUID();
```
chwarr: Minor: the following syntax avoids duplication and should work with all the compilers we use:

```
var bondGuid = new GUID
{
    Data1 = BitConveter.ToUInt32(...).
   ...
};
```

cs/test/core/GuidConversionTests.cs
```diff
@@ -0,0 +1,29 @@
...
+    public class GuidConversionTests
+    {
+        [Test]
+        public void ImplicitConversion()
+        {
+            var systemGuid = Guid.NewGuid();
+            var bondGuid = (GUID)systemGuid;
+
+            // The Datas should be populated if it converted correctly
+            Assert.AreNotEqual(bondGuid.Data1, 0);
```
chwarr: I'd be more confident in the test if we used a static GUID and asserted the values of the Data1..Data4 members explicitly to make sure that each representation of the GUID is correct, instead of just that we can round trip a GUID.

(When you do this, you'll likely need to make an assumption about the endianess of the memory layout of the test machine. That's fine. Bonus points for encoding such an assumption in a test case.)

chwarr:
Can you please reword the commit message to follow [the style we like to use](https://github.com/Microsoft/bond/blob/master/CONTRIBUTING.md#rebase-commits)?

cs/test/core/GuidConversionTests.cs
jasonzio: static const int one = 1;
#define CPU_LITTLE_ENDIAN  (*(char*)(&one))

The CPU_LITTLE_ENDIAN macro will evaluate as True on a little-endian system, False on a Big-endian system.

Using Visual Studio, uint32_t _byteswap_long(uint32_t val) returns the byte-swapped value of val. For non-VS compilers, the needed code is easy to find. If the 32-bit part of the broken-out guid doesn’t match what it ought to, check to see if matches the byte-swapped form; if it does, emit a different failure message.

chwarr: For C#, there's also `BitConverter.IsLittleEndian`

cs/test/core/Core.csproj
EduardOstrovsky: Visual Studio messed things up for me, fixing it

cs/test/core/Core.csproj
EduardOstrovsky: Again Visual Studio doing its annoying thing. Fixed

cs/test/core/GuidConversionTests.cs
EduardOstrovsky: Tried to use your IPAddress trick in the conversion. Could not since it is not available in the portable .NET library used int he project.
Instead used BitConverter.IsLittleEndian and had two separate paths. Hopefully it's not too horrible

cs/test/core/GuidConversionTests.cs
```diff
@@ -0,0 +1,29 @@
...
+    public class GuidConversionTests
+    {
+        [Test]
+        public void ImplicitConversion()
+        {
+            var systemGuid = Guid.Parse("340FAA04-D19C-4B3D-A6F3-BA6581CCDCB8");
+            var bondGuid = (GUID)systemGuid;
+
+            // The Datas should be populated if it converted correctly
+            Assert.AreEqual(bondGuid.Data1, 873441796u);
```
chwarr: Minor: hex constants would make this easier to read.

cs/src/core/Bond.cs
```diff
@@ -3,7 +3,52 @@
...
 {
-    using System.Collections.Generic;
+    using System;
+    using System.Collections.Generic;
+
+    public partial class GUID
+    {
+        public static implicit operator Guid(GUID bondGuid)
+        {
+            return BitConverter.IsLittleEndian ?
```
chwarr: Bond doesn't yet support big endian machines (nor does .NET AFAIK), so it's going to be sufficient to just have the little endian code and a test like
`Assert.IsTrue(BitConcerted.IsLittleEndian, "GUID conversion is only little endian for now.");`
so that the big endian port will break until this code is revisited.

EduardOstrovsky: Can't use Assert.IsTrue in this context.
Can only use:
Debug.Assert(BitConverter.IsLittleEndian, "GUID conversion is only little endian for now");
Is this fine to have? Otherwise if bond doesn't support BigEndian than more should break than just this so we may be able to do without the check altogether

cs/test/core/Core.csproj
EduardOstrovsky: I don't know why this keeps happening, attempting to fix it again

chwarr:
Something's different in the Mono implementation that's causing the serialization/deserialization tests to fail. More digging is needed, but that won't happen until tomorrow at the earliest.

chwarr:
Still not sure what's going on with the Mono builds. I _think_ I was able to reproduce the problem without your changes (just building commit 39a8dbf5ad276649f7a52c28e8e41fa955961107), but CI for that change built fine on its own...

cs/src/core/Bond.cs
```diff
@@ -3,7 +3,39 @@
 
 namespace Bond
 {
-    using System.Collections.Generic;
+    using System;
+    using System.Collections.Generic;
+
+    public partial class GUID
```
lalo: we have carriage return in this file too, you may have misconfigured your git check this out https://help.github.com/articles/dealing-with-line-endings/
or you could replace the extra line endings with powershelgl
https://github.com/PowerShell/Win32-OpenSSH/wiki/Dos2Unix---Text-file-format-converters

chwarr:
We've figure out the Mono test failures. @lalo is working on a fix. Details in #154.

chwarr:
Closing & re-opening to re-merge & re-run CI builds after the fix for #154.

Issue #70, opened 2015-07-13
CompactBinary Protocol V2 doesn't work in C#


kylemc:
I updated the Serialization sample to use CBV2 and now it throws a runtime exception. I've seen similar errors with other IDL, but it appears there's some amount of corruption going on with V2.

Code (program.cs):

    var writer = new CompactBinaryWriter<OutputBuffer>(output, 2 /* used to be 1 */);

    // The first calls to Serialize.To and Deserialize<T>.From can take
    // a relatively long time because they generate the de/serializer 
    // for a given type and protocol.
    Serialize.To(writer, src);

    var input = new InputBuffer(output.Data);
    var reader = new CompactBinaryReader<InputBuffer>(input, 2 /* used to be 1 */);

Exception:

> System.IO.InvalidDataException was unhandled
  HResult=-2146233087
  Message=Invalid type BT_UINT64, expected BT_STRING
  Source=Bond
  StackTrace:
       at Bond.Expressions.ThrowExpression.ThrowInvalidTypeException(BondDataType expectedType, BondDataType actualType)
       at lambda_method(Closure , CompactBinaryReader`1 )
       at Bond.Deserializer`1.Deserialize[T](R reader)
       at Bond.Deserialize`1.From[R](R reader)
       at Examples.Program.Main() in c:\Users\kylemc\Documents\Visual Studio 2013\Projects\serialization\program.cs:line 29
       at System.AppDomain._nExecuteAssembly(RuntimeAssembly assembly, String[] args)
       at System.AppDomain.ExecuteAssembly(String assemblyFile, Evidence assemblySecurity, String[] args)
       at Microsoft.VisualStudio.HostingProcess.HostProc.RunUsersAssembly()
       at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
       at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
       at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
       at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
       at System.Threading.ThreadHelper.ThreadStart()
  InnerException:

sapek:
Yes, CBv2 writer is not support in C# at the moment. Reader is supported.

kylemc:
When do you think the CBv2 writer will be supported?

[...]

PR #146, opened 2016-04-08
C# Compact Binary v2 serialization

jdubrule:
Implement Compact Binary v2 serialization support in C#, which previously only supported v2 deserialization (and both for v1)


lalo:
When merged, this would close #70

chwarr:
Can you please rebase the commits to get rid of the work-in-progress and fix up commits? It seems like we should be able to "tell the story" of this change in one to three commits.

cs/src/core/io/IntegerHelper.cs
```diff
@@ -15,6 +15,22 @@ internal static class IntegerHelper
         public const int MaxBytesVarInt32 = 5;
         public const int MaxBytesVarInt64 = 10;
 
+        public static int GetVarUInt16Length(ushort value)
```
chwarr: Did you consider basing these on the ranges. E.g.,

```
if (value < 0x80)
{
    return 1;
}
else if (value < 0x4000)
{
    return 2;
}
else
{
    return 3;
}
```

And so on?

Here's a list of powers of 128, if you were wondering...

```
128
16384
2097152
268435456
34359738368
4398046511104
562949953421312
72057594037927900
9223372036854780000
1180591620717410000000
```

cs/src/core/io/IntegerHelper.cs
```diff
@@ -65,6 +111,65 @@ public static int EncodeVarUInt32(byte[] data, uint value, int index)
...
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            return len;
+        }
```
jasonzio:
```c#
public static int GetVarUInt64Length(ulong value)  
{  
    int len = 1;
    while (value >= 0x80) {
        value >>= 7;
        len++;
    }
    return len;
}
```
In C#, when right-shifting uint or ulong, a logical-shift is performed, zero-filling the high bits.

cs/src/core/io/IntegerHelper.cs
```diff
@@ -34,6 +50,36 @@ public static int EncodeVarUInt16(byte[] data, ushort value, int index)
...
+                        value >>= 7;
+                        // byte 3
+                        if (value >= 0x80)
+                        {
+                            len++;
+                        }
+                    }
+                }
+            }
+            return len;
```
jasonzio: See comment on `GetVarUInt64Length()` below.

cs/src/core/protocols/CompactBinary.cs
```diff
@@ -148,6 +155,27 @@ public CompactBinaryWriter(O output, ushort version = 1)
         {
             this.output = output;
             this.version = version;
+            if (version == 2)
+            {
+                this.lengths = new List<uint>();
+                this.firstPassWriter = new CompactBinaryCounter(lengths);
+#if DEBUG
```
chwarr: I think that moving the lengthCheck code (other than the member variable) into conditional methods will result in cleaner code.

```
public CompactBinaryWriter(O output, ushort version = 1)
{
    this.output = output;
    this.version = version;
    if (version == 2)
    {
        this.lengths = new List<uint>();
        this.firstPassWriter = new CompactBinaryCounter(lengths);
    }
    else
    {
        this.lengths = null;
        this.firstPassWriter = null;
    }

    InitLengthCheck(version);
}

[Conditional(DEBUG)]
void InitLengthCheck(int version)
{
    if (version == 2)
    {
        this.lengthCheck = new Stack<long>();
    }
   ...
}
```
and similarly for `PushLengthCheck` and `PopLengthCheck`.

cs/src/core/io/IntegerHelper.cs
jdubrule: Going with Chris' suggestion of a series of 'if's.

cs/src/attributes/Attributes.cs
```diff
@@ -199,4 +199,18 @@ public SerializerAttribute(Type type)
 
         internal Type Type { get; private set; }
     }
+
+    /// <summary>
+    /// Applied to 2-pass protocol writers to indicate the implementation of IProtocolWriter used to generate the first-pass serializer
```
chwarr: Minor: `see` around `IProtocolWriter`, please.

cs/src/core/Serializer.cs
```diff
@@ -92,7 +99,52 @@ public Serializer(Type type, bool inlineNested)
         /// </remarks>
         public void Serialize(object obj, W writer)
         {
-            serialize[0](obj, writer);
+            helper.Serialize(obj, writer);
+        }
+
+        SerializerHelper helper;
```
chwarr: Minor: can we put `helper` up at the top of the class where `serialize` used to be.

chwarr: Minor: looks like `helper` can be `readonly`.

cs/src/core/Serializer.cs
```diff
@@ -76,10 +75,18 @@ public class Serializer<W>
...
             var parser = new ObjectParser(type);
-            serialize = SerializerGeneratorFactory<object, W>.Create(
-                    (o, w, i) => serialize[i](o, w), type, inlineNested)
-                .Generate(parser)
-                .Select(lambda => lambda.Compile()).ToArray();
+
+            var firstPassAttribute = typeof(W).GetAttribute<FirstPassWriterAttribute>();
+            if (firstPassAttribute != null)
+            {
+                Type firstPassHelperType = typeof(TwoPassSerializerHelper<>).MakeGenericType(typeof(W), firstPassAttribute.Type);
```
chwarr: We should check that `W` implements `ITwoPassProtocolWriter` here so that we can fail fast with a good error message. Throwing `ArgumentException` seems reasonable to me. If we don't do this, then we'll get an `InvalidCastException` from `TwoPassSerializerHelper.Serialize` down the line.

cs/src/core/Transcoder.cs
```diff
@@ -93,16 +110,76 @@ public Transcoder()
         /// <param name="writer">Writer instance</param>
         public void Transcode(R reader, W writer)
         {
-            transcode[0](reader, writer);
+            helper.Transcode(reader, writer);
         }
 
-        Action<R, W>[] Generate<S>(S schema)
+        TranscoderHelper helper;
```
chwarr: Minor: move to top of class and make readonly if possible.

chwarr:
Heads up that this change and #153 (or whatever number it comes back as) are very likely to conflict with each other. I'll merge them in the order that the get completed in, so consider this an incentive to respond to feedback quickly. :-)

cs/src/core/Transcoder.cs
```diff
@@ -61,15 +63,20 @@ static class Cache<R, W>
...
         /// <summary>
         /// Create a transcoder for payloads with specified runtime schema
         /// </summary>
         /// <param name="schema">Payload schema, required for transcoding from untagged protocols</param>
         public Transcoder(RuntimeSchema schema)
         {
-            transcode = Generate(schema);
+            if (typeof(W).GetAttribute<FirstPassWriterAttribute>() != null)
+            {
+                helper = new TwoPassTranscoderHelper<CompactBinaryCounter>(schema);
```
chwarr: Shouldn't the type here depend on the value of `FirstPassWriterAttribute.Type` instead of being hard-coded to `CompactBinaryCounter`?

chwarr:
I'm trying an experiment. When you make changes to address PR feedback, please add them as a new commit. I'll see how difficult rebasing to meld is when I merge.

Adding them as a new commit lets me easily see the differences before and after the comments were addressed.

cs/src/core/Transcoder.cs
```diff
@@ -93,16 +110,76 @@ public Transcoder()
...
+            {
+                transcode = Generate(type);
+            }
+
+            virtual public void Transcode(R reader, W writer)
+            {
+                transcode[0](reader, writer);
+            }
+
+            Action<R, W>[] Generate<S>(S schema)
```
chwarr: Minor: explicit protection level, please.

cs/src/core/Transcoder.cs
```diff
@@ -93,16 +110,76 @@ public Transcoder()
...
+            {
+                firstPassTranscode = GenerateFirstPass(type);
+            }
+
+            public override void Transcode(R reader, W writer)
+            {
+                var firstPassWriter = ((ITwoPassProtocolWriter)writer).GetFirstPassWriter();
+                if (firstPassWriter != null)
+                {
+                    R clonedReader = ((ICloneable<R>)reader).Clone();
```
chwarr: Not all readers are clonable. This should fail more gracefully when they aren't. We'll also want to check that the writer actually implemented `ITwoPassProtocolWriter` in the ctor like in `Serializer`.

cs/src/core/Transcoder.cs
```diff
@@ -93,16 +110,76 @@ public Transcoder()
...
+                    firstPassTranscode[0](clonedReader, (FPW)firstPassWriter);
+                }
+
+                base.Transcode(reader, writer);
+            }
+
+            Action<R, FPW>[] GenerateFirstPass<S>(S schema)
```
chwarr: Minor: explicit protection level, please.

cs/src/core/protocols/CompactBinary.cs
```diff
@@ -125,19 +125,25 @@ namespace Bond.Protocols
     using System.IO;
     using System.Runtime.CompilerServices;
     using System.Text;
+    using System.Collections.Generic;
     using Bond.IO;
+    using System.Diagnostics;
```
chwarr: Minor: we try to keep these sorted.

cs/src/core/protocols/CompactBinaryCounter.cs
```diff
@@ -0,0 +1,390 @@
...
+
+        /// <summary>
+        /// Write protocol magic number and version
+        /// </summary>
+#if NET45
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+#endif
+        public void WriteVersion()
+        {
+//             output.WriteUInt16(Magic);
```
chwarr: Minor: removed commented out code

cs/src/core/protocols/CompactBinary.cs
```diff
@@ -171,7 +195,16 @@ public void WriteVersion()
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
 #endif
         public void WriteStructBegin(Metadata metadata)
-        {}
+        {
+            if (version == 2)
+            {
+                uint length = lengths[0];
+                lengths.RemoveAt(0);
```
chwarr: `RemoveAt(0)` is going to slide all the lengths down taking O(n) operations, for a total complexity of O(n^2). Did you look at using a different data structure or an index?

cs/src/attributes/Attributes.cs
jdubrule: Not sure I understand.

cs/src/core/Serializer.cs
```diff
@@ -76,10 +77,23 @@ public class Serializer<W>
...
+                }
+
+                Type firstPassHelperType = typeof(TwoPassSerializerHelper<>).MakeGenericType(typeof(W), firstPassAttribute.Type);
+
+                helper = (SerializerHelper)Activator.CreateInstance(firstPassHelperType, parser, type, inlineNested);
+            }
+            else
+            {
+                helper = new SerializerHelper(parser, type, inlineNested);
+            }
```
alexsatch: 1. The logic from line 80 to line 96 just wants to be extracted into its own method; and it suspiciously looks very similar with the one in Transcoder.  
2. Also, it should be called only once per type `W` - please push it to the static constructor of Serializer<W>, especially since it calls Activator.CreateInstance

cs/src/attributes/Attributes.cs
alexsatch: I think he meant to replace `IProtocolWriter` by xml tag `<see cref="IProtocolWriter" />` to improve documentation.

cs/src/core/Serializer.cs
jdubrule: 1.  Not sure how to cleanly merge Serializer&Transcoder, 'cause they use different helper types, have different # of generic parameters, and their constructors take different args.
2.  Done.

cs/src/core/protocols/CompactBinaryCounter.cs
```diff
@@ -0,0 +1,390 @@
...
+    using System.IO;
+    using System.Runtime.CompilerServices;
+    using System.Text;
+    using System.Collections.Generic;
+    using Bond.IO;
+
+    /// <summary>
+    /// Length-calculator for Bond CompactBinary protocol V2
+    /// </summary>
+    ///
```
chwarr: Minor: odd blank line 18

cs/src/core/protocols/CompactBinaryCounter.cs
```diff
@@ -0,0 +1,390 @@
...
+    [Reader(typeof(CompactBinaryReader<>))]
+    public struct CompactBinaryCounter : IProtocolWriter
+    {
+        private class CounterStackFrame
+        {
+            public Int32 lengthSlot;
+            public int currentLength = 0;
```
chwarr: Minor: `= 0` is implied so can be omitted.

cs/src/core/protocols/CompactBinaryCounter.cs
```diff
@@ -0,0 +1,390 @@
...
+        /// <summary>
+        /// End writing a struct
+        /// </summary>
+#if NET45
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+#endif
+        public void WriteStructEnd()
+        {
+            CounterStackFrame frame = counterStack.Peek();
+            uint structLength = (uint)frame.currentLength + 1;
```
chwarr: Can we document with a comment why there's the `+ 1`?

cs/src/core/Transcoder.cs
```diff
@@ -93,16 +124,79 @@ public Transcoder()
...
+            {
+                firstPassTranscode = GenerateFirstPass(type);
+            }
+
+            public override void Transcode(R reader, W writer)
+            {
+                var firstPassWriter = ((ITwoPassProtocolWriter)writer).GetFirstPassWriter();
+                if (firstPassWriter != null)
+                {
+                    if (!typeof(ICloneable<R>).IsAssignableFrom(typeof(R)))
```
chwarr: Can we do this check in the static ctor?

cs/src/core/protocols/CompactBinaryCounter.cs
```diff
@@ -0,0 +1,388 @@
...
+    /// Length-calculator for Bond CompactBinary protocol V2
+    /// </summary>
+    ///
+    [Reader(typeof(CompactBinaryReader<>))]
+    public struct CompactBinaryCounter : IProtocolWriter
+    {
+        private class CounterStackFrame
+        {
+            public LinkedListNode<UInt32> lengthSlot;
+            public int currentLength = 0;
```
chwarr: - [ ] Minor: ` = 0` is implied, so it can be omitted.

chwarr:
Looking good. Just a few minor things left. Aside from the line-level comments elsewhere,

- [ ] remove Parser.hs from review

cs/src/core/Transcoder.cs
jdubrule: We don't know whether the CompactBinaryWriter passed in for W is using v1 or v2 until we get to here, so we may not need R to be ICloneable.

chwarr:
This PR now conflicts with the head of master after I merged #153. Can you rebase atop the new head of master (56b5914a5bb41178521e01f6ce078d429e3e6b71)?

jdubrule:
OK; I think it's cleaned up & ready to go.

-jeff

chwarr:
Looking...

cs/src/core/Transcoder.cs
```diff
@@ -164,20 +165,25 @@ virtual public void Transcode(R reader, W writer)
...
 
-        private class TwoPassTranscoderHelper<FPW> : TranscoderHelper
+        class TwoPassTranscoderHelper<FPW> : TranscoderHelper
         {
-            readonly Action<R, FPW>[] firstPassTranscode;
+            readonly Lazy<Action<R, FPW>[]> firstPassTranscode;
 
             public TwoPassTranscoderHelper(RuntimeSchema schema, IParser parser):
-                base(schema, null)
+                base(schema, parser)
```
chwarr: This change looks like it's addressing a bug. Did the unit tests catch the bug before this fix? If not, we'll want to add a regression test.

chwarr: This change looks like it's addressing a bug. Did the unit tests catch the bug before this fix? If not, we'll want to add a regression test.

cs/src/attributes/Attributes.cs
```diff
@@ -199,4 +199,18 @@ public SerializerAttribute(Type type)
 
         internal Type Type { get; private set; }
     }
+
+    /// <summary>
+    /// Applied to 2-pass protocol writers to indicate the implementation of <see cref="IProtocolWriter" /> used to generate the first-pass serializer
```
chwarr: I know I asked for `see` to be added in comments--and thank you for doing so--but in this comment in particular we can't use `see`, as Bond.Attributes.dll doesn't refer to Bond.dll, and we don't want it to. This particular `see` should be removed and just left as "IProtocolWriter".

chwarr:
Travis CI failure is a build timeout, after 45 minutes, which we occasionally see.

cs/test/compat/Program.cs
```diff
@@ -106,6 +106,11 @@ static void Write<R>(R reader, OutputStream output, string toProtocol)
                 var writer = new CompactBinaryWriter<OutputStream>(output);
                 Serialize.To(writer, Deserialize<Compat>.From(reader));
             }
+            else if (toProtocol == "compact2")
+            {
+                var writer = new CompactBinaryWriter<OutputStream>(output, 2);
+                Serialize.To(writer, Deserialize<Compat>.From(reader));
+            }
```
sapek: I think this is not used unless you enable it [here](https://github.com/Microsoft/bond/blob/eb19aef894cf9c884654c14fdae0b65fdc1c0aea/cpp/test/compat/compat.cmake#L18-L21).

chwarr: Yup, I've taken care of this as part of the merge.

https://github.com/Microsoft/bond/commit/5bbe73c7b19f1969483d4743b4fa4b6ab51e8451

chwarr:
Merged with some other changes in commit e80c126ba9d44632e5069c71694d33f890c235dc.

Jeff, thanks again for your contribution!

PR #172, opened 2016-06-02
Allow creation of user-defined TypeMapping's.

rblaze:
I need one to create Haskell-stype qualified names, but can't create it, and have to reimplement several functions.

### DIFF
--- a/compiler/src/Language/Bond/Codegen/TypeMapping.hs
+++ b/compiler/src/Language/Bond/Codegen/TypeMapping.hs
@@ -17,7 +17,8 @@ type system of a target programming language.
 module Language.Bond.Codegen.TypeMapping
     ( -- * Mapping context
       MappingContext(..)
-    , TypeMapping
+    , TypeMapping(..)
+    , TypeNameBuilder
       -- * Type mappings
     , idlTypeMapping
     , cppTypeMapping
@@ -48,6 +49,11 @@ module Language.Bond.Codegen.TypeMapping
     , getNamespace
     , getDeclNamespace
     , customAliasMapping
+      -- * TypeMapping helper functions
+    , elementTypeName
+    , aliasTypeName
+    , declTypeName
+    , declQualifiedTypeName
     ) where
 
 import Data.List
@@ -74,7 +80,7 @@ data MappingContext = MappingContext
     , namespaces :: [Namespace]
     }
 
--- | An opaque type representing a type mapping.
+-- | A type representing a type mapping.
 data TypeMapping = TypeMapping
     { language :: Maybe Language
     , global :: Builder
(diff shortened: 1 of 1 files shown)


lalo:
We would just have to bump the GBC version number to reflect the change or bundle it on the next release which already a different version number. The failed travis iteration is not related to this PR.

rblaze:
As a side note, it would be great to have next release, also because of aeson version bump. Currently bond can't be built with stackage lts-6 because of it.

lalo:
@rblaze by Aeson bump you mean the one that is already on master right? https://github.com/Microsoft/bond/commit/0fead8d8b6bb26e8d682a430f19a88768ac9a039

rblaze:
Exactly, 0.12 works just fine.

compiler/src/Language/Bond/Codegen/TypeMapping.hs
```diff
@@ -17,7 +17,7 @@ type system of a target programming language.
 module Language.Bond.Codegen.TypeMapping
     ( -- * Mapping context
       MappingContext(..)
-    , TypeMapping
+    , TypeMapping(..)
```
chwarr: `TypeMapping` also refers to `TypeNameBuilder`. While we don't need to export that from the module because its just a type alias, doing so would allow you to use `TypeNameBuilder` instead of `Reader MappingContext Builder`.

chwarr: Comment for `TypeMapping` needs to be updated. It's no longer opaque.

sapek: It is also worth considering if we shouldn't export some of the helpers that are usually needed to create a `TypeMapping`. In particular I'm thinking about `elementTypeName`, `instanceTypeName`, `annotatedTypeName`, `aliasTypeName`, `declTypeName` and `declQualifiedTypeName`.

rblaze: Changed this, pushed updated commit.

lalo:
I think it looks good, the only nit I see is the new uncommented functions (elementTypeName, instanceTypeName, annotatedTypeName, aliasTypeName, declTypeName and declQualifiedTypeName) since I noticed a pattern that if its exposed it has a small comment.

rblaze:
Could you provide me with descriptions to add? I'm a bit struggling to come up with good ones.

sapek:
How about something along these lines:

```
-- | Builder for nested element types (e.g. list elements) in context of 'TypeNameBuilder' monad. 
-- Used to implement 'mapType' function of 'TypeMapping'.
elementTypeName :: Type -> TypeNameBuilder

-- | Builder for the qualified name for a 'Declaration' in context of 'TypeNameBuilder' monad.
-- Used to implement 'mapType' function of 'TypeMapping'.
declQualifiedTypeName :: Declaration -> TypeNameBuilder
```

Also I think we actually don't need to export `instanceTypeName` and `annotatedTypeName`. They are never used to build type mappings.

rblaze:
Updated pull request.

chwarr:
LGTM. I'll merge this tomorrow.

chwarr:
Version 0.5.0.0 of the bond package is now live on Hackage. That version includes this change.

Issue #167, opened 2016-06-02
Improving the C++ nullable implementation


glenfe:
1. [x] **1.** When `BOND_NO_CXX11_RVALUE_REFERENCES`  is not defined, and `arg1` is a forwarding-reference (instead of a const l-value reference) then construction of `T` should use `std::forward<Arg1>(arg1)` instead of `arg1`.
   
   For example:
   
   ```
   template<typename Arg1>
   pointer new_value(detail::no_allocator&, Arg1&& arg1)
   {
       return new T(std::forward<Arg1>(arg1));
   }
   ```
2. [x] **2.** In `destroy()` when the C++11 allocator model is available (e.g. `BOOST_NO_CXX11_ALLOCATOR` is not defined) then do not use `alloc.destroy(_value)` but instead use `std::allocator_traits::destroy` because a valid C++11 allocator is not required to have a `destroy` member.
3. [x] **3.** Either make `new_value()` use `std::allocator_traits::construct` (in C++11 - or the allocator's `construct` in C++03) or make `destroy()` invoke the destructor instead of calling `destroy`. i.e. Be consistent: If placement new is used, then manually invoke the destuctor. If allocator's `construct` is used, then use allocator's `destroy`.
4. [x] **4.** Currently `nullable` is confined to custom allocators whose `pointer` type is a raw pointer type compatible with `T*`. In the C++11 allocator model, this is not required to be true: i.e. you can have valid custom allocators whose `pointer` type is a kind of smart pointer. So the `pointer` member type of `nullable` has to be inferred from the allocator's `pointer` member type (via `std::allocator_traits::pointer` in C++11).
5. [x] **5.** It would be better for the first specialization of `nullable` to store an object of type `std::aligned_storage_t<sizeof(value_type), alignof(value_type)>` (instead of storing an object of type `value_type`) to have a better `make_value()` implementation. i.e. To not be reliant on compiler optimizations like RVO to elide any potential copy or moves and construct that `value_type` in-place with the desired arguments. (Of course `boost::aligned_storage` and `boost::alignment_of` can be used instead for compatibility with older compilers).
[...]

PR #174, opened 2016-06-06
Improvements and fixes for C++ nullable implementation and reduce dynamic allocation 

glenfe:
Covers points 1, 2, 3, 4, and 6 that I raised in issue #167 as well as reduce two additional dynamic allocations.

### DIFF
--- a/cpp/inc/bond/core/config.h
+++ b/cpp/inc/bond/core/config.h
@@ -57,7 +57,7 @@
 #define BOND_NO_CX11_HDR_MUTEX
 #endif
 
-#if defined(BOOST_NO_CXX11_ALLOCATOR)
+#if defined(BOOST_NO_CXX11_ALLOCATOR) || defined(_LIBCPP_HAS_NO_TEMPLATE_ALIASES)
 #define BOND_NO_CXX11_ALLOCATOR
 #endif
 
--- a/cpp/inc/bond/core/nullable.h
+++ b/cpp/inc/bond/core/nullable.h
@@ -494,18 +494,14 @@ private:
         delete _value;
     }
 
-    template<typename AllocatorT, typename Arg1>
 #ifndef BOND_NO_CXX11_RVALUE_REFERENCES
+    template<typename AllocatorT, typename Arg1>
     pointer new_value(AllocatorT& alloc, Arg1&& arg1)
-#else
-    pointer new_value(AllocatorT& alloc, const Arg1& arg1)
-#endif
     {
         T* p = alloc.allocate(1);
         try
         {
-            void* p1 = p;
-            return ::new(p1) T(arg1);
+            return ::new(static_cast<void*>(p)) T(std::forward<Arg1>(arg1));
         }
         catch (...)
         {
@@ -514,20 +510,38 @@ private:
         }
     }
 
+    template<typename Arg1>
+    pointer new_value(detail::no_allocator&, Arg1&& arg1)
+    {
+        return new T(std::forward<Arg1>(arg1));
+    }
+#else
+    template<typename AllocatorT, typename Arg1>
+    pointer new_value(AllocatorT& alloc, const Arg1& arg1)
+    {
+        T* p = alloc.allocate(1);
+        try
+        {
+            return ::new(static_cast<void*>(p)) T(arg1);
+        }
+        catch (...)
+        {
+            alloc.deallocate(p, 1);
+            throw;
+        }
+    }
+
+    template<typename Arg1>
+    pointer new_value(detail::no_allocator&, const Arg1& arg1)
+    {
+        return new T(arg1);
+    }
+#endif
+
     template<typename AllocatorT>
     pointer new_value(AllocatorT& alloc)
     {
         return new_value(alloc, alloc);
-    }
-
-    template<typename Arg1>
-#ifndef BOND_NO_CXX11_RVALUE_REFERENCES
-    pointer new_value(detail::no_allocator&, Arg1&& arg1)
-#else
-    pointer new_value(detail::no_allocator&, const Arg1& arg1)
-#endif
-    {
-        return new T(arg1);
     }
 
     pointer new_value(detail::no_allocator&)
--- a/cpp/inc/bond/core/nullable.h
+++ b/cpp/inc/bond/core/nullable.h
@@ -84,6 +84,7 @@ class nullable;
 
 template<typename T, typename Allocator>
 class nullable<T, Allocator, true>
+    : Allocator
 {
 public:
     typedef T           value_type;
@@ -105,7 +106,7 @@ public:
 
     void swap(nullable& src)
     {
-        std::swap(_alloc, src._alloc);
+        std::swap(base(), src.base());
         std::swap(_hasvalue, src._hasvalue);
         std::swap(_value, src._value);
     }
@@ -117,7 +118,7 @@ public:
 
     explicit
     nullable(const allocator_type& alloc)
-        : _alloc(alloc),
+        : Allocator(alloc),
           _value(make_value<value_type>()),
           _hasvalue(false)
     {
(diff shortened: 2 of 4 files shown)


glenfe:
Looks like your AppleClang test envirnoment is using a non-conforming `allocator_traits` implementation. Most likely the variant present when `_LIBCPP_HAS_NO_TEMPLATE_ALIASES` is defined.

glenfe:
The space saving is actually more than just 1 byte. That is, the empty base optimization there actually saves `alignof(value_type)` bytes in the case of the first specialization, and saves `alignof(pointer)` bytes in the case of the second specialization. e.g. This could be up to 8 bytes on 64-bit, or 4 bytes on 32-bit for certain `nullable<T>` types.

cpp/inc/bond/core/nullable.h
```diff
@@ -117,7 +118,7 @@ class nullable<T, Allocator, true>
 
     explicit
     nullable(const allocator_type& alloc)
-        : _alloc(alloc),
+        : allocator_type(alloc),
```
chwarr: I think this would be clearer if we used `Allocator` here, as that's the name of the base.

cpp/inc/bond/core/config.h
```diff
@@ -57,7 +57,7 @@
 #define BOND_NO_CX11_HDR_MUTEX
 #endif
 
-#if defined(BOOST_NO_CXX11_ALLOCATOR)
+#if defined(BOOST_NO_CXX11_ALLOCATOR) || defined(_LIBCPP_HAS_NO_TEMPLATE_ALIASES)
```
chwarr: `_LIBCPP_HAS_NO_TEMPLATE_ALIASES` seems redundant. Is this for really old versions of Boost?

cpp/inc/bond/core/nullable.h
```diff
@@ -501,7 +505,13 @@ class nullable<T, Allocator, false>
         T* p = alloc.allocate(1);
         try
         {
-            return ::new(static_cast<void*>(p)) T(std::forward<Arg1>(arg1));
+#ifndef BOND_NO_CXX11_ALLOCATOR
+            std::allocator_traits<AllocatorT>::construct(alloc, p,
```
chwarr: Minor: let's put the arguments either all on the same line or all on different lines.

cpp/inc/bond/core/nullable.h
```diff
@@ -482,32 +474,37 @@ class nullable<T, Allocator, false>
...
-        std::allocator_traits<AllocatorT>::destroy(alloc, _value);
+    typedef typename std::allocator_traits<allocator_type>::
+        template rebind_alloc<value_type> rebind_alloc;
+
+    typedef typename std::allocator_traits<rebind_alloc>::pointer real_pointer;
 #else
-        _value->~T();
+    typedef typename allocator_type::template rebind<value_type>::other rebind_alloc;
+
+    typedef pointer real_pointer;
```
chwarr: Above we get `real_pointer` from the traits of `rebind_alloc`, but here we get it from ourselves (where it is `T*`) instead of from `rebind_alloc`. Intentional difference?

cpp/inc/bond/core/nullable.h
```diff
@@ -435,7 +427,7 @@ class nullable<T, Allocator, false>
     {
         if (_value)
         {
-            destroy(_alloc);
+            delete_value();
             _value = 0;
```
chwarr: Does this assignment to zero still make sense if, say, custom smart pointer types are being employed by the allocator?

cpp/inc/bond/core/nullable.h
```diff
@@ -482,32 +474,37 @@ class nullable<T, Allocator, false>
...
     {
-        T* p = alloc.allocate(1);
+        real_pointer p(alloc.allocate(1));
         try
         {
 #ifndef BOND_NO_CXX11_ALLOCATOR
-            std::allocator_traits<AllocatorT>::construct(alloc, p,
-                std::forward<Arg1>(arg1));
+            std::allocator_traits<AllocatorT>::construct(alloc,
+                boost::addressof(*p), std::forward<Arg1>(arg1));
```
chwarr: Why is `addressof` needed here? It seems wrong if there's a custom pointer type in play. Don't we need to construct where it tells us to construct via its overloaded `operator&`?

chwarr:
Left some comments. Be sure to look at the comments on outdated diffs, because GitHub doesn't understand that reviewing individual commits is a thing. :-(

cpp/inc/bond/core/config.h
glenfe: This is for versions of Boost before 1.56.0.

cpp/inc/bond/core/nullable.h
glenfe: Sure.

cpp/inc/bond/core/nullable.h
glenfe: Intentional.

cpp/inc/bond/core/nullable.h
glenfe: Even those smart pointer types must satisfy the nullable pointer concept. Ideally, `nullptr` would be used here if C++03 compatibility was not required, but I'll use default construction which is equally readable.

cpp/inc/bond/core/nullable.h
glenfe: Sure.

cpp/inc/bond/core/nullable.h
glenfe: Sure.

cpp/inc/bond/core/nullable.h
glenfe: No; you're confusing the `pointer` type (which can be a smart or raw pointer type, and which is a property of the allocator type) with the type of `&std::declval<value_type>()` (which could be an entirely different kind of smart or raw pointer type, and which is a property of the value type, independent of what allocator type is in play).

glenfe:
If you don't want to suppress that warning about `static_cast<bool>(pointer)` then that expression can instead be written as `!!pointer`.

jasonzio:
I'm not a big fan of disabling warnings. If you chose to do so, please wrap it tightly around the offending line of code; don't do it globally for the entire file or, worse yet, the entire project. Warnings really do help catch problems; a tightly-wrapped deviation (disable, then the offending line of code, then re-enable) shows the warning was investigated, determined to be benign, and silenced.

glenfe:
I'm a big fan of disabling insensible diagnostics, even globally.

cpp/inc/bond/core/config.h
chwarr: Figured. Thanks for the confirmation.

glenfe:
I'm a big fan of disabling stupid warnings, especially globally. It shows a sensible intolerance to a defective C++ implementation.

cpp/inc/bond/core/nullable.h
chwarr: Okay. I think I see it now. The `*p` in the argument to `boost::addressof()` might invoke the `operator*` of some smart pointer class. That's what we want: otherwise how would we get a reference to the instance of `value_type` we're interested in out of the smart pointer? But then, we really do want to construct in the storage location for said instance regardless of what `value_type::operator&` does.

glenfe: You're on the right track. The type of `p` could be some smart pointer `X`, but `X` does not have to be convertible to type `U*` (for template parameter `U`) which is what `construct` takes as of C++11. The type of `&*p` could be some smart pointer `Y`, but `Y` also does not have to be convertible to the aforementioned `U*`.

The idiomatic thing, which is what I do here, which is also what C++ standard library allocator-aware containers do, which is also what you'll find in any correct allocator-aware code (e.g. in Boost.Container) is: Construct in the true address obtained by `addressof(*p)`.

A similar rationale governs the use of `::new((void*)p) Type...` instead of just `new(p) Type...`. Even if there exists a placement new overload that the latter would select, we don't care, in favor of choosing the most true placement new (for `void*` and in the global scope).

chwarr:
`!!pointer` seems like a good compromise.

chwarr:
Confirmed size reduction of `bond::nullable<std::string>` a pointer size (8 bytes with Visual Studio 14 2015 Win64 and 4 bytes with Visual Studio 14 2015). :tada:

I'm going to run some x86 builds myself, as these aren't part of the CI builds.

chwarr:
Bond version 4.3.0 is now live on NuGet.org. These changes are included in that version. You'll also need to pull down the latest C++ source code as well.

PR #175, opened 2016-06-09
Make accept failures in EpoxyListener non-fatal

chwarr:
We anticipate that sometimes we won't be able to accept a connection, so
such a case should be logged as a regular error instead of a fatal (so
take the process down) error.

### DIFF
--- a/cs/src/comm/epoxy-transport/EpoxyListener.cs
+++ b/cs/src/comm/epoxy-transport/EpoxyListener.cs
@@ -114,7 +114,7 @@ namespace Bond.Comm.Epoxy
                 }
                 catch (SocketException ex)
                 {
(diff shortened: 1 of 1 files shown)


chwarr:
AppVeyor builds failed while re-building the documentation. When the documentation build is run, we attempt to fetch `https://img.shields.io/nuget/v/Bond.Core.CSharp.svg?style=flat`, and shields.io was down for a bit last week.

This is because we're attempting to build self-contained pages in some places where it doesn't make sense. E.g., the embedded NuGet version images should be fetched each time the page is rendered, not each time it is generated. A fix for this in in pull request [#179](https://github.com/Microsoft/bond/pull/179).

Issue #176, opened 2016-06-10
ghc 7.8.4 OS X Travis CI builds are failing


chwarr:
Our OS X builds using [ghc 7.8.4](https://github.com/Microsoft/bond/blob/dde93cb1621478949f2924761e22c73ee608de24/.travis.yml#L41) have been failing with [timeouts recently](https://travis-ci.org/Microsoft/bond/jobs/136315467). ghc 7.6.3 and 7.4.1 are passing, however.

lalo:
https://travis-ci.org/Microsoft/bond/builds/137432412

Got a successful run by just increasing the timeout.

chwarr:
Good to know that the build isn't stalled. It's a bit interesting that the different ghc versions have such different build times. 7.6.3 takes 4 minutes 23 seconds, but 7.8.4 takes 36 minutes 17 seconds.

I'm hesitant to just up the limit without understanding where the 700% increase in time comes from.

lalo:
7.10.3 also timeouts
https://travis-ci.org/Microsoft/bond/builds/137610412

something fisghcy is going on

PR #180, opened 2016-06-13
Add info if nullable or blob to RuntimeSchemaDef

### LINKED ISSUE #161: The type of nullable<T> and vector<T> fields are indistinguishable using only a SchemaDef
We work with some schemas which we have a SchemaDef for, but not the .bond file (or the corresponding C# class) used to originally create the SchemaDef. Any fields with a type of `nullable<T>` in the original .bond file show up as a `BT_LIST` in the SchemaDef.

For example, here is a struct defined in a .bond file:

```
struct NullableAndVector
{
    1: nullable<int64> nullableInt;
    2: vector<int64> vectorInt;
}
```

The properties of the resulting C# class match up perfectly:

```
[global::Bond.Id(1), global::Bond.Type(typeof(global::Bond.Tag.nullable<long>))]
public long? nullableInt { get; set; }

[global::Bond.Id(2)]
public List<long> vectorInt { get; set; }
```

However, for both `nullableInt` and `vectorInt`, the SchemaDef (originally obtained using `Schema<NullableAndVector>.RuntimeSchema.SchemaDef`) shows both fields as being a `BT_LIST` with `BT_INT64` elements, and all metadata is identical.

Is there any way to distinguish between `nullable<T>` and `vector<T>` using only information in a SchemaDef? If not, can a change be made to make that possible?

lalo:
- Add ListSubType to TypeDef definition
- Bond data types that are of type BT_LIST will show under ListSubType if the idl was of nullable or blob type. Treat as list if its BT_LIST and UNKNOWN and disregard this field if it is not of type BT_LIST
- Added support in c++/c# and respective tests

Closes #161


compiler/src/Language/Bond/Syntax/SchemaDef.hs
```diff
@@ -63,11 +63,15 @@ data FieldDef =
...
     TypeDef
+          -- Domain of Int is BondDataType
         { id :: Maybe Int
+          -- Index of struct definition in SchemaDef.structs when id == BT_STRUCT
         , struct_def :: Maybe Int
         , element :: Maybe [TypeDef]
         , key :: Maybe [TypeDef]
         , bonded_type :: Maybe Bool
+          -- Domain of Int is ListSubType
+        , list_sub_type :: Maybe Int
```
chwarr: Can you also document this in compiler.md?

cpp/inc/bond/core/bond_const.bond
```diff
@@ -25,6 +25,18 @@ enum BondDataType
     BT_UNAVAILABLE= 127
 }
 
+// Enumerator to distinguish the different subtypes treated as BT_LIST
+enum ListSubType
+{
+    // Not a BT_LIST or info not available to infer the ListSubType
```
chwarr: How about this?

```
    // Not a sub-type of BT_LIST. Used when the field is just a list.
    // Also used when the field is not a BT_LIST or the sub type cannot be determined.
    NONE = 0,
```

chwarr:
CI build failures are known and unrelated to this change: #176.

PR #183, opened 2016-06-19
Add C# example to demonstrate storing Guid as a binary blob.

minghan:
Useful for those who prefer storing Guid in a more compact form.

### DIFF
--- a/examples/cs/core/guid/program.cs
+++ b/examples/cs/core/guid/program.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Diagnostics;
+    using System.IO;
     using Bond;
     using Bond.Protocols;
     using Bond.IO.Unsafe;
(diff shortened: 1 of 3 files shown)


chwarr:
There's a fair amount of boilerplate between this example and the existing GUID one.

Can you incorporate this example into that one instead?

That is, something like this:

```
namespace Examples

using guid = string;
using guid_bin = blob;

struct Example
{
    0: guid id;
    1: guid_bin id_bin;
}
```

minghan:
Thanks @chwarr , good suggestion! I have updated the pull request.

examples/cs/core/guid/program.cs
```diff
@@ -17,6 +19,24 @@ public static string Convert(Guid value, string unused)
...
+        public static Guid Convert(ArraySegment<byte> value, Guid unused)
+        {
+            var bits = new byte[16];
+            Buffer.BlockCopy(value.Array, value.Offset, bits, 0, 16);
+            return new Guid(bits);
+        }
+
+        public static ArraySegment<byte> Convert(Guid value, ArraySegment<byte> unused)
+        {
+            return new ArraySegment<byte>(value.ToByteArray(), 0, 16);
```
chwarr: Can be simplified to `return new ArraySegment<byte>(value.ToByteArray());`

examples/cs/core/guid/program.cs
```diff
@@ -17,6 +19,24 @@ public static string Convert(Guid value, string unused)
...
             return value.ToString();
         }
+
+        #endregion
+
+        #region Guid-blob conversion
+
+        public static Guid Convert(ArraySegment<byte> value, Guid unused)
+        {
+            var bits = new byte[16];
```
chwarr: We can do this without an allocation.

1. Check `value.Count == 16` otherwise throw something like `InvalidDataException`.
1. Extract the bytes themselves and use the `Guid(int, short, short, byte, ..., byte)` constructor:

```
    int a =
          ((int)value.Array[value.Offset + 3] << 24)
        | ((int)value.Array[value.Offset + 2] << 16)
        | ((int)value.Array[value.Offset + 1] <<  8)
        | value.Array[value.Offset + 0];
    short b = /* similar for the next two bytes */
    short c = /* similar for the next two bytes */
    return new Guid(a, b, c,
        value.Array[value.Offset + 8],
        ...
        value.Array[value.Offset + 15]);
```

chwarr:
Thanks for the update. Can you address the comments I've left by adding fixup commits. I'll squash them when I merge.

minghan:
Thanks for the feedback! Added a fixup commit as requested.

examples/cs/core/guid/program.cs
```diff
@@ -26,14 +27,36 @@ public static string Convert(Guid value, string unused)
...
-            var bits = new byte[16];
-            Buffer.BlockCopy(value.Array, value.Offset, bits, 0, 16);
-            return new Guid(bits);
+            if (value.Count != 16)
+            {
+                throw new InvalidDataException("value must be of length 16");
+            }
+
+            byte[] array = value.Array;
+            int offset = value.Offset;
```
chwarr: Much more readable than my proposal. Thanks.

PR #187, opened 2016-06-23
Correctly test for implementing IBonded in runtime-defined schemas ut…

mserd:
…ilizing Bond.Bonded<>


mserd:
Associated issue https://github.com/Microsoft/bond/issues/188

mserd:
This wasn't the cause of my issue.

PR #190, opened 2016-06-27
Scoped enums

jdubrule:
Use bond::EnumType::EnumValue for internal use of Bond enums.  This will make it easier, in the future to provide an option to produce 'enum class EnumType' declarations.

This is a WIP PR to run the check-build, prior to cleanup/rebase/etc.


chwarr:
Can you `@` ping me when this is ready for review?

jdubrule:
I'm still having issues with g++ not accepting bond::EnumName::EnumValue
syntax.

Is that something that's currently part of our unit-tests? The Enumeration
example demos it; is that compiled for g++? I don't know why it wouldn't be
working here, if it works elsewhere.

-jeff

glenfe:
If you're trying to compile it with g++ in C++03 mode, that's probably why. It will compile with  g++ with C++11 or higher (and with clang++ in any mode, because of an extension, but that extension should emit a diagnostic).

chwarr:
I'm seeing the behavior that @glenfe describes when I try toy examples with g++ and clang.

jdubrule:
Yeah; looks like there isn't a reliable way to setup these enums which will
make all the different compilers & C++-flavors happy & let you do both
"MyEnum x = MyEnum::abc" and "MyEnum x = abc".  Or, if there is, it's a
hellish amount of macros or something to make it happen.

The enumerations example works because it switches g++ into C++11 mode.

-jeff

cpp/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+use_cxx11 ()
+
```
sapek: Are we stopping to support C++03 completely?

chwarr: No. We need to keep support C++03 for a while yet.

I think this is just @jdubrule playing around with things to see what works and what doesn't.

jdubrule:
This change is fundamentally incompatible with C++03, and is thus unsuitable for Bond at this time.
I thought we'd done some clever namespace-thing to allow scoped-enum syntax to work in C++03, but that relied on a Microsoft compiler extension, and the enumeration-example which demonstrates the behavior is setup to compile in C++11.

PR #191, opened 2016-06-28
[c++] generate `noexcept` move constructors

chwarr:
* Generated C++ code has conditional `noexcept` move constructors
* Core Bond types' move constructors have been marked as conditional
  `noexcept`
  * There are many places that we'd like to use `T(T&& other) =
    default`, but not all the compilers we target support `= default`
    for `noexcept` move constructors (in particular, [Visual Studio
    2013][1]), so we hand write very similar move constructors.

    We've chosen to explicitly write out things like `Bar(Bar&& other)
    BOND_NOEXCEPT_IF(bond::is_nothrow_move_constructible<Foo>)` instead
    of analyzing `Foo` directly. This should be more maintainable under
    future modifications. Otherwise, future editors of `Foo` would have
    to know to look in `Bar` and adjust the `noexcept` specifications if
    the `noexcept` status of `Foo` were to change.

[1]: https://msdn.microsoft.com/en-us/library/hh567368(v=vs.120).aspx#defaultedanddeleted


glenfe:
If you're doing all this just for MSVC2013 because it doesn't properly support defaulted move constructors, you're getting nothing. i.e. In MSVC2013 `BOOST_NO_CXX11_NOEXCEPT` should be defined, which means your `BOND_NOEXCEPT_IF(anything)` equates to nothing at all.

You're better off not doing this, and using `T(T&&)=default;` for compilers which support it, and for the other compilers, using a cleaner approach than this.

chwarr:
Good point that MSVC2013 doesn't do `noexcept` either. I'll see whether I can simplify the generated code.

For the hand-written code with behavior inside the constructor body, do you have suggestions to make it more elegant than using conditional `noexcept`? Things like the implementation of `nullable` where we ["empty" the moved from object](https://github.com/Microsoft/bond/pull/191/files?w=1#diff-4aeedb45ed9d21d0c86b331f25046d8aL237).

(Also, I've split the x86 build fixed out into its own PR.)

cpp/inc/bond/core/exception.h
```diff
@@ -30,6 +30,8 @@ class Exception
 protected:
     Exception(const char* msg) BOND_NOEXCEPT
     {
+        // CHWARR TODO: this isn't actually NOEXCEPT.
```
chwarr: Clean this up.

glenfe:
For constructors where there is a meaningful body, using `noexcept` with `is_nothrow_move_constructible_v<T>` is fine.

On the note of `nullable` specifically, for that first specialization, after it "moves" the other, it shouldn't set the other's `has_value` to false.

This is because the other object still **has** a value; it just **has a value in a moved-from state**. These are the same semantics we use for `std::experimental::optional` move constructors.

chwarr:
Simplified by using `= default` when possible. The generated code is much cleaner now, and the Haskell is about the same. Thanks for the suggestion.

For nullable, I understand what you're saying. You've piqued my interested in how this sausage was made. Was this point discussed for `std::(experimental)::optional`? If so, do you have any pointers to the discussions about this? (E.g., mailing list discussions)

I looked at [N4594](http://open-std.org/JTC1/SC22/WG21/docs/papers/2016/n4594.pdf) and found the specification for this behavior, but no exposition.

glenfe:
The rationale is performance, to avoid the call to `~T()`. The discussion is not public.

The alternative acceptable behavior for `optional<T>`, would be that the move constructor calls the other's `~T()`. But the `nullable<T>` specialization doesn't even do that: It sets the other's `has_value` to false, but the object is still alive.

This is wrong, and inconsistent with every other standard library facility that constructs and destroys objects (and even inconsistent with the other `nullable<T>` specialization).

The `nullable<T>` specialization should either do one of two things: 1. If it sets `has_value` to false, then it should destroy the object.  2. If it does not destroy the object, do not set `has_value` to false.

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
```diff
@@ -265,9 +265,24 @@ namespace std
...
+#if !defined(#{CPP.defaultedMoveCtors})
+        #{declName}(#{declName}&& other) = default;
+#elif !defined(#{CPP.rvalueReferences})
+        #{define}
+#endif|]
+
+            -- define ctor to perform member-by-member move and--if
+            -- needed--initialize meta fields
+            define = [lt|
+        #{declName}(#{declName}&&#{param})#{initList}#{ctorBody}|]
```
sapek: I think the names don't help with understanding the logic, e.g. `implicitlyDeclared` in fact generates a conditional implicit/explicit ctor. Consider something like this:

```
moveCtor = if hasMetaFields then [lt|
#if !defined(#{CPP.rvalueReferences})
        #{explicit}
#endif
|]
                            else lt|
#if !defined(#{CPP.defaultedMoveCtors})
        #{implicit}
#elif !defined(#{CPP.rvalueReferences})
        #{explicit}
#endif
|]

```

chwarr: Fixup commit--that I'll squash before merging for real--pushed.

chwarr:
@glenfe, thanks for the details. Both of the two approaches make sense. For nullable, since we've taken approach 2 (to clear the flag) for a few years now, I feel better keeping that behavior as-is.

I looked at both specializations, and I don't see where a moved-from `nullable` doesn't destroy the `value_type`. Here's my analysis. What have I missed?

## `useValue == true` specialization

The code in question is lines [85-296](https://github.com/Microsoft/bond/blob/f540eb5370c4ba65bcd385e7f2ea53402d0a8e67/cpp/inc/bond/core/nullable.h#L85-296).

This specialization doesn't do any manual construction or allocation. There's also no custom dtor. `value_type::~value_type` will run when the `nullable` implicit dtor runs, regardless of the value of `_hasvalue`.

When we move a `nullable`, the moved from `nullable` has `_hasvalue == false`. We don't run `value_type::~value_type` on the moved from `_value` at this time. It will be run when the `nullable` itself is destroyed.

Is this gap in time when the dtor is run the problem you're highlighting?

At this point, things like `nullable::operator=` and `nullable::set(T)` rely on `value_type` being in a state that it can be assigned to after it was moved from.

`nullable::set()` is perhaps dubious, as it will "resurrect" a moved from object without resetting it to something else.

## `useValue == false` specialization

The code in question is lines [301-601](https://github.com/Microsoft/bond/blob/f540eb5370c4ba65bcd385e7f2ea53402d0a8e67/cpp/inc/bond/core/nullable.h#L301-601).

There's a pointer indirection here. We use the pointer to control the conditional destruction of the `value_type`. When we move from one of these `nullable` instances, we move the pointer, which is also the flag. The moved from instance is set back to "`nullptr`". We don't destroy the `value_type` in the moved from instance, but that's because we moved the pointer not the `value_type`.

The pointers themselves are destroyed with the `nullable` when all it's members are destroyed.

Again, I don't see where destruction is missed after moving.

glenfe:
In my earlier comment:

> The following applies to `nullable<T>` after that final point 5 is implemented from issue #167 (where T's contsruction and destruction are correctly reflected in its has_value):
> 
> The `nullable<T>` specialization should either do one of two things: 1. If it sets has_value to false, then it should destroy the object. 2. If it does not destroy the object, do not set has_value to false.

Currently it's doing neither 1 nor 2. It sets `src._hasvalue = false;` while `src`'s object still has a value (just a value in a moved-from state). That is, `src`'s value hasn't been destroyed.

When the final point in issue #167 is implemented (i.e. construction only happens when necessary) this would not be the right thing, because `_hasvalue` should be the indicator of whether the object is constructed or not (whether or not it was moved-from).

What you want, after #167 is completed, is either for that move constructor to either look like:

```
nullable(nullable&& src)
    : _hasvalue(src._hasvalue) {
    if (_hasvalue) {
        ::new(addressof(storage)) value_type(std::move(src.value()));
        src.reset(); 
    }
}
```

Where `src.reset()` destroys src's value and sets its `hasvalue` to false.

Alternatively, it could look like (i.e. If you wanted `optional` semantics):

```
nullable(nullable&& src)
    : _hasvalue(src._hasvalue) {
    if (_hasvalue) {
        ::new(addressof(storage)) value_type(std::move(src.value()));
    }
}
```

So `hasvalue` is only `true` after `value` is constructed. It is `false` if `value` is not yet constructed, or has been destroyed.

glenfe:
I missed your earlier question. The answer to:

> Is this gap in time when the dtor is run the problem you're highlighting?

Is yes.

> At this point, things like nullable::operator= and nullable::set(T) rely on value_type being in a state that it can be assigned to after it was moved from.

If those operations need to be well defined on a moved-from nullable, that's fine, they can achieve that by constructing again (after which `_hasvalue` will be `true` again) if necessary.

But currently one specialization of `nullable` has one kind of semantics for construction and destruction in relation to `_hasvalue` and the other specialization of `nullable` has a different kind of semantics.

chwarr:
Got it now. We'll address this when we finish up #167.

chwarr:
Mono build failures are a known issue. Looks like a fix is being worked on.

https://github.com/travis-ci/travis-ci/issues/6239

PR #196, opened 2016-07-07
[epoxy] Specify CBv1 marshalled payloads

chwarr:
Restrict the first version of the Epoxy transport wire format only use
Compact Binary v1 marshalled data for payload and layer. This can be
relaxed in the future, but some handshaking will be needed to do things
like agree on a common protocol, exchanges schemas, &c.


tstein:
LGTM.

PR #197, opened 2016-07-07
[epoxy] Add support for custom resolution of host names

chwarr:
* Default resolver uses DNS
* Custom resolvers can be provided via
  `EpoxyTransportBuilder.SetResolver`

### DIFF
--- a/cs/src/comm/epoxy-transport/EpoxyTransport.cs
+++ b/cs/src/comm/epoxy-transport/EpoxyTransport.cs
@@ -4,6 +4,7 @@
 namespace Bond.Comm.Epoxy
 {
     using System;
+    using System.Linq;
     using System.Net;
     using System.Net.Sockets;
     using System.Threading;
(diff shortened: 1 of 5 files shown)


cs/src/comm/epoxy-transport/EpoxyTransport.cs
```diff
@@ -118,14 +149,26 @@ public override Error GetLayerStack(out ILayerStack stack)
...
-                socket.BeginConnect, socket.EndConnect, endpoint.Host, endpoint.Port, state: null);
-
-            // TODO: keep these in some master collection for shutdown
-            var connection = EpoxyConnection.MakeClientConnection(this, socket, logger, metrics);
-            await connection.StartAsync();
-            return connection;
+
+            try
+            {
+                IPAddress ipAddress = await resolver(endpoint.Host);
```
tstein: Can we limit the try block to this statement?

chwarr: Yup. Fixed in most recent commit.

tstein:
LGTM.

PR #201, opened 2016-07-13
[epoxy] Add IPv6 support

chwarr:
* Both listeners and connections can be made to IPv6 addresses now.
* Default resolver no longer prefers IPv4 addresses. It just uses the first
  address returned.


tstein:
LGTM.

Issue #202, opened 2016-07-13
gbc cpp generates ambiguous references when a struct and enum val share a name


tstein:
When compiling a bond file containing a struct and an enum value with the same name, gbc generates C++ that makes unqualified references to that name.

For example, this bond file:

    namespace conflict;
    struct Foo {};
    enum E { Foo; };

builds with the latest master. Trying to compile the resulting conflict_apply.cpp with clang causes a number of errors like this:

```
./conflict_apply.h:28:31: error: reference to 'Foo' is ambiguous
    bool Apply(const bond::To<Foo>& transform,
                              ^
./conflict_types.h:35:12: note: candidate found by name lookup is 'conflict::Foo'
    struct Foo
           ^
./conflict_types.h:97:13: note: candidate found by name lookup is 'conflict::_bond_enumerators::E::E::Foo'
            Foo
            ^
```

chwarr:
Bond version 4.3.0 is now live on NuGet.org. These changes are included in that version.

PR #204, opened 2016-07-14
Fully qualify names in cpp apply overloads.

tstein:


tstein:
This should fix #202.

chwarr:
Can you also do these?

* update the changelog
* update the golden master C++ files for gbc-tests
* amend the commit message to say `Fixes https://github.com/Microsoft/bond/issues/202, "gbc cpp generates ambiguous references when a struct and enum val share a name"`

sapek:
I'm not sure this is a good change. 

I think it would be better to catch this in .bond compiler as syntax error. It happens that it is possible to make this work in generated C++ because we generate enums into a distinct namespace. But it won't necessary always be that way. We only do that today as a ways to get poor man's scoped enums. With C++ `enum class` that will not be necessary and in C++ enum and class share the same namespace. Also it is probably safer to not assume that other languages that we might want to support in the future would allow enum and class of the same name.

compiler/src/Language/Bond/Codegen/Cpp/ApplyOverloads.hs
```diff
@@ -22,20 +23,22 @@ data Protocol =
 
 
 -- Apply overloads
-applyOverloads :: [Protocol] -> Text -> Text -> Declaration -> Text
-applyOverloads protocols attr body Struct {..} | null declParams = [lt|
+applyOverloads :: [Protocol] -> MappingContext -> Text -> Text -> Declaration -> Text
+applyOverloads protocols cpp attr body s@Struct {..} | null declParams = [lt|
     //
     // Overloads of Apply function with common transforms for #{declName}.
```
chwarr: Use qualified name here as well?

tstein: I originally did. I changed it back because I don't think it's ambiguous to a human, and if it is, the qualified names in the code make it clear which Foo we're talking about.

tstein:
I think we should consider disallowing that - and that was my first thought, too - but that change can easily break customers' public APIs. I hit this when I took the comm bond file that we're happily building for C# and tried to build it for C++. All this does is make gbc generate valid C++ for this currently-legal bond.

tstein:
@sapek, I think you're talking about a slightly different issue than this change addresses. We're looking at structs and enum *values* with the same name. e.g.:    

    namespace conflict;
    struct Foo {};
    enum E { Foo; };

enum classes appear to actually solve this problem for C++ by requiring references to enum values to be qualified with the name of the enum, just like C# and Java do.

sapek:
@tstein, you are right, I misread the issue. The conflict between enum value names and type names applies only to the old style enums in C++ so the case for supporting this is stronger.

compiler/src/Language/Bond/Codegen/Cpp/ApplyOverloads.hs
```diff
@@ -22,20 +23,22 @@ data Protocol =
...
     // These overloads will be selected using argument dependent lookup
     // before bond::Apply function templates.
     //
-    #{attr}bool Apply(const bond::To<#{declName}>& transform,
-               const bond::bonded<#{declName}>& value)#{body}
+    #{attr}bool Apply(const bond::To< #{qualifiedName}>& transform,
```
chwarr: Minor: we don't need to add spaces here to deal with things like `<<` and `>>`. We have a fix-up step that adds spaces where needed (and handles things like trigraphs).

compiler/src/Language/Bond/Codegen/Cpp/ApplyOverloads.hs
chwarr: Makes sense.

chwarr: Makes sense.

compiler/src/Language/Bond/Codegen/Cpp/ApplyOverloads.hs
chwarr: Minor: we don't need to add spaces here to deal with things like << and >>. We have a fix-up step that adds spaces where needed (and handles things like trigraphs).

chwarr:
LGTM.
@lalo is working on fixing the AppVeyor build problems.

compiler/src/Language/Bond/Codegen/Cpp/ApplyOverloads.hs
tstein: Without those spaces, the build breaks for me.

    [ 50%] Building CXX object cpp/CMakeFiles/bond_apply.dir/bond/core/bond_apply.cpp.o
    In file included from /Users/ted/code/bond/build/cpp/bond/core/bond_apply.cpp:13:
    /Users/ted/code/bond/build/cpp/bond/core/bond_apply.h:28:30: error: found '<::' after a template name which forms the digraph '<:' (aka '[') and a ':', did you mean '< ::'?
        bool Apply(const bond::To<::bond::SerializableExceptionBase>& transform,
                                 ^~~
                                 < ::

PR #205, opened 2016-07-18
Fix AppVeyor CI

lalo:
- Use AppVeyor pre-installed Boost
- Run C++ unit tests
- Fail if Boot unit_test_framework is not found

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -129,6 +129,7 @@
     build_script:
         - ps: >-
             $env:PreferredToolArchitecture = "x64"
+            
             $env:_IsNativeEnvironment = "true"
 
             if ($env:BOND_BUILD -eq "Doc") {
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,6 +21,7 @@
         BOND_TOKEN:
             secure: 30K3L1s8erWFnnS/Iyb2LP7kiGZBr/GzA24Y82mCsGMBjzSUyk7uV7MvZ3NepNvB
         matrix:
+            - BOND_BUILD: Doc
             - BOND_BUILD: Python
               BOND_VS: "Visual Studio 14 2015"
               BOND_VS_NUM: 14
@@ -34,7 +35,6 @@
             - BOND_BUILD: C#
               BOND_OUTPUT: Properties
               BOND_CONFIG: Debug
-            - BOND_BUILD: Doc
             - BOND_BUILD: C++
               BOND_VS: "Visual Studio 12 2013"
               BOND_VS_NUM: 12
@@ -65,10 +65,9 @@
             cabal update
 
             if ($env:BOND_BUILD -eq "Doc") {
-                
-                cabal install -v3 lifted-base-0.2.3.7
 
-                cabal install --verbose=0 pandoc
+                # Temp constraint until further investigation
+                cabal install --verbose=0 pandoc --constraint='lifted-base<=0.2.3.6'
 
                 choco install doxygen.install -y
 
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -23,17 +23,23 @@
...
               BOND_BOOST: 59
             - BOND_BUILD: C#
               BOND_OUTPUT: Properties
               BOND_CONFIG: Debug
             - BOND_BUILD: Doc
             - BOND_BUILD: C++
               BOND_VS: "Visual Studio 12 2013"
-              BOND_BOOST: 55
+              BOND_VS_NUM: 12
+              BOND_ARCH: 32
```
chwarr: Can we build 64-bit for our only 2013 flavor? It's probably slower, but it's what the vast majority of our customers run in production.

appveyor.yml
```diff
@@ -44,19 +50,15 @@
...
             if ($env:BOND_BUILD -eq "Python") {
 
-                if (!(Test-Path boost.exe)) {
-
-                    echo "Downloading Boost 1.${env:BOND_BOOST} ..."
-
-                    Start-FileDownload "http://iweb.dl.sourceforge.net/project/boost/boost-binaries/1.${env:BOND_BOOST}.0/boost_1_${env:BOND_BOOST}_0-msvc-14.0-32.exe" -FileName boost.exe -Timeout 1200000
-
+                if ($env:BOND_BOOST -eq 56) {
+                    # Hard-coded Boost path as per https://www.appveyor.com/docs/installed-software#languages-libraries-frameworks
```
chwarr: Minor: just "per" here.

appveyor.yml
```diff
@@ -185,6 +185,9 @@
 
             if ($env:BOND_BUILD -eq "C++") {
 
+                $env:PreferredToolArchitecture = "x64"
```
chwarr: We can leave this above and have it apply to everything. It shouldn't hurt and might help.

appveyor.yml
```diff
@@ -76,19 +78,15 @@
 
             if ($env:BOND_BUILD -eq "C++") {
 
-                if (!(Test-Path boost.7z)) {
-
-                    echo "Downloading Boost 1.${env:BOND_BOOST} ..."
-
-                    Start-FileDownload "http://iweb.dl.sourceforge.net/project/boost/boost/1.${env:BOND_BOOST}.0/boost_1_${env:BOND_BOOST}_0.7z" -FileName boost.7z -Timeout 1200000
-
+                if ($env:BOND_BOOST -eq 56) {
```
chwarr: Can we do this once for all the builds instead of having duplicate logic in two places?

appveyor.yml
lalo: I have a hunch that it doesn't persist to the build_script stage and I want to see if I get any difference when the CI completes. I had problems with line 129 that's why I had to move it down to right before running cmake since setting it before wouldn't work. If it does work I'll move it before the if logic but still during build_script and not during before_build.

chwarr: Ah. Could be that is doesn't survive across stages.

appveyor.yml
lalo: It's hitting the 1 hour limit; see https://ci.appveyor.com/project/sapek/bond/build/1.0.807

chwarr:
LGTM :+1:

chwarr:
Looks like `pandoc` is failing to install. Not sure why. It installs from source for me locally, but I think I have a different version of `ghc`.

@lalo, can you dig in to this?

lalo:
@chwarr doesn't seem related to the latest change we had a successful run before and the last change only affects the C++ Visual Studio 2012 task.

https://ci.appveyor.com/project/sapek/bond/build/1.0.814

appveyor.yml
chwarr: Okay. Back to 32 bits then.

lalo:
@chwarr all green finally

PR #207, opened 2016-07-19
[gbc] Fully qualify names in generated C++ swap()s.

tstein:
This fixes another bit of invalid C++ generated when a bondfile contains a struct and an enum value with the same name.

Previously: #202, #204

### DIFF
--- a/compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
+++ b/compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
@@ -146,13 +146,13 @@ namespace std
         #{initMetadata}
     };
 
-    #{template}inline void swap(#{structName}& left, #{structName}& right)
+    #{template}inline void swap(#{qualifiedStructName}& left, #{qualifiedStructName}& right)
     {
         left.swap(right);
     }|]
       where
         template = CPP.template s
-        structName = CPP.structName s
+        qualifiedStructName = [lt|#{getDeclTypeName cpp s}#{CPP.structParams s}|]
 
         otherParam = if hasOnlyMetaFields then mempty else [lt| other|]
         hasOnlyMetaFields = not (any (not . getAny . metaField) structFields) && isNothing structBase
--- a/compiler/tests/generated/alias_key_types.h
+++ b/compiler/tests/generated/alias_key_types.h
@@ -79,7 +79,7 @@ namespace test
         }
     };
 
-    inline void swap(foo& left, foo& right)
+    inline void swap(::test::foo& left, ::test::foo& right)
     {
         left.swap(right);
     }
--- a/compiler/tests/generated/alias_with_allocator_types.h
+++ b/compiler/tests/generated/alias_with_allocator_types.h
@@ -132,7 +132,7 @@ namespace test
         }
     };
 
-    inline void swap(foo& left, foo& right)
+    inline void swap(::test::foo& left, ::test::foo& right)
     {
         left.swap(right);
     }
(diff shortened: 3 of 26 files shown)


lalo:
The change log doesn't need to be updated right? since it's basically part of #204

tstein:
It's the same issue as #202, which is already referenced.

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
```diff
@@ -146,13 +146,13 @@ namespace std
...
 
-    #{template}inline void swap(#{structName}& left, #{structName}& right)
+    #{template}inline void swap(#{qualifiedStructName}& left, #{qualifiedStructName}& right)
     {
         left.swap(right);
     }|]
       where
         template = CPP.template s
-        structName = CPP.structName s
+        qualifiedStructName = CPP.qualifiedStructName cpp s
```
chwarr: Can we use this in ApplyOverloads.hs now?

chwarr:
@lalo is working on a fix for the AppVeyor failures in #205.

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
tstein: I originally tried to make this change the same way I did d1efb601, but the generated code for the Box<T> in `cpp/inc/bond/core/bond.bond` didn't compile. It appears that `Box::swap(Box left, ...)` is okay, but `Box::swap(::bond::Box left ...)` fails to build because `::bond::Box` is generic. Looks like we don't generate apply functions for parameterized types, so it's not strictly necessary that we carry this change over there.

Additionally, for reasons that are not immediately obvious to me, Types deals in strings, but Apply deals in Text. To use this in Apply, I'd either have to make this function break with its neighbors and return Text, or I'd have to reTextify the string that I just unTextified. I'd like to leave it as is.

chwarr:
Going to wait until we've resolved the build failures before merging.

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
chwarr: Gotcha.

sapek: Another option is to do for `swap` what we already do [here](https://github.com/tstein/bond/blob/9661ff940190b99569d6585af1a2224691336b15/compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs#L106).

tstein: I think what you're suggesting is that we inline qualifiedStructName. I'd rather not - it's a sensible function given its neighbors, and it's long enough to be uncomfortable when used twice on one line like in swap().

sapek: I was suggesting changing this [line](https://github.com/tstein/bond/blob/9661ff940190b99569d6585af1a2224691336b15/compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs#L155) to:

    qualifiedStructName = [lt|#{getDeclTypeName cpp s}#{CPP.structParams s}|]

It is a sensible function but the implementation which creates `Text` then converts it to `String` which then will be converted to `Text` in the template is less than perfect. Maybe such a function should be a public API in `TypeMapping`?

tstein: Done. I looked at doing it in `TypeMapping`, and I sorta see how it will work, but given that we don't have a second use for this yet, I don't think it's worth the time to figure it out right now.

chwarr:
@tstein, can you re-open this to trigger the CI builds again?

PR #208, opened 2016-07-20
[epoxy] Add basic IDN support to conn. strings

chwarr:
Use `Uri.DnsSafeHost`, which can be configured at an application level
to return Punycode-encoded ACEs.

.NET 4.6 includes `Uri.IdnHost`, which is always Punycode-encoded if
needed, but Mono doesn't yet support that, so we can't use it.

### DIFF
--- a/cs/src/comm/epoxy-transport/EpoxyTransport.cs
+++ b/cs/src/comm/epoxy-transport/EpoxyTransport.cs
@@ -132,7 +132,7 @@ namespace Bond.Comm.Epoxy
                 return null;
             }
 
-            return new Endpoint(uri.Host, port);
+            return new Endpoint(uri.DnsSafeHost, port);
         }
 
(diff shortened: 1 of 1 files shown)


tstein:
Can you add a test? It'd help even if only to show what difference this makes.

chwarr:
Unfortunately, no, I haven't been able to figure out a way to add a test. Whether or not `DnsSafeHost` does Punycode encoding is an [application-wide configuration setting](https://msdn.microsoft.com/en-us/library/system.uri.dnssafehost(v=vs.110).aspx#Anchor_2), and we don't control the settings that our test runners use. `Uri.IdnHost` solves that, but Mono doesn't have `Uri.IdnHost`.

I've been able to test this manually by making the configuration changes, though.

This seems too small for an entire example.

chwarr:
Closing and re-opening to re-run CI.

chwarr:
Re-run CI.

PR #209, opened 2016-07-25
[c#] Thread request IDs through to InternalServerErrors.

tstein:
The internal consequences of this change are somewhat widespread. I've made RequestMetrics creation and emission the responsibility of Connections, rather than the ServiceHost, and I've added both flavors of metrics the base Context structure, with the immediate intent to make them visible to Layers.

This change also includes an end-to-end metrics test for requests and events. We still need a similar test to make sure valid metrics are emitted in various error cases, but this diff is uncomfortably large as it is.

### DIFF
--- a/cs/src/comm/interfaces/Errors.cs
+++ b/cs/src/comm/interfaces/Errors.cs
@@ -11,7 +11,12 @@ namespace Bond.Comm
         /// <summary>
         /// (Opaque) error message returned to client for server-side issues.
         /// </summary>
-        public static readonly string InternalErrorMessage = "The server has encounted an error";
+        public const string InternalErrorMessage = "The server has encounted an error";
+
+        /// <summary>
+        /// A unique ID for calls of MakeInternalServerError that are passed a null ID.
+        /// </summary>
+        private const string UniqueIdForNullId = "[4e2bec1b-42ca-482b-8895-b8586bc8f491]";
 
         /// <summary>
         /// Checks if an <see cref="Error" /> is an <see cref="InternalServerError"/>.
@@ -37,16 +42,17 @@ namespace Bond.Comm
         /// Creates an <see cref="InternalServerError"/> with the given message.
         /// </summary>
         /// <param name="message">An error message.</param>
+        /// <param name="uniqueId">The ID of the request that caused this error.</param>
         /// <returns>An InternalServerError representing the exception.</returns>
-        public static InternalServerError MakeInternalServerError(string message)
+        public static InternalServerError MakeInternalServerError(string message, string uniqueId)
         {
             var internalServerError = new InternalServerError
             {
-                error_code = (int)ErrorCode.InternalServerError,
-                unique_id = Guid.NewGuid().ToString("D")
+                error_code = (int) ErrorCode.InternalServerError,
+                unique_id = uniqueId ?? UniqueIdForNullId,
+                message = message ?? InternalErrorMessage
             };
-            
-            internalServerError.message = message ?? InternalErrorMessage;
+
 
             return internalServerError;
         }
@@ -55,17 +61,18 @@ namespace Bond.Comm
         /// Creates an <see cref="InternalServerError"/> from an exception.
         /// </summary>
         /// <param name="exception">An exception.</param>
+        /// <param name="uniqueId">The ID of the request that caused this error.</param>
         /// <param name="includeDetails">
         /// <c>true</c> if debugging details should be included; <c>false</c>
         /// to omit this potentailly sensitive information
         /// </param>
         /// <returns>An InternalServerError representing the exception.</returns>
-        public static InternalServerError MakeInternalServerError(Exception exception, bool includeDetails)
+        public static InternalServerError MakeInternalServerError(Exception exception, string uniqueId, bool includeDetails)
         {
             var internalServerError = new InternalServerError
             {
                 error_code = (int)ErrorCode.InternalServerError,
-                unique_id = Guid.NewGuid().ToString("D")
+                unique_id = uniqueId ?? UniqueIdForNullId
             };
 
             if (includeDetails && exception != null)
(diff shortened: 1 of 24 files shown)


lalo:
Don't forget to update the changelog

cs/src/comm/interfaces/Errors.cs
```diff
@@ -55,17 +61,18 @@ public static InternalServerError MakeInternalServerError(string message)
         /// Creates an <see cref="InternalServerError"/> from an exception.
         /// </summary>
         /// <param name="exception">An exception.</param>
+        /// <param name="uniqueId">The ID of the request that caused this error.</param>
```
chadwalters: Explain null param behavior.

cs/src/comm/interfaces/Listener.cs
```diff
@@ -17,6 +17,14 @@ public class ConnectedEventArgs : EventArgs
         public readonly Connection Connection;
 
         /// <summary>
+        /// A Bond-generated unique identifier for this connection. If you need
+        /// to return an <see cref="Error"/>, give the client this identifier,
+        /// and you will be able to cross-reference it with Bond-generated log
+        /// messages.
+        /// </summary>
+        public string ConnectionId;
```
chadwalters: Shouldn't this hang off the Connection in some way?

cs/src/comm/interfaces/Metrics.cs
```diff
@@ -39,5 +42,29 @@ public void Emit(RequestMetrics metrics)
         {
             Sink?.Emit(metrics);
         }
+
+        public static string NewId()
+        {
+            return Guid.NewGuid().ToString();
```
chadwalters: Clear C#-isms from GUID string format.

cs/src/comm/interfaces/Errors.cs
```diff
@@ -11,7 +11,12 @@ public static class Errors
         /// <summary>
         /// (Opaque) error message returned to client for server-side issues.
         /// </summary>
-        public static readonly string InternalErrorMessage = "The server has encounted an error";
+        public const string InternalErrorMessage = "The server has encounted an error";
+
+        /// <summary>
+        /// A unique ID for calls of MakeInternalServerError that are passed a null ID.
+        /// </summary>
+        private const string UniqueIdForNullId = "[4e2bec1b-42ca-482b-8895-b8586bc8f491]";
```
chadwalters: Clear C#-isms from GUID string format.

cs/src/comm/layers/LayerStack.cs
```diff
@@ -230,7 +230,7 @@ Error GetNewLayers(out ILayer<TLayerData>[] layers)
                     layers[i] = layerProviders[i].GetLayer();
                     if (layers[i] == null)
                     {
-                        result = Errors.MakeInternalServerError($"Layer provider {i} produced null layer");
+                        result = Errors.MakeInternalServerError($"Layer provider {i} produced null layer", null);
```
chadwalters: Pass ID into GetNewLayers().

cs/src/comm/layers/LayerStack.cs
```diff
@@ -239,7 +239,7 @@ Error GetNewLayers(out ILayer<TLayerData>[] layers)
                 catch (Exception ex)
                 {
                     logger.Site().Error(ex, "Layer provider {0} threw exception", i);
-                    result = Errors.MakeInternalServerError(ex, includeDetails: true);
+                    result = Errors.MakeInternalServerError(ex, null, includeDetails: true);
```
chadwalters: Pass ID into GetNewLayers().

cs/src/comm/service/ServiceHost.cs
```diff
@@ -109,125 +113,108 @@ public void ValidateServiceMethods(IEnumerable<ServiceMethodInfo> serviceMethods
...
+                    var errorMessage = "Method [" + methodName + "] invoked as if it were " +
+                                       ServiceCallbackType.RequestResponse + ", but it was registered as " +
+                                       methodInfo.CallbackType + ".";
 
                     logger.Site().Error(errorMessage);
                     var error = new Error
                     {
                         message = errorMessage,
-                        error_code = (int)ErrorCode.MethodNotFound
+                        error_code = (int) ErrorCode.InvalidInvocation
```
chadwalters: Minor: Look at other places and determine consistent formatting.

cs/src/comm/interfaces/Errors.cs
tstein: Done.

cs/src/comm/interfaces/Listener.cs
tstein: Done. This had a few other effects - check out the second commit I'll push shortly.

cs/src/comm/interfaces/Metrics.cs
tstein: I was wrong - Guid.ToString() doesn't contain the square brackets. We're good.

cs/src/comm/service/ServiceHost.cs
tstein: This is Resharper's default, and we haven't overridden it in cs.sln.DotSettings, so I believe the space is what we want to stick to.

cs/src/comm/interfaces/Errors.cs
```diff
@@ -16,7 +16,7 @@ public static class Errors
         /// <summary>
         /// A unique ID for calls of MakeInternalServerError that are passed a null ID.
         /// </summary>
-        private const string UniqueIdForNullId = "[4e2bec1b-42ca-482b-8895-b8586bc8f491]";
+        private const string UniqueIdForNullId = "4e2bec1b-42ca-482b-8895-b8586bc8f491";
```
chadwalters: Should this be public?

cs/src/comm/service/ServiceHost.cs
chadwalters: OK.

cs/src/comm/layers/LayerStack.cs
tstein: Done in the third commit.

cs/src/comm/interfaces/Errors.cs
tstein: I don't think so. No one should be using it.

chadwalters:
LGTM. Will merge when CI passes.

PR #213, opened 2016-08-09
[c++] Refine noexcept specifications

chwarr:
* nullable's move constructors and set member function sometimes invoke
  new_value, which may perform an allocation that may throw.
* value_common's move constructor copies the input Reader, so it needs
  to be conditionally noexcept on being able to make a copy.


chwarr:
Whoops. This one isn't ready yet. Closing it until I actually get it working. :-)

PR #214, opened 2016-08-09
[comm] Switch Error to only have one inner_error

chwarr:
Error now has a `nullable<bonded<Error>> inner_error` field. If multiple
errors need to be represented, the `AggregateError` struct and its error
code, `MultipleErrorsOccured`, can be used to collect these up.


chwarr:
Rebased atop latest master.

chwarr:
Re-opening to merge the right way.

PR #220, opened 2016-08-16
Update readme

lalo:
PreferredToolArchitecture=x64 should be set before running the cmake generate command. This way cmake cache will have the correct compilers and linkers.


tstein:
LGTM.

PR #221, opened 2016-08-17
[c#] Fix metrics tests.

tstein:
Previously, the tests were failing, but the failure exceptions were getting
swallowed by an async void function. Additionally, this increases the length of
the pause for metrics to settle.


lalo:
LGTM, I'll merge when the CI ends.

PR #223, opened 2016-08-23
[c++] Generate FromEnum for string conversions

chwarr:
* Generated enum types now have a `FromEnum` method that can be used to
  convert from an enum value to a string. Now generated enum types have
  all four of `ToEnum`, `FromEnum`, `ToString`, and `FromString`. (The
  `...Enum` variants return false on failure, while the `...String`
  variants throw.)
* enumerations example updated to demonstrate all four of these
  functions.
* Existing throwing `FromString` re-implemented in terms of non-throwing
  `ToEnum`.
* `ToString` is intentionally not implemented in terms of `FromEnum`, as
  `ToString` returns a reference to the name stored in the map.
  `FromEnum` copies this name into the output paramater.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,21 @@ tag versions. The Bond compiler (`gbc`) and
 different versioning scheme, following the Haskell community's
 [package versioning policy](https://wiki.haskell.org/Package_versioning_policy).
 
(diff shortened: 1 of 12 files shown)


tstein:
LGTM.

PR #225, opened 2016-08-31
[C# comm] Add docs for the epoxy address format and fix the broken PingPong example.

tstein:


examples/cs/comm/pingpong/PingPong.cs
```diff
@@ -31,7 +32,9 @@ public static void Main()
 
             Shutdown(transport);
 
-            Console.WriteLine("Done with all requests.");
+            // Let everything else finish so we can print last.
+            Thread.Sleep(50);
```
chwarr: There's some part of `Shutdown(transport)` that keeps running after the shutdown tasks are completed?

tstein: I think it's just Console IO buffering.

chwarr: `Console.Out.Flush()` then instead of a thread sleep?

tstein: Doesn't solve the problem. The writes I'm trying to wait for are happening in the metrics handler, which are being triggered asynchronously, so it's not that surprising that this doesn't help.

This is only an issue because we're running an entire world in one process and care about the order things make it to the console. I'd like to leave it as is.

chwarr:
:+1:  LGTM

PR #227, opened 2016-09-03
Temporarily remove bond.TypeDef.list_sub_type

chwarr:
The addition of `bond.TypeDef.list_sub_type` in commit
08fcff0c788eb7942c57d2ff9e9c917bf1c73dab is causing compatibility issues
with users of Bond that have already serialized SchemaDef structs in
Simple Binary.

Until we can figure out a way around this, we're reverting the addition
of this field, returning `bond.TypeDef` to the schema is has had since
3.0.7.

The rest of the infrastructure for determining a ListSubType value
remains. Unit tests have been adjusted to exercise this infrastructure
directly instead of via SchemaDef/FieldDef/TypeDef.

Re-opens https://github.com/Microsoft/bond/issues/161


chwarr:
FYI, @csdahlberg

lalo:
LGTM, ready for merge?

PR #229, opened 2016-09-08
[c# comm] Remove error code from EpoxyHeaders

chwarr:
Indicate transmission of an Error using a new framelet instead. In
addition, remove ErrorCode.OK and rename Epoxy's PayloadType to
EpoxyMessageType.

Left to do:
- [x] Update SimpleInMemTransport like we did Epoxy

@chadwalters, @tstein: please review, as you've already seen some parts of this change elsewhere.

### DIFF
--- a/cs/src/comm/epoxy-transport/EpoxyConnection.cs
+++ b/cs/src/comm/epoxy-transport/EpoxyConnection.cs
@@ -160,7 +160,7 @@ namespace Bond.Comm.Epoxy
             return $"{nameof(EpoxyConnection)}(local: {LocalEndPoint}, remote: {RemoteEndPoint})";
         }
 
-        internal static Frame MessageToFrame(ulong conversationId, string methodName, PayloadType type, IMessage message, IBonded layerData, Logger logger)
+        internal static Frame MessageToFrame(ulong conversationId, string methodName, EpoxyMessageType type, IMessage message, IBonded layerData, Logger logger)
         {
             var frame = new Frame(logger);
 
@@ -168,18 +168,9 @@ namespace Bond.Comm.Epoxy
                 var headers = new EpoxyHeaders
                 {
                     conversation_id = conversationId,
-                    payload_type = type,
+                    message_type = type,
                     method_name = methodName ?? string.Empty, // method_name is not nullable
                 };
-
-                if (message.IsError)
-                {
-                    headers.error_code = message.Error.Deserialize<Error>().error_code;
-                }
-                else
-                {
-                    headers.error_code = (int)ErrorCode.OK;
-                }
 
                 const int initialHeaderBufferSize = 150;
                 var outputBuffer = new OutputBuffer(initialHeaderBufferSize);
@@ -200,15 +191,16 @@ namespace Bond.Comm.Epoxy
             }
 
             {
-                var userData = message.IsError ? (IBonded)message.Error : (IBonded)message.RawPayload;
+                FrameletType frameletType = message.IsError ? FrameletType.ErrorData : FrameletType.PayloadData;
+                IBonded userData = message.IsError ? message.Error : message.RawPayload;
 
-                const int initialPayloadBufferSize = 1024;
-                var outputBuffer = new OutputBuffer(initialPayloadBufferSize);
+                const int initialMessageBufferSize = 1024;
+                var outputBuffer = new OutputBuffer(initialMessageBufferSize);
                 var compactWriter = new CompactBinaryWriter<OutputBuffer>(outputBuffer);
                 compactWriter.WriteVersion();
                 userData.Serialize(compactWriter);
 
-                frame.Add(new Framelet(FrameletType.PayloadData, outputBuffer.Data));
+                frame.Add(new Framelet(frameletType, outputBuffer.Data));
             }
 
             return frame;
@@ -266,7 +258,7 @@ namespace Bond.Comm.Epoxy
                 return Message.FromError(layerError);
             }
 
-            var frame = MessageToFrame(conversationId, methodName, PayloadType.Request, request, layerData, logger);
+            var frame = MessageToFrame(conversationId, methodName, EpoxyMessageType.Request, request, layerData, logger);
 
             logger.Site().Debug("{0} Sending request {1}/{2}.", this, conversationId, methodName);
             var responseTask = responseMap.Add(conversationId, layerStack);
@@ -318,7 +310,7 @@ namespace Bond.Comm.Epoxy
                 response = Message.FromError(Errors.CleanseInternalServerError(layerError));
             }
 
-            var frame = MessageToFrame(conversationId, null, PayloadType.Response, response, layerData, logger);
+            var frame = MessageToFrame(conversationId, null, EpoxyMessageType.Response, response, layerData, logger);
             logger.Site().Debug("{0} Sending reply for conversation ID {1}.", this, conversationId);
 
             bool wasSent = await SendFrameAsync(frame);
@@ -350,7 +342,7 @@ namespace Bond.Comm.Epoxy
                 return;
             }
 
-            var frame = MessageToFrame(conversationId, methodName, PayloadType.Event, message, layerData, logger);
+            var frame = MessageToFrame(conversationId, methodName, EpoxyMessageType.Event, message, layerData, logger);
 
             logger.Site().Debug("{0} Sending event {1}/{2}.", this, conversationId, methodName);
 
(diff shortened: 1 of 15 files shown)


cs/src/comm/epoxy-transport/EpoxyConnection.cs
```diff
@@ -698,49 +689,54 @@ private void DoDisconnected()
...
-                ILayerStack layerStack;
-                Error layerError = parentTransport.GetLayerStack(requestMetrics.request_id, out layerStack);
-
-                if (layerError == null)
-                {
-                    layerError = LayerStackUtils.ProcessOnReceive(
-                            layerStack, MessageType.Request, receiveContext, bondedLayerData, logger);
-                }
-
+                ILayerStack layerStack = null; // CHWARR TODO: is not giving layers a chance on this response okay?
```
tstein: Is it okay for layers to never see a request with an error in it? Yeah, I think so - it's malformed.

cs/src/comm/epoxy-transport/Frame.cs
```diff
@@ -13,11 +13,12 @@ namespace Bond.Comm.Epoxy
...
     internal enum FrameletType
     {
-        EpoxyConfig = 0x4743,
-        EpoxyHeaders = 0x5248,
-        LayerData = 0x594C,
-        PayloadData = 0x5444,
-        ProtocolError = 0x5245,
+        EpoxyConfig = 0x4743,           // "GC"
+        EpoxyHeaders = 0x5248,          // "RH"
+        ErrorData = 0x4445,             // "DE"
```
tstein: ~ Seems like this should live under PayloadData.

cs/src/comm/epoxy-transport/Frame.cs
```diff
@@ -48,10 +50,12 @@ public Framelet(FrameletType type, ArraySegment<byte> contents)
 
         public static bool IsKnownType(UInt16 value)
         {
-            return value == (UInt16)FrameletType.LayerData
-                | value == (UInt16)FrameletType.PayloadData
-                | value == (UInt16)FrameletType.EpoxyHeaders
+            return false
```
tstein: Is the false just to line up all the other checks? What's with all the single pipes (rather than double)?

cs/test/comm/Epoxy/EpoxyProtocolTests.cs
```diff
@@ -22,35 +22,32 @@ class EpoxyProtocolTests
         private const string GoodMethod = "ShaveYaks";
         private const ProtocolErrorCode MeaninglessErrorCode = ProtocolErrorCode.GENERIC_ERROR;
         private static readonly IMessage<Dummy> meaninglessMessage = new Message<Dummy>(new Dummy());
```
tstein: I feel like this guy should now be meaninglessPayload.

cs/test/comm/Epoxy/EpoxyProtocolTests.cs
```diff
@@ -281,72 +280,72 @@ public void TransitionExpectOptionalLayerData_MalformedFrame()
...
 
-            var after = EpoxyProtocol.TransitionExpectPayload(
-                EpoxyProtocol.ClassifyState.ExpectPayload, goodRequestFrame, goodRequestHeaders, emptyLayerData,
-                ref payload, ref errorCode, LoggerTests.BlackHole);
+            var after = EpoxyProtocol.TransitionExpectMessageData(
+                EpoxyProtocol.ClassifyState.ExpectMessageData, goodRequestFrame, goodRequestHeaders, emptyLayerData,
+                ref  message, ref errorCode, LoggerTests.BlackHole);
             Assert.AreEqual(EpoxyProtocol.ClassifyState.ExpectEndOfFrame, after);
-            Assert.NotNull(payload.Array);
+            Assert.NotNull(message.Data.Array);
```
tstein: Assert.False(message.IsError)?

cs/test/comm/Epoxy/EpoxyProtocolTests.cs
```diff
@@ -281,72 +280,72 @@ public void TransitionExpectOptionalLayerData_MalformedFrame()
...
 
-            after = EpoxyProtocol.TransitionExpectPayload(
-                EpoxyProtocol.ClassifyState.ExpectPayload, goodRequestLayerDataFrame, goodRequestHeaders, goodLayerData,
-                ref payload, ref errorCode, LoggerTests.BlackHole);
+            after = EpoxyProtocol.TransitionExpectMessageData(
+                EpoxyProtocol.ClassifyState.ExpectMessageData, goodRequestLayerDataFrame, goodRequestHeaders, goodLayerData,
+                ref  message, ref errorCode, LoggerTests.BlackHole);
             Assert.AreEqual(EpoxyProtocol.ClassifyState.ExpectEndOfFrame, after);
-            Assert.NotNull(payload.Array);
+            Assert.NotNull(message.Data.Array);
```
tstein: Assert.False(message.IsError)?

tstein:
LGTM with some small stuff.

cs/src/comm/epoxy-transport/EpoxyConnection.cs
chadwalters: Yes I agree. A request received with an error in it shouldn't hit layers on the way in and shouldn't hit layers on the way out -- it is a broken request.

cs/src/comm/epoxy-transport/EpoxyProtocol.cs
```diff
@@ -368,7 +385,7 @@ internal static ClassifyResult Classify(Frame frame, Logger logger)
 
             if (frame.Count < 2)
             {
-                logger.Site().Error("Frame had headers but no payload.");
+                logger.Site().Error("Frame had headers but no message.");
```
chadwalters: Kind of a nit but might be better as "Frame had headers but no message data.".

cs/src/comm/epoxy-transport/Frame.cs
chwarr: Yes, the false is just to line up the other things. I'm assuming that compiler will optimize that out.
The single pipe is the way of saying that there's no need to perform short circuit evaluation on this.

cs/src/comm/epoxy-transport/EpoxyProtocol.cs
```diff
@@ -381,40 +398,43 @@ internal static ClassifyResult Classify(Frame frame, Logger logger)
...
                 return ClassifyState.InternalStateError;
             }
 
-            int payloadDataIndex = (layerData.Array == null ? 1 : 2);
-            if (payloadDataIndex >= frame.Count)
+            int messageDataIndex = (layerData.Array == null ? 1 : 2);
+            if (messageDataIndex >= frame.Count)
             {
-                logger.Site().Error("Frame had headers but no payload.");
+                logger.Site().Error("Frame had headers but no message.");
```
chadwalters: Kind of a nit but might be better as "Frame had headers but no message data.".

cs/src/comm/epoxy-transport/Frame.cs
chwarr: They're in alphabetical order now, which, in my experience is the only order for things like this that is maintainable over time, especially when doing three-way merges.

cs/src/comm/epoxy-transport/Frame.cs
chadwalters: I think it is weird to have false in the list.

cs/src/comm/interfaces/Message.cs
```diff
@@ -5,47 +5,46 @@ namespace Bond.Comm
...
+    /// A message can contain either a payload or an <see cref="Bond.Comm.Error"/>.
+    /// </remarks>
     public interface IMessage
     {
         /// <summary>
         /// Gets the payload as a <see cref="IBonded"/> value.
         /// </summary>
         /// <exception cref="InvalidOperationException">
-        /// Thrown when the message contains an error and not a payload.
+        /// Thrown when the message contains an Error and not a payload.
```
chadwalters: Technically, lower-case error seems better than upper-case Error, due to the edge case where the payload is of type Error.

cs/src/comm/interfaces/Message.cs
```diff
@@ -201,46 +197,36 @@ public static IMessage<TPayload> FromError<TPayload>(IBonded<Error> err)
         }
 
         /// <summary>
-        /// Creates an error message from the given error.
+        /// Creates an Error message from the given error.
```
chadwalters: You may want to visit the casing in this whole file. To me, the following is more accurate:
Creates an error message from the given Error.

cs/src/comm/epoxy-transport/Frame.cs
tstein: No big.

cs/src/comm/epoxy-transport/Frame.cs
tstein: Yeah, I don't think it's worth doing anything even slightly clever here. The false causes an unnecessary pause while reading it and the single pipes make me wonder what bit twiddling you're up to. return value == a || value == b?

tstein:
Looks good to me, too. Feel free to merge.

Issue #212, opened 2016-08-05
Newtonsoft JSON version 9.0.1


yagilofir:
Hi, is there a chance you'll upgrade to the latest version of Netonsoft.Json ? (v9.0.1)
We are having some troubles using your json serializer in a .NET project that uses json 9.0.1 (and it seems that Bond currently works with Newtonsoft.Json v7.0.1 which is a bit obsolete.)
Thanks.

chwarr:
We'll need to look at what sort of breaking changes came in version Newtonsoft.Json versions 8 and 9. If it "just" works, we can relax the upper bound, otherwise, some more work will need to be done to make Bond work with 7, 8, and 9.

lalo:
@yagilofir Bond 4.3.0 supports Netonsoft.Json v9.0.1

chwarr:
Re-reading the nuspec verisioning documentation, I think what we [had before commit e2be0bf][1] was the correct thing. It specified a lower bound of 7.0.1 but no upper bound.

Looking through [the history of NuGet's update behavior][2], the default is to install the _lowest_ version that satisfied the constraints. If you'd like to use Newtonsoft JSON 9.0.1, you can install/upgrade to that version of Newtonsoft.Json explicitly, and Bond will still work. You can also use `Install-Package -DependencyVersion Highest`:

```
PM> install-package bond.csharp -DependencyVersion Highest -WhatIf

Attempting to gather dependency information for package 'bond.csharp.4.3.0' with respect to project 'Playground', targeting '.NETFramework,Version=v4.6.1'
Gathering dependency information took 584.53 ms
Attempting to resolve dependencies for package 'bond.csharp.4.3.0' with DependencyBehavior 'Highest'
Resolving dependency information took 0 ms
Resolving actions to install package 'bond.csharp.4.3.0'
Resolved actions to install package 'bond.csharp.4.3.0'
Install Bond.Core.CSharp.4.3.0
Install Newtonsoft.Json.9.0.1
Install Bond.Runtime.CSharp.4.3.0 : Bond.Core.CSharp [4.3.0, 4.3.0], Newtonsoft.Json [7.0.1, 10.0.0)
Install Bond.CSharp.4.3.0 : Bond.Core.CSharp [4.3.0, 4.3.0], Bond.Runtime.CSharp [4.3.0, 4.3.0]
Time Elapsed: 00:00:01.2734569
```

I'm preparing a PR to relax the upper bound in the .nuspec back to what we had before.

[...]

PR #230, opened 2016-09-10
Undo Bond.Runtime's limit to Newtonsoft.Json < 10

chwarr:
The Bond.Runtime NuGet package again only specifies a lower bound on its
Newtonsoft.JSON dependency. This reverts commit
e2be0bf3ec12e97555cf6204dc56cb8b6a4985f4, which artificially limited the
upper bound to < 10. The previous behavior did not limit Bond.Runtime to
only working with Newtonsoft.Json 7.0.1. _However_, the default behavior
of NuGet (since 2.8) is to install the _lowest_ version of a package
that satisfied the constraints. This behavior can be overridden when a
package is installed, but that decision needs to be made by the consumer
of Bond.Runtime.

Closes https://github.com/Microsoft/bond/issues/212

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,9 @@ different versioning scheme, following the Haskell community's
   consumers of serialized SchemaDef. We plan to restore this field in the
   future.
   [Issue #161 re-opened](https://github.com/Microsoft/bond/issues/161)
+* The Bond.Runtime NuGet package no longer artificially limits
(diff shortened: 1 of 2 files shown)


chadwalters:
This behavior of NuGet seems to be a poor choice and contrary to semantic versioning conventions.

lalo:
But wouldn't we want to keep a tighter control of what versions we support as a reliability factor? that is if we would have the proper tests for each major version.

PR #233, opened 2016-09-13
[IDL Comm] Move and rename SimpleInMem .bond file

chadwalters:
Move SimpleInMem transport's .bond file to IDL directory and rename
to align with naming convention for other .bond files.


lalo:
The test that failed probably reached the 1 hour limit per test.

PR #234, opened 2016-09-30
CMake improvements for stand-alone builds

chwarr:
Improves the Bond CMake build to make it easier to build the minimum amount needed to consume Bond in a C++ project.
* Improve logic to locate a pre-compiled gbc
* Add ability to request that just the libraries be built and no additional tools (like bf)


chwarr:
MSVC C++ 2013 build timed out. Merging anyway.

PR #236, opened 2016-10-13
Bring C++ and C# serialization examples in sync

chwarr:
Both examples now use the same schema and populate the object with the same
values.


chwarr:
It looks like linking the unit tests on macOS with clang is timing out...

chadwalters:
Travis failures seem unrelated. Committing.

PR #237, opened 2016-10-13
Move logo to above the table of contents

chwarr:
* Logo is now embedded in the template and above the table of contents
* Move logo in README to be at the top and separated with a hr. This looks
  okay when rendered by GitHub and when rendered by Pandoc.
* README.md has gotten long enough that it can use a table of contents
* Use inline bash syntax highlighting for README.md: this doesn't highlight
  extra things like the python mode was doing, and looks reasonable for all
  the snippets, including the Windows CMD snippets
  * This also lets us to get rid of the index.html template--it was almost
    identical to the default.html template, but had some formatting colors
    overridden to black.


chadwalters:
Failures are not related to changes.

PR #239, opened 2016-10-17
[c#] Fix rebuilds when using BondCodegen.Options

chwarr:
If all of the items in the BondCodegen ItemGroup had their Options
metadata set, codegen was always invoked, as the dummy output file
`bondfiles.tmp` was never generated.

Now, the dummy output file (renamed to `bondcodegen.done`) is always
output, and the batched input file has been renamed to
`bonddefaultcodegen.in`.


chadwalters:
Failures are unrelated. Committing.

PR #240, opened 2016-10-17
Demonstrate kind enum in polymorphic_container example

chwarr:
Switch polymorphic_container example from using `bond_meta::full_name` to using a kind enum to show current best practices.


chadwalters:
Failures are unrelated. Committing.

PR #241, opened 2016-10-17
[C# Comm] Remove logger from LayerStackProvider

chadwalters:
LayerStackProvider's constructor no longer requires a logger.
Defer initialization of the cached layer stack in the stateless case.
Pass in logger during send/receive operations instead.


lalo:
Doing some experiments with travis timeouts on another branch https://github.com/Microsoft/bond/tree/lalo-travis-timeout

PR #246, opened 2016-10-20
[c# comm] Simplify redundant return await

chwarr:
Methods that never await anything except what they return can simply return
the Task that they were awaiting, eliminating some compiler-generated async
machinery.


tstein:
LGTM.

PR #247, opened 2016-10-20
Add warning about nullable<blob> fields

rblaze:
Nullable<blob> type behavior is surprising and should be explained in manual.


chwarr:
I haven't forgotten about this. I'm in the middle of some other higher priority work at the moment. I expect to be able to look at this on Wednesday.

chadwalters:
We should consider whether we can correct this behavior.

chwarr:
I want to copy edit this a bit before merging. However, the build is currently broken, so we're not mering anything other than build fixes. We're working on fixing the build, and then I'll merge this in after some editing.

PR #248, opened 2016-10-27
[c# epoxy] Accept multiple connections in parallel

chwarr:
- Don't block in the listener's accept loop waiting for one connection
  to be established before accepting the next one.

### DIFF
--- a/cs/src/comm/epoxy-transport/EpoxyListener.cs
+++ b/cs/src/comm/epoxy-transport/EpoxyListener.cs
@@ -24,6 +24,7 @@ namespace Bond.Comm.Epoxy
         readonly object connectionsLock = new object();
         readonly HashSet<EpoxyConnection> connections;
 
+        // used to request that other components start shutting down
         readonly CancellationTokenSource shutdownTokenSource;
 
         public IPEndPoint ListenEndpoint { get; }
@@ -134,8 +135,10 @@ namespace Bond.Comm.Epoxy
                         connections.Add(connection);
                     }
 
-                    await connection.StartAsync();
-                    logger.Site().Debug("Started server-side connection for {0}", connection.RemoteEndPoint);
+                    logger.Site().Debug("Setup server-side connection for {0}", connection.RemoteEndPoint);
+                    // Don't need to wait for the connection to get fully established before
+                    // accepting the next one, so fire and forget (with logging) StartAsync.
+                    connection.StartAsync().ContinueWith(startTask => LogClientStartResult(connection, startTask)).Forget();
                 }
                 catch (AuthenticationException ex)
                 {
@@ -174,6 +177,25 @@ namespace Bond.Comm.Epoxy
             {
                 // We tried to cleanly shutdown the socket, oh well.
                 logger.Site().Debug(ex, "Exception encountered when shutting down a socket.");
+            }
+        }
+
+        private void LogClientStartResult(EpoxyConnection connection, Task startTask)
+        {
+            if (startTask.IsFaulted)
+            {
+                logger.Site().Error(
+                    startTask.Exception,
+                    "Starting connection for {0} failed",
+                    connection.RemoteEndPoint);
+            }
+            else if (startTask.IsCanceled)
+            {
+                logger.Site().Debug("Starting connection for {0} was canceled", connection.RemoteEndPoint);
+            }
+            else
+            {
+                logger.Site().Debug("Finished starting connection for {0}", connection.RemoteEndPoint);
             }
         }
     }
--- a/cs/src/comm/epoxy-transport/EpoxyListener.cs
+++ b/cs/src/comm/epoxy-transport/EpoxyListener.cs
@@ -106,57 +106,67 @@ namespace Bond.Comm.Epoxy
             logger.Site().Information("Accepting connections on {0}", ListenEndpoint);
             while (!t.IsCancellationRequested)
             {
-                Socket socket = null;
-                EpoxyNetworkStream epoxyStream = null;
-
                 try
                 {
-                    socket = await listener.AcceptSocketAsync();
-                    logger.Site().Debug("Accepted connection from {0}.", socket.RemoteEndPoint);
+                    Socket socket = await listener.AcceptSocketAsync();
+                    EndPoint remoteEndpoint = socket.RemoteEndPoint;
+                    logger.Site().Debug("Accepted connection from {0}.", remoteEndpoint);
 
-                    EpoxyTransport.ConfigureSocketKeepAlive(socket, timeoutConfig, logger);
-
-                    epoxyStream = await EpoxyNetworkStream.MakeServerStreamAsync(socket, tlsConfig, logger);
-                    socket = null; // epoxyStream now owns the socket
-
-                    var connection = EpoxyConnection.MakeServerConnection(
-                        parentTransport,
-                        this,
-                        serviceHost,
-                        epoxyStream,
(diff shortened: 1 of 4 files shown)


cs/src/comm/epoxy-transport/EpoxyListener.cs
```diff
@@ -134,8 +135,10 @@ internal void InvokeOnDisconnected(DisconnectedEventArgs args)
                         connections.Add(connection);
                     }
 
-                    await connection.StartAsync();
-                    logger.Site().Debug("Started server-side connection for {0}", connection.RemoteEndPoint);
+                    logger.Site().Debug("Setup server-side connection for {0}", connection.RemoteEndPoint);
```
chwarr: If TLS is being used, this still blocks accepting the next connection until the previous one has finished the TLS handshake.

chwarr: Resolved in fixup commit.

chwarr:
Note to reviewer/merger: please squash these two commits.

cs/test/comm/Epoxy/EpoxyListenerTests.cs
```diff
@@ -120,6 +122,25 @@ public class EpoxyListenerTests : EpoxyTestBase
...
+            // This will just establish a TCP connection. It won't perform the Epoxy handshake, so
+            // the listener will just be sitting there waiting for the client to send some data.
+            await noHandshakeConnection.ConnectAsync(localhostEndpoint.Address, localhostEndpoint.Port);
+
+            var connectTask = transport.ConnectToAsync(localhostAddress);
+            bool didConnect = connectTask.Wait(TimeSpan.FromSeconds(30));
+            Assert.IsTrue(didConnect, "Timed out waiting for connection to be established.");
+
+            await transport.StopAsync();
+        }
```
lalo: I'm guessing this test would fail without your code? Like completely stall since the listener would be waiting forever?

cs/src/comm/epoxy-transport/EpoxyListener.cs
```diff
@@ -105,57 +106,69 @@ internal void InvokeOnDisconnected(DisconnectedEventArgs args)
...
-                    ShutdownSocketSafe(socket, epoxyStream);
+                    Socket socket = await listener.AcceptSocketAsync();
+                    EndPoint remoteEndpoint = socket.RemoteEndPoint;
+                    logger.Site().Debug("Accepted connection from {0}.", remoteEndpoint);
+
+                    // Don't need to wait for the connection to get fully established before
+                    // accepting the next one, so fire and forget (with logging) StartAsync.
+                    StartClientConnectionAsync(socket)
+                        .ContinueWith(startTask => LogClientStartResult(remoteEndpoint, startTask))
+                        .Forget();
```
tstein: I don't like extension methods, especially on well-known types like this. Just spent 10 minutes trying to search for a Task.Forget, and would have spent more if someone else hadn't tipped me off. :)

Would you be okay with `var _ = Start...`?

cs/src/comm/epoxy-transport/EpoxyListener.cs
```diff
@@ -176,5 +189,24 @@ private void ShutdownSocketSafe(Socket socket, EpoxyNetworkStream epoxyStream)
...
+            if (startTask.IsFaulted)
+            {
+                logger.Site().Error(
+                    startTask.Exception,
+                    "Starting connection for {0} failed",
+                    remoteEndPoint);
+            }
+            else if (startTask.IsCanceled)
+            {
+                logger.Site().Debug("Starting connection for {0} was canceled", remoteEndPoint);
```
tstein: My gut is telling me this should be a higher log level. I can't think of a time this would happen because everything's okay. Shutdown?

cs/test/comm/Epoxy/EpoxyListenerTests.cs
```diff
@@ -120,6 +122,25 @@ public class EpoxyListenerTests : EpoxyTestBase
...
+            listener = transport.MakeListener(localhostEndpoint);
+            await listener.StartAsync();
+
+            var noHandshakeConnection = new TcpClient();
+            // This will just establish a TCP connection. It won't perform the Epoxy handshake, so
+            // the listener will just be sitting there waiting for the client to send some data.
+            await noHandshakeConnection.ConnectAsync(localhostEndpoint.Address, localhostEndpoint.Port);
+
+            var connectTask = transport.ConnectToAsync(localhostAddress);
+            bool didConnect = connectTask.Wait(TimeSpan.FromSeconds(30));
```
tstein: This is a huge amount of time. I would think (hope?) this test would be reliable with a fraction of a second.

cs/test/comm/Epoxy/EpoxyListenerTests.cs
chwarr: Yes, that's correct. The test was written first, then the code to fix it.

cs/src/comm/epoxy-transport/EpoxyListener.cs
chwarr: When we thread cancellation support through properly the user will be able to request cancellation.

cs/test/comm/Epoxy/EpoxyListenerTests.cs
chwarr: This timeout is just to make the test fail way before the CI system's hour limit to get faster feedback. I can make it 10 seconds, but we do need to be careful about making it too aggressive, as the unit tests will be running on highly-loaded machines.

cs/src/comm/epoxy-transport/EpoxyListener.cs
chwarr: `var _ = Start()` doesn't work, because then we get a warning about an unused variable. I've inserted a `#pragma` to disable the unobserved task warning for this block of code instead.

lalo:
Should this be documented on the release notes?

cs/test/comm/Epoxy/EpoxyListenerTests.cs
```diff
@@ -114,13 +116,32 @@ public class EpoxyListenerTests : EpoxyTestBase
...
+            listener = transport.MakeListener(localhostEndpoint);
+            await listener.StartAsync();
+
+            var noHandshakeConnection = new TcpClient();
+            // This will just establish a TCP connection. It won't perform the Epoxy handshake, so
+            // the listener will just be sitting there waiting for the client to send some data.
+            await noHandshakeConnection.ConnectAsync(localhostEndpoint.Address, localhostEndpoint.Port);
+
+            var connectTask = transport.ConnectToAsync(localhostAddress);
+            bool didConnect = connectTask.Wait(TimeSpan.FromSeconds(10));
```
tstein: I get the concern about load, but this still seems three orders of magnitude too big. If it's taking 100 ms to accept a connection, either we have a huge regression, or we might want to check out other CI services. Either way, I want to know about it. Also, this is going to be really annoying if this test is ever broken and you're debugging it - no test should block that long.

I won't speculate on the relative weights of these operations, but for what it's worth, I needed a delay to make metrics tests reliable. It's [50 ms](https://github.com/Microsoft/bond/blob/master/cs/test/comm/Interfaces/MetricsTests.cs#L23).

chwarr:
Talked with @tstein: the wait times are fine for these unit tests.
Also, rebased atop the latest Mono fixes. Re-running CI builds.

PR #249, opened 2016-10-27
[macOS] Find brew-installed libpythons correctly.

tstein:


cmake/Config.cmake
```diff
@@ -44,7 +44,20 @@ endif()
 # -DPython_ADDITIONAL_VERSIONS=Major.Minor
 # if your python version is not implicitly supported by cmake
 find_package (PythonInterp 2.7)
-find_package (PythonLibs 2.7)
```
lalo: We need this line for the other non-mac configs, I believe

chwarr: `find_package` has the `HINTS` option to tell it where to look (before it does thing like look in system-default locations). If `HINTS` isn't early enough, does setting one of the [other variables][find-package-vars] work better?

I don't want to re-implement parts of `find_package` if we can make it work for us.

[find-package-vars]: https://cmake.org/cmake/help/v3.0/command/find_package.html

lalo:
Didn't have the expected result on the python mac tests:

https://travis-ci.org/Microsoft/bond/jobs/171242956

cmake/Config.cmake
tstein: We need it on Windows because python-config doesn't appear to be available. It produces the correct answer on Linux.

tstein: Based on that doc and the output of several different attempts, I don't think `find_package` works unless there is a `.cmake` available. The homebrew-provided `Python.framework` does not provide a `.cmake`.

How about running `find_package` and this code unconditionally and then setting `PYTHON_LIBRARY` if this code got a different answer?

cmake/Config.cmake
```diff
@@ -44,7 +44,18 @@ endif()
...
+if (APPLE and NOT PYTHON_LIBRARY)
+    # find_package (PythonLibs) doesn't find libpython in /usr/local/Cellar, so
+    # we have to find it ourselves.
+    execute_process (
+        COMMAND "python-config" "--prefix"
+        OUTPUT_VARIABLE PYTHON_PREFIX
+        OUTPUT_STRIP_TRAILING_WHITESPACE)
+    string (CONCAT PYTHON_LIBRARY ${PYTHON_PREFIX} "/lib/libpython2.7.dylib")
+    message (STATUS "Found PythonLibs: ${PYTHON_LIBRARY}")
```
chwarr: Are we missing an `if (EXISTS ${PYTHON_LIBRARY})` check?

tstein: Added.

PR #250, opened 2016-11-03
[c#] Swap in a custom list for SynchronizedCollection

chwarr:
SynchronizedCollection is not present in .NET Standard 1.6, among others. It
was only being used to demonstrate how collection type aliasing works, so it
has been replaced with SomeCustomList<T> which just wraps List<T> to
demonstrate how this would work.

### DIFF
--- a/cs/test/core/Aliases.bond
+++ b/cs/test/core/Aliases.bond
@@ -1,6 +1,6 @@
 ﻿namespace UnitTest.Aliases
 
-using SynchronizedList<T> = list<T>;
+using CustomList<T> = list<T>;
 using OrderedSet<T> = set<T>;
 using Decimal = blob;
 using EnumString<T> = string;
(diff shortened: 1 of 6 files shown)


chwarr:
Rebased atop latest Mono fixes. Re-running CI.

chadwalters:
The CI failures seem to be real failures related to this change.

chwarr:
Yup. I broke something somewhere during the rebase. Looking...

chwarr:
Fixed it for .NET. Letting CI run to check Mono.

Also opened ["Custom collections need to implement ICollection and not just ICollection<T>"][i254] to fix a discrepancy between the documentation and the implementation around custom containers.

[i254]: https://github.com/Microsoft/bond/issues/254

chwarr:
AppVeyor failure is a C++ timeout. PR #256 should help with these. Going ahead with the merge.

PR #252, opened 2016-11-04
Split C# reflection accommodations into its own assembly

chwarr:
Note to reviewer/merger: please either perform a full merge or a rebase. Do not squash.

[c#] Move .NET reflection adapter into an assembly

Move the code we have to deal with how the various different frameworks
handle reflection into its own assembly, Bond.Reflection. This will
allow use to use the code in more places in the future, like unit tests,
without needing to expose all the internals of Bond.dll to them.

None of the "Bond reflection" (e.g., What Bond fields does this Bond
struct have? What is the wire type of this field?) stuff has been moved.

While these reflection types are useful helpers for implementing many
parts of Bond, they are not a public API we're committing to maintaining.
To signal this, we're keeping the types internal, and will continue to
use `InternalsVisibleTo` to expose them to other assemblies.


chwarr:
Local builds with Mono and MSBuild have passed, as have the unit tests. I also ran the NuGet packing scripts and tested these packages with our test projects. Everything is working. Going to merge.

PR #255, opened 2016-11-08
[C++ comm] Fixes for io_service lifetime management

chadwalters:
Address issues with io_service lifetime management. Enable C++ comm tests.


cpp/inc/bond/comm/transport/epoxy.h
```diff
@@ -772,27 +739,29 @@ class EpoxyTransport
...
                         const SocketAddress& address,
                         const boost::shared_ptr<INetworkConnectionSink>& handler,
-                        boost::asio::io_service& ioService)
+                        const boost::shared_ptr<boost::asio::io_service>& ioServicePtr)
             : m_allocator(allocator)
             , m_address(address)
+            , m_ioServicePtr(ioServicePtr)
             , m_handler(handler)
-            , m_socket(ioService)
+            , m_socket(*ioServicePtr)
```
chwarr: We're unconditionally dereferencing a `shared_ptr` that might be empty. Similar things happen throughout this PR.

cpp/inc/bond/comm/transport/epoxy.h
```diff
@@ -944,81 +913,133 @@ class EpoxyTransport
...
-                m_socket,
-                [this](const boost::system::error_code& ec) {
-                    if (!ec)
-                    {
-                        const boost::asio::ip::tcp::endpoint endpoint =  m_socket.remote_endpoint();
-                        const SocketAddress address(endpoint);
+            void OnAccept()
+            {
+                const boost::asio::ip::tcp::endpoint endpoint = m_socket.remote_endpoint();
+                const SocketAddress address(endpoint);
```
chwarr: Minor: Can we just pass `m_socket.remote_endpoint()` directly as an argument to `SocketAddress::SocketAddress`? We're not using `endpoint` elsewhere in this file and the expression is small enough that I don't think it needs a name.

cpp/inc/bond/comm/transport/epoxy.h
```diff
@@ -944,81 +913,133 @@ class EpoxyTransport
...
+                               const boost::shared_ptr<INetworkServerSink>& handler,
+                               const boost::shared_ptr<boost::asio::io_service>& ioService)
+                : m_allocator(allocator)
+                , m_ioService(ioService)
+                , m_handler(handler)
+                , m_socket(*ioService)
+                , m_acceptor(*ioService)
+            {
+                BOOST_ASSERT(nullptr != m_handler);
+                BOOST_ASSERT(nullptr != m_ioService);
```
chwarr: In reference to my other comment about unconditionally dereferencing empty shared_ptr instances, we appear to be asserting here, but the asserts will run after the dereference.

lalo:
http://stackoverflow.com/questions/27552028/who-is-failing-boost-clang-or-gcc-issue-with-stdchrono-used-with-boostas

Related to travis failure

cpp/inc/bond/comm/transport/epoxy.h
chadwalters: See https://github.com/Microsoft/bond/pull/255/commits/a2a1d6da6ff08d8a3b3805e79b8dcbe0fe321b49

cpp/inc/bond/comm/transport/epoxy.h
chadwalters: See https://github.com/Microsoft/bond/pull/255/commits/a2a1d6da6ff08d8a3b3805e79b8dcbe0fe321b49

lalo:
1.60 failed, does not make sense it is clearly on the white list https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise

chadwalters:
I think this is now basically ready to go (assuming it passes on AppVeyor). The unit tests build on all platforms but we only run the Comm unit tests for Visual. We can dig into the issues passes the tests in the Travis environment a bit deeper once we break the Comm unit tests apart since that will help us sort through the issues faster.

cpp/test/comm/CMakeLists.txt
```diff
@@ -1,7 +1,15 @@
...
+        unit_test_comm.cpp
+        unit_test_comm.bond
+        COMM)
+else()
+    add_bond_test (unit_test_comm
+        unit_test_comm.cpp
+        unit_test_comm.bond
+        COMM
+        BUILD_ONLY)
+endif()
```
chwarr: Consider something like this to reduce duplication:

```
if (MSVC)
    set(should_build_only "")
else()
    set(should_build_only BUILD_ONLY)
endif()

add_bond_test (... COMM ${should_build_only})
```

chadwalters: I will address this in a separate PR.

PR #256, opened 2016-11-08
[C++ Core] Shatter the unit_test monolith.

tstein:
This commit makes several changes:
* All of the unit test cpps that were previously built into the single unit_test
  binary are now built into their own binaries. serialization_test.cpp still
  includes a number of files' worth of tests, still takes a long time to link,
  and should be broken up in a future commit.
* cpp/test/core/CMakeLists.txt defines a new add_unit_test geared toward
  one-or-few-files-per-suite testing, and has lost a bunch of code related to
  the old build setup. These definitions are useful for comm, too, and should
  probably move higher up the tree in a future commit.
* The hack that suppressed assertion failures from deserializing into dirty
  objects has been replaced with an #ifndef. Hopefully, the name of the macro is
  obnoxious enough to discourage its use outside unit tests.


cpp/test/core/validate_tests.cpp
```diff
@@ -172,3 +172,10 @@ void ValidateTest::Initialize()
         ValidationTests<0x2001>("Validation tests");
     );
 }
+
+
+bool init_unit_test()
+{
+    ValidateTest::Initialize();
+    return true;
+}
```
chadwalters: Minor: Add newline at end of file

cpp/test/core/skip_type_tests.cpp
```diff
@@ -74,3 +74,10 @@ void SkipTest::InitializeMismatchedTypeTests()
             bond::FastBinaryWriter<bond::OutputBuffer> >("Skip mismatched type tests for FastBinary");
     );
 }
+
+
+bool init_unit_test()
+{
+    SkipTest::InitializeMismatchedTypeTests();
+    return true;
+}
```
chadwalters: Minor: Add newline at end of file

cpp/test/core/protocol_test.cpp
```diff
@@ -277,3 +277,10 @@ void ProtocolTest::Initialize()
             bond::FastBinaryWriter<bond::OutputBuffer> >("FastBinary protocol test");
     );
 }
+
+
+bool init_unit_test()
+{
+    ProtocolTest::Initialize();
+    return true;
+}
```
chadwalters: Minor: Add newline at end of file

cpp/test/core/metadata_tests.cpp
```diff
@@ -299,3 +299,10 @@ void MetadataTest::Initialize()
             (suite, "Runtime metadata test");
     );
 }
+
+
+bool init_unit_test()
+{
+    MetadataTest::Initialize();
+    return true;
+}
```
chadwalters: Minor: Add newline at end of file

cpp/test/core/validate_tests.cpp
tstein: Did all of these.

cpp/test/core/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
-function (add_unit_test_library name)
-    list (INSERT ARGV 1 EXCLUDE_FROM_ALL)
-    add_library (${ARGV})
+function (add_unit_test)
+    STRING (REPLACE ".cpp" "" name ${ARGV0})
```
chwarr: Does [`get_filename_component(name ${ARGV0} NAME_WE)`][doc] work here?

[doc]: https://cmake.org/cmake/help/v3.0/command/get_filename_component.html

cpp/test/core/main.cpp
```diff
@@ -1,9 +1,24 @@
...
+/*
+ * We need to include different boost unit test headers on Windows and *nix.
+ *
+ * Using boost/test/unit_test.hpp on Windows causes linker errors for
+ * unit_test_main.
+ *
+ * Using boost/test/included/unit_test.hpp on *nix causes segfaults during
+ * teardown in some suites.
+ */
+#if defined(_MSC_VER)
```
chwarr: NOTE TO SELF: wre

cpp/test/core/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
-function (add_unit_test_library name)
-    list (INSERT ARGV 1 EXCLUDE_FROM_ALL)
-    add_library (${ARGV})
+function (add_unit_test)
```
tstein: Done.

cpp/test/core/CMakeLists.txt
tstein: Done.

cpp/test/core/main.cpp
tstein: Done.

cpp/test/CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
 # warning C4310: cast truncates constant value
 cxx_add_compile_options(MSVC /wd4310)
 
+# Core unit tests don't build under VS2013 without this.
+cxx_add_compile_options(MSVC /EHa)
```
chwarr: Let me see if I can figure out a way to eliminate this switch. It's a scary switch that make `catch(...)` also catch things like hardware exceptions.

chwarr: This is needed for Boost 1.53 compatibility (and possibly 1.54-1.58 as well). It's scoped to just the unit tests, so we're OK for now.

tstein:
In summary:

* Travis macOS C++ tests time out, but their test env is so anemic that we need another plan anyway.
* VS2013 with Boost 1.56 appears to fail without /EHa. We have enabled it for the moment and believe this to only be necessary for pretty old versions of Boost. We will revisit this in the immediate future.
* All other lights are green.

PR #259, opened 2016-11-08
[c#] Handle $BondOutputDirectory missing a final slash

chwarr:
The C# code gen targets now work whether or not the $BondOutputDirectory
ends with a trailing slash or not.


chadwalters:
LGTM

PR #261, opened 2016-11-10
[c#] Rearrange nuspec files by target

chwarr:
The intent is to make merging easier when we add/remote target
frameworks in the future.

* List target first so everything going to the same target can easily be
  sorted together.
* Sort lines and split into logical sections.

### DIFF
--- a/cs/nuget/.dir-locals.el
+++ b/cs/nuget/.dir-locals.el
@@ -1,0 +1,8 @@
+;;; Directory Local Variables
+;;; For more information see (info "(emacs) Directory Variables")
+
+; In the .nuspec files we want line sorting for things like <files ...> to
+; be CASE SENSITIVE, otherwise Bond.dll and Bond.xml get sorted really far
+; away from each other.
+((nil
+  (sort-fold-case)))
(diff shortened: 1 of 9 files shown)


chwarr:
I did the rearranging with the Emacs regexp search/replace of
`<file \(src=.+\) \(target=.+\) />` to `<file \2 \1 />`

I then used `sort-lines` (with `sort-fold-case` set to `nil`) to sort the files sections before adding blank lines between each target.

cs/nuget/.dir-locals.el
```diff
@@ -0,0 +1,8 @@
+;;; Directory Local Variables
+;;; For more information see (info "(emacs) Directory Variables")
+
+; In the .nuspec files we want line sorting for things like <files ...> to
+; be CASE SENSITIVE, otherwise Bond.dll and Bond.xml get sorted really far
+; away from each other.
+((nil
+  (sort-fold-case)))
```
lalo: Do we need to commit this file?

chwarr: Yes. It automatically configures case-sensitive sorting for anyone using Emacs. I'm happy to add similar Vim magic. Just let me know what it is.

PR #265, opened 2016-11-11
[travis] Use Trusty Tahr for Linux CI.

tstein:


.travis.yml
```diff
@@ -1,4 +1,5 @@
-sudo: false
+dist: trusty
+sudo: required
```
chwarr: Can we add a comment explaining why we have the `sudo: required` line? IIUC, it's not because we need sudo features, but, it's to get the trusty image.

PR #268, opened 2016-11-14
Add option to compile and not execute a specific test

lalo:
- Add BUILD_ONLY as an argument to add_bond_test() CMake function


chwarr:
LGTM.

PR #270, opened 2016-11-14
[C++] Remove now-useless use_cxx11 build directives.

tstein:


chwarr:
Might want to amend commit message to mention that the individual ones are no longer needed, as commit 7a837b4b changed the config for the compilers globally C++11 mode.

tstein:
Pushed after being so amended.

PR #272, opened 2016-11-15
[comm] Remove duplicate transport .bond files

chwarr:
These files are left over after mering in C++ Comm (commit
7a837b4b9696cd20915b07100d9859acee20e6a2).


chadwalters:
LGTM. CI failure unrelated.

PR #273, opened 2016-11-15
[C++] Split unit_test_common into core and comm flavors.

tstein:


chadwalters:
2nd iteration resolves the MSVC issues on my local build. Ship it.

PR #274, opened 2016-11-15
[c++] Fully implement RapidJSON output stream

chwarr:
The [RapidJSON (output) stream concept][1] needs a `Ch` typedef and a
`Flush` method.

Fixes https://github.com/Microsoft/bond/issues/271

I've tested with fix with the current version of RapidJSON we use, as
well as v1.0.0 and v1.1.0.

[1]: http://rapidjson.org/md_doc_stream.html#CustomStream


tstein:
LGTM.

PR #276, opened 2016-11-18
Reduce AppVeyor CI time for VC2013

chadwalters:
Adds some knobs to reduce individual build times by skipping portions of test/example build/run time.

Split the VC2013 run on AppVeyor CI into Core and Comm portions and stop running the gbc tests on VC2013 configurations.

VC2013 Comm portion now runs quickly. VC2013 Core portion still exceeds the 1 hour limit so will require further work but I think this PR represents a chunk of progress.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,7 +25,9 @@ add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} -C ${CMAKE_CFG_INTDIR} --
 
 if (NOT BOND_GBC_PATH)
     add_subfolder (compiler "compiler")
-    add_dependencies (check gbc-tests)
+    if (NOT BOND_SKIP_GBC_TESTS)
+        add_dependencies (check gbc-tests)
+    endif()
 endif()
 
 add_subdirectory (cpp)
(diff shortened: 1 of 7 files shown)


tstein:
LGTM, but `BOND_CORE_ONLY` feels wrong next to the new `BOND_SKIP*`. Any thoughts on changing it to `BOND_SKIP_COMM` in this PR?

chadwalters:
I'd prefer to keep BOND_CORE_ONLY as is for now. I think BOND_CORE_ONLY represents a different semantic concept than BOND_SKIP_COMM. I realize that they are effectively equivalent at the moment because we only have Core and Comm. We will probably end up adding individual BOND_SKIP_* for Comm and other modules if/when more modules show up.

tstein:
Works for me. Go ahead and merge once you see the desired effect on appveyor.

chadwalters:
All AppVeyor builds passed. Travis failure was related to enabling Comm examples on Linux/Clang. Spinning a new version of this PR that only enables Comm examples for MSVC. Linux/Clang issues will be addressed in a separate PR.

chadwalters:
Pushed outside of GitHub UI. Closing.

PR #277, opened 2016-11-18
[C++ Comm] Enable examples on Linux/macOS.

tstein:


examples/cpp/comm/logging/logging.cpp
```diff
@@ -53,9 +53,15 @@ void LogHandler(const char* functionName,
         ++fn;
     }
 
+    struct tm* now_tm;
+    time_t now_time = time(NULL);
+    now_tm = localtime(&now_time);
```
chwarr: `localtime` isn't thread-safe, is it?

chadwalters:
LGTM. Travis failure on MacOS unrelated to this change. Committed to master.

PR #278, opened 2016-11-19
[C++ Comm] Remove a test that binds to INADDR_ANY.

tstein:


chadwalters:
Commit message has a typo: "IPADDR_ANY"->"INADDR_ANY"

Are we losing something by just removing the test?

Also, what about this comment in cpp/inc/bond/comm/address.h?
// Create socket address, with INADDR_ANY as an ip address and explicit port:

Seems like this might go a bit deeper.

chwarr:
Yeah, upon deeper inspection, uses of the default ctor may need to be purged from the tests as well.

tstein:
Fixed the typo. Thanks for the catch.

Nothing lost. This test only checks that you can bind INADDR_ANY and connect on 127.0.0.1, which is a test of your networking stack, not of bond.

I can't find any use of that form of the ctor in cpp/test/comm.

PR #280, opened 2016-11-22
Disable Travis CI for C++ on macOS.

tstein:
Travis's macOS VMs are just too slow to complete the C++ tests in any reasonable
amount of time.


chwarr:
We're working on an alternative for macOS CI builds.

PR #288, opened 2017-01-03
Extend bf utility to support multiple payloads

sapek:
The `bf` C++ example is a handy utility to work with files containing
arbitrary Bond payloads. This change extends it to support files with
more than one Bond payload.

The most obvious use case is files with a sequence of records (e.g.
logs). The less obvious but very powerful scenario is extracting some
kind of header that precedes the actual Bond payload of interest. This
is possible because a Bond schema in Simple Binary protocol can be used
to model many kinds of arbitrary headers (e.g. any fixed size header
aligned to octet boundary).

The change is backward compatible and existing command line arguments
retain their old semantics. In order to process multiple payloads user
can specified multiple `--schema` and/or multiple `--from` arguments,
e.g.:

    bf --from=simple --schema=header.json,payload.json file

In the above example `bf` will first use Simple Binary to decode header
in schema specified in `header.json` and then will try to guess the
protocol of the next payload (since only one `--from` argument was
specified) and decode it using schema specified by the `payload.json`
file.

Multiple values for the `--schema` and `--from` arguments can be
specified either as comma delimited values (like in the example above)
or by passing the argument multiple times, e.g.:

    bf --from=fast --from=fast file

### DIFF
--- a/examples/cpp/core/bf/main.cpp
+++ b/examples/cpp/core/bf/main.cpp
@@ -102,7 +102,7 @@ bond::SchemaDef LoadSchema(const std::string& file)
 
     char c;
     tryJson.Read(c);
-        
+
     return (c == '{')
         ? bond::Deserialize<bond::SchemaDef>(bond::SimpleJsonReader<InputFile>(input))
         : bond::Unmarshal<bond::SchemaDef>(input);
@@ -111,9 +111,9 @@ bond::SchemaDef LoadSchema(const std::string& file)
 template <typename Reader, typename Writer>
 void TranscodeFromTo(Reader& reader, Writer& writer, const Options& options)
 {
-    if (!options.schema.empty())
+    if (!options.schema.empty() && !options.schema.front().empty())
     {
-        bond::SchemaDef schema(LoadSchema(options.schema));
+        bond::SchemaDef schema(LoadSchema(options.schema.front()));
         bond::bonded<void, typename bond::ProtocolReader<typename Reader::Buffer> >(reader, bond::RuntimeSchema(schema)).Serialize(writer);
     }
     else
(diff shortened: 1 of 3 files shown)


examples/cpp/core/bf/main.cpp
```diff
@@ -113,7 +113,7 @@ void TranscodeFromTo(Reader& reader, Writer& writer, const Options& options)
 {
     if (!options.schema.empty())
     {
-        bond::SchemaDef schema(LoadSchema(options.schema));
+        bond::SchemaDef schema(LoadSchema(options.schema.front()));
```
chwarr: This doesn't look like it handles an empty string like the old code used to.
Based on my reading and testing of [`boost::escaped_list_separator`][els], which is used for the cmd arg parsing, it will return an empty string for the first item in an input like  `--schema=,second.schema`.


[els]: http://www.boost.org/doc/libs/1_63_0/libs/tokenizer/escaped_list_separator.htm

examples/cpp/core/bf/main.cpp
```diff
@@ -193,23 +193,28 @@ bool TranscodeFrom(Reader reader, const Options& options)
...
-
-bool Transcode(InputFile& input, const Options& options)
+template <typename Input>
+bool Transcode(Input input, const Options& options)
 {
-    switch (options.from)
+    bf::Protocol from = options.from.empty() ? guess : options.from.front();
+
+    if (from == guess)
+        std::cerr << std::endl << "Guessed " << ToString(from = Guess(input)) << std::endl;
```
chwarr: Let's put the `from = Guess(input)` on its own line so we don't mix interesting logic with diagnostic output.

examples/cpp/core/bf/main.cpp
```diff
@@ -226,11 +231,32 @@ int main(int argc, char** argv)
...
+                        return 1;
+                }
+
+                if (!options.schema.empty())
+                    options.schema.pop_front();
+
+                if (!options.from.empty())
+                    options.from.pop_front();
+            }
+            while (!options.schema.empty() || !options.from.empty());
```
chwarr: An extension, if you're interested, would be to have a way to specify that the last from/schema pair should be used to continually transcode until EOF.

sapek: I like the idea but let me think about interface for this separately...

examples/cpp/core/bf/main.cpp
sapek: It wasn't changed from the previous version. But you are right, it is cleaner when separate.

examples/cpp/core/bf/main.cpp
sapek: good catch

sapek:
@chwarr, make the changes you requested.

Issue #284, opened 2016-12-01
Failed to build c++ examples at release mode


jacobwpeng:
OS: linux x86_64
Kernel version: 3.10.0
Compiler: clang-3.9.0
Source version: bond-5.0.0 release
Failed files:
examples/cpp/comm/generic/generic.cpp
examples/cpp/comm/passthrough/passthrough.cpp

Both files contain a variable "result" which only used in the next assert, but assert is not working at release mode.
Since release mode use -Wall -Werror, maybe it's better to define a macro to fix these compile errors;
eg:
#define BOND_UNUSED(var) (void)var
BOND_UNUSED(result);

chwarr:
Actually, it would be better to either always enable asserts in the examples or switch to something that isn't specific to the `NDEBUG` macro. The examples are run as part of the check target, and we do want to validate the results, even for a release build.

If you are just trying to _consume_ a release build of Bond, you can get everything you need by building the default target.

PR #290, opened 2017-01-04
Reference variable, which would otherwise only be referenced in an assert()

jdubrule:
Fix small compile error triggered in VS2015, update 3, in non-debug flavors.

### DIFF
--- a/examples/cpp/comm/error/error.cpp
+++ b/examples/cpp/comm/error/error.cpp
@@ -61,6 +61,7 @@ int BOND_CALL main()
     }
     catch (const bond::comm::CommException& ex)
     {
+        boost::ignore_unused_variable_warning(ex);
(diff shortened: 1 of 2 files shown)


chwarr:
LGTM. I'll merge this after CI builds finish.

A more robust fix is to not use `assert()` for these sort of things. As currently written, the tests don't test much when `assert()` is disabled, so retail test runs do very little.

jdubrule:
Yeah; seemed too easy not to submit :-)

-jeff

chwarr:
Issue #284 is about the same root problem.

chwarr:
I just pushed a branch (https://github.com/Microsoft/bond/tree/chwarr/comm-unref) with this fix and one for the problems in #284. After those CI builds pass, I'll merge it in, which will close this PR.

PR #295, opened 2017-01-11
[c# epoxy] Handle client-side connection failures

chwarr:
* Client-side connection failures now close the sockets and streams that
  are created during the connection process if an error occurs while
  setting up the TCP or TLS connection.
* Refactored similar code that is used in the listener so that we have
  one copy of the core algorithm.
* Removed some redundant try...catch blocks that were just logging
  exceptions and installed top-level connect/accept exception handles
  that then log.

TODO

* [x] Check that configuring TCP keep alive can be done after opening a client-side connection. - It can be done.

### DIFF
--- a/cs/src/comm/epoxy-transport/EpoxyListener.cs
+++ b/cs/src/comm/epoxy-transport/EpoxyListener.cs
@@ -106,11 +106,15 @@ namespace Bond.Comm.Epoxy
             logger.Site().Information("Accepting connections on {0}", ListenEndpoint);
             while (!t.IsCancellationRequested)
             {
+                // We only need to handle exceptions from AcceptSocketAsync. Exceptions from other
+                // code are handled via StartClientConnectionAsync and LogClientStartResult.
                 try
                 {
-                    Socket socket = await listener.AcceptSocketAsync();
-                    EndPoint remoteEndpoint = socket.RemoteEndPoint;
-                    logger.Site().Debug("Accepted connection from {0}.", remoteEndpoint);
+                    // We need a Task<Socket> later on, so we don't directly await this call here.
+                    Task<Socket> socket = listener.AcceptSocketAsync();
+                    // We await here to block this loop until the next connection has come in.
+                    await socket;
+                    logger.Site().Debug("Accepted TCP connection from {0}.", socket.Result.RemoteEndPoint);
 
                     // Disabling CS4014 "Because this call is not awaited,
                     // execution of the current method continues before the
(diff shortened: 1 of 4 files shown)


cs/src/comm/epoxy-transport/EpoxyNetworkStream.cs
```diff
@@ -58,6 +58,62 @@ private EpoxyNetworkStream(Socket socket, Stream stream, Logger logger)
             this.logger = logger;
         }
 
+        /// <summary>
+        /// Safely creates an EpoxyNetworkStream from a TCP socket, handling cleaning up resources
+        /// when errors are encountered.
+        /// </summary>
+        /// <remarks>
+        /// This is a helper function that centralizes error handling and socket shutdown.
```
chwarr: Rework this comment.

cs/src/comm/epoxy-transport/EpoxyNetworkStream.cs
```diff
@@ -238,5 +316,78 @@ public void Shutdown()
...
+            if (epoxyStream != null)
+            {
+                epoxyStream.Shutdown();
+                // epoxyStream owns the socket, so we shouldn't try to shutdown the socket
+                socket = null;
+            }
+
+            try
+            {
+                socket?.Shutdown(SocketShutdown.Both);
```
chwarr: Consider switching to an explicit `if (socket != null)` check.

cs/src/comm/epoxy-transport/EpoxyNetworkStream.cs
```diff
@@ -238,5 +316,78 @@ public void Shutdown()
...
+                    // clean up take over.
+                }
+                catch (Exception ex) when (ex is SocketException || ex is PlatformNotSupportedException)
+                {
+                    logger.Site().Warning(ex, "Socket keep-alive could not be configured");
+                }
+            }
+        }
+
+        static void ShutdownSocketSafe(Socket socket, EpoxyNetworkStream epoxyStream, Logger logger)
```
tstein: This can just be `ShutdownSafe`(`ly`?).

cs/src/comm/epoxy-transport/EpoxyTransport.cs
```diff
@@ -395,60 +406,25 @@ public static IPEndPoint ParseStringAddress(string address)
...
+                    state: null);
+
+                logger.Site().Information(
+                    "Established TCP connection to {0} at {1}:{2}",
+                    endpoint.Host, ipAddress, endpoint.Port);
+                return socket;
+            }
+            catch (SocketException ex)
+            {
+                logger.Site().Error(ex, "Failed to establish TCP connection to {0}", endpoint);
```
chwarr: See whether this can be removed. There should be a top-level try/catch that will log this.

cs/src/comm/epoxy-transport/EpoxyListener.cs
```diff
@@ -106,11 +106,15 @@ internal void InvokeOnDisconnected(DisconnectedEventArgs args)
...
+                // We only need to handle exceptions from AcceptSocketAsync. Exceptions from other
+                // code are handled via StartClientConnectionAsync and LogClientStartResult.
                 try
                 {
-                    Socket socket = await listener.AcceptSocketAsync();
-                    EndPoint remoteEndpoint = socket.RemoteEndPoint;
-                    logger.Site().Debug("Accepted connection from {0}.", remoteEndpoint);
+                    // We need a Task<Socket> later on, so we don't directly await this call here.
+                    Task<Socket> socket = listener.AcceptSocketAsync();
+                    // We await here to block this loop until the next connection has come in.
```
tstein: This comment seems to create more confusion than clarity.

cs/src/comm/epoxy-transport/EpoxyNetworkStream.cs
tstein: I'm okay with repeated `?.` - it does what it claims to.

cs/src/comm/epoxy-transport/EpoxyListener.cs
```diff
@@ -121,79 +125,55 @@ internal void InvokeOnDisconnected(DisconnectedEventArgs args)
...
+            // now.)
+            var connection = EpoxyConnection.MakeServerConnection(
+                parentTransport,
+                this,
+                serviceHost,
+                epoxyStream,
+                logger,
+                metrics);
+
+            lock (connectionsLock)
```
chwarr: Will do, but not in this PR.

cs/src/comm/epoxy-transport/EpoxyNetworkStream.cs
chwarr: It became a moot point after some other refactoring.

chwarr:
This PR now needs to be squashed when mering.

cs/src/comm/epoxy-transport/EpoxyNetworkStream.cs
```diff
@@ -58,6 +58,62 @@ private EpoxyNetworkStream(Socket socket, Stream stream, Logger logger)
...
+            EpoxyTransport.TimeoutConfig timeoutConfig,
+            Logger logger)
+        {
+            Socket socket = null;
+            EpoxyNetworkStream epoxyStream = null;
+
+            try
+            {
+                socket = await getSocketFunc();
+                ConfigureSocketKeepAlive(socket, timeoutConfig, logger);
```
chwarr: I just confirmed that .NET (on Windows at least) can reconfigure the socket keep alive times after the socket has been configured.

cs/src/comm/epoxy-transport/EpoxyNetworkStream.cs
```diff
@@ -59,41 +59,37 @@ private EpoxyNetworkStream(Socket socket, Stream stream, Logger logger)
         }
 
         /// <summary>
-        /// Safely creates an EpoxyNetworkStream from a TCP socket, handling cleaning up resources
-        /// when errors are encountered.
+        /// Creates an EpoxyNetworkStream from a TCP socket, handling failure that may occur during
```
tstein: s/failure/failures?

cs/src/comm/epoxy-transport/EpoxyNetworkStream.cs
```diff
@@ -59,41 +59,37 @@ private EpoxyNetworkStream(Socket socket, Stream stream, Logger logger)
...
+        /// TCP/TLS connection establishment.
         /// </summary>
         /// <remarks>
-        /// This is a helper function that centralizes error handling and socket shutdown.
+        /// This is a helper function that centralizes error handling on during connection creation.
```
tstein: s/on during/during/

cs/src/comm/epoxy-transport/EpoxyNetworkStream.cs
```diff
@@ -369,19 +365,12 @@ public void Shutdown()
             }
         }
 
-        static void ShutdownSocketSafe(Socket socket, EpoxyNetworkStream epoxyStream, Logger logger)
+        static void SafeShutdownSocket(Socket socket, Logger logger)
         {
-            if (epoxyStream != null)
-            {
-                epoxyStream.Shutdown();
```
tstein: Why isn't this necessary anymore?

chwarr: After the refactoring, there's no longer a failure mode that has an unowned, non-null `EpoxyNetworkStream`.

The next thing we do--both client- and server-size--after calling `MakeAsync` is start the Epoxy handshake by creating an `EpoxyConnection` (this connection object now owns the `EpoxyNetworkStream`) and calling `EpoxyConnection.StartAsync()`. If the Epoxy handshake fails, the connection object will shut down the `EpoxyNetworkStream` it owns.

Issue #294, opened 2017-01-11
Expoy in C++ Comm cannot be #include-d in multiple .cpp files


chwarr:
If you `#include <bond/comm/transport/epoxy.h>` in multiple .cpp files, linking the resultant executable results in multiply defined symbols errors from the linker:

``` 
"S:\src\build-bond\162\x64\examples\cpp\comm\multi-trans-units\multi-trans-units.vcxproj" (default target) (1) ->
(Link target) ->
pingpong_server.obj : error LNK2005: "class std::shared_ptr<class std::vector<class bond::blob,class std::allocator<class bond::blob> > > __cdecl bond::comm::epoxy::detail::MakeConfigFrame(class std::allocator<char>)" (?MakeConfigFrame@detail@epoxy@comm@bond@@YA?AV?$shared_ptr@V?$vector@Vblob@bond@@V?$allocator@Vblob@bond@@@std@@@std@@@std@@V?$allocator@D@6@@Z) already defined in pingpong_main.obj
[S:\src\build-bond\162\x64\examples\cpp\comm\multi-trans-units\multi-trans-units.vcxproj]
S:\src\build-bond\162\x64\examples\cpp\comm\multi-trans-units\Debug\multi-trans-units.exe : fatal error LNK1169: one or more multiply defined symbols found [S:\src\build-bond\162\x64\examples\cpp\comm\multi-trans-units\multi-trans-units.vcxproj]
```

A simple repro of this is available in https://github.com/chwarr/bond/tree/chwarr/repro-cpp-comm-multi-trans-units-symbols

PR #296, opened 2017-01-11
Address issue with multiple defines in C++ Comm

chadwalters:
Addresses https://github.com/Microsoft/bond/issues/294
Includes a new example written by Christopher that reproduces the issue.

### DIFF
--- a/cpp/inc/bond/comm/transport/detail/epoxy_data_structs.h
+++ b/cpp/inc/bond/comm/transport/detail/epoxy_data_structs.h
@@ -35,7 +35,7 @@ struct FrameState
     std::vector<std::pair<FrameletType, blob>> framelets;
 };
 
-
+inline
 std::shared_ptr<std::vector<blob>> MakeConfigFrame(std::allocator<char> allocator)
 {
     OutputBuffer stream(256, static_cast<uint32_t>(256));


tstein:
LGTM. What do you think about putting this in cpp tests instead of examples, though? It's not demonstrating functionality so much as it's proving we don't have a particular bug.

chadwalters:
Our examples tend to provide both -- examples of how to make certain scenarios work and verification that those scenarios compile and function. I agree the scenario seems a bit degenerate (this kind of thing should work out of the box) but I think it still fits better in the context of our examples than our tests.

examples/cpp/comm/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ add_subdirectory (event)
 add_subdirectory (generic)
 add_subdirectory (layer)
 add_subdirectory (logging)
+add_subdirectory (multi-trans-units)
```
chwarr: We shouldn't have an explicit multiple translation unit example like this. What I wrote was just to reproduce the reported problem. We should just take one of these existing examples and split it across translation units. E.g., implement a service separate from `main()`.

chadwalters: Let me just check in the fix itself for now then. We will need to find a good approach for making sure this kind of thing doesn't regress.

chadwalters:
Failure was the usual time out failure on VS12.

PR #298, opened 2017-01-12
Support aliases like 'using Wrapper<T> = T;', where T is an enum type

jdubrule:
Currently, you get the "enumValue: impossible happened" error, but you can't leave it without a default-value, either, since Bond expects one for an enum-type.


compiler/src/Language/Bond/Codegen/Cpp/Util.hs
```diff
@@ -103,6 +103,7 @@ defaultValue _ _ _ = error "defaultValue: impossible happened."
 enumValue :: ToText a => MappingContext -> Type -> a -> Text
 enumValue cpp (BT_UserDefined e@Enum {..} _) x =
     [lt|#{getQualifiedName cpp $ getDeclNamespace cpp e}::_bond_enumerators::#{declName}::#{x}|]
+enumValue cpp (BT_UserDefined a@Alias {..} args) e = enumValue cpp (resolveAlias a args) e
```
chwarr: We should also update one of the golden tests (like [aliases.bond](https://github.com/Microsoft/bond/blob/master/compiler/tests/schema/aliases.bond)) to show how code gets generated for this construct and to make sure it changes only when we want it to.

To update the golden tests:
```
cd compiler
cabal test # expect this to fail
dist/build/gbc-tests/gbc-tests --accept
cabal test # this should now pass
git add -u :/
# Other git commands to update the PR
```

cpp/test/core/unit_test.bond
```diff
@@ -664,6 +664,14 @@ struct WithSimpleList<T>
     1: required_optional simple_list<T> field;
 };
 
+using ValueWrapper<T> = T;
+
+struct EnumValueWrapper
```
chwarr: There's no code making any assertions about this struct, so all we're doing is testing that it compiles. We should also briefly test that is serializes/deserializes correctly.

chwarr:
LGTM now. I'll squash and merge once the CI builds pass.

PR #299, opened 2017-01-13
[C++ Comm] Break unit tests into multiple files

chadwalters:
First step in breaking up monolithic C++ Comm unit tests.
Also, address MacOS build issue related to warnings from boost::asio.


chadwalters:
NOTE: Do not squash this commits when adding to master.

Issue #300, opened 2017-01-13
No good way to generate a DateTime property in C# that serializes to wstring


DmitryTsygankov:
On Bond 5.1.0, I have a .bond file which basically looks like this: 
```
struct Sample
{
    80: optional wstring StartDateTime;
}
```

It is generating a C# file with the below property:
```
[global::Bond.Id(80), global::Bond.Type(typeof(global::Bond.Tag.wstring))]
public string StartDateTime { get; set; }

protected Sample (string fullName, string name)
{
    StartDateTime = "";
}
```

For obvious reasons, I would like it to generate a DateTime property in C#, but I would like to do it without changing the network protocol parts. When I try to do the following type alias trick, the runtime schema generation fails:
```
using DateTime = wstring;
struct Sample
{
    80: optional DateTime StartDateTime;
}

    public static class BondTypeAliasConverter
    {
        public static string Convert(DateTime value, string unused)
        {
            return value == default(DateTime) ? "" : value.ToString("o", CultureInfo.InvariantCulture);
        }

        public static DateTime Convert(string value, DateTime unused)
        {
            return value == "" ? default(DateTime) : DateTime.Parse(value);
        }
    }

    <BondCodegen Include="Sample.bond">
      <Options>--using="DateTime=System.DateTime"</Options>
    </BondCodegen>
```

The failure is happening inside of the Schema<Sample>.RuntimeSchema, namely:
```
System.InvalidCastException: Unable to cast object of type 'System.DateTime' to type 'System.String'.
    at Bond.Schema`1.Cache.GetDefaultValue(ISchemaField schemaField) in S:\bondlab\nuget\bond\cs\src\core\Schema.cs: line 224
```
For whatever reason, that code is trying to cast the default DateTime value to string, even though I am providing the BondTypeAliasConverter: variant.wstring_value = (string) defaultValue;
I think this is an issue that is significantly limiting the flexibility and usefulness of type aliases.

So I have to abandon that attempt, and try it the other way, by making the DateTime property nullable:
```
      <Options>--using="DateTime=System.Nullable&lt;System.DateTime&gt;"</Options>

    public static class BondTypeAliasConverter
    {
        public static string Convert(DateTime value, string unused)
        {
            return value == default(DateTime) ? "" : value.ToString("o", CultureInfo.InvariantCulture);
        }

        public static DateTime Convert(string value, DateTime unused)
        {
            return value == "" ? default(DateTime) : DateTime.Parse(value);
        }

[...]

PR #301, opened 2017-01-14
[c#] Compute correct default value for aliases

chwarr:
The default value of aliased bool and wstring fields was being computed
incorrectly. We were attempting to cast the aliased C# type (e.g., DateTime)
to the field type instead of using the field's natural default value
(false/empty).

Fixes https://github.com/Microsoft/bond/issues/300

# TODO #

* [x] Figure out how to reduce boiler plate in the tests
* [x] Come up with a better 50 character summary: "[c#] Fix default value for aliased bool/wstr field"
* [x] Open [issue](https://github.com/Microsoft/bond/issues/304) about `Convert<T>(Box<T> value, T unused)`

### DIFF
--- a/cs/test/core/Core.csproj
+++ b/cs/test/core/Core.csproj
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <BuildFramework Condition="'$(Configuration)' == 'Net40'">net40</BuildFramework>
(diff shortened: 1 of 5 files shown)


cs/test/core/TypeAliasTests.cs
```diff
@@ -107,6 +108,130 @@ public static ArraySegment<byte> Convert(byte[] value, ArraySegment<byte> unused
         {
             return new ArraySegment<byte>(value);
         }
+
+        #region Box<T> for basic types
```
chwarr: Does anyone have ideas for how to reduce this boiler plate with today's Bond?

Today, the code that finds Convert methods doesn't find methods like `Convert<T>(Box<T> value, T unsued)`. This could be improved, but that's inappropriate for this commit. I need to leave now, but I've added a TODO to open an issue for this sort of improvement.

cs/test/core/TypeAliasTests.cs
```diff
@@ -284,6 +434,82 @@ static BlobAlias InitBlobAlias()
...
+        {
+            Value = value;
+        }
+
+        public static implicit operator Box<T>(T value)
+        {
+            return new Box<T>(value);
+        }
+
+        public static bool operator ==(Box<T> left, Box<T> right)
```
chwarr: Check whether switching to a `struct` has the correct equality semantics and can get rid of a lot of this code.

Or, find some other dummy wrapper type elsewhere in the code base.

cs/test/core/Core.csproj
```diff
@@ -56,8 +56,8 @@
...
+      <Options Condition="'$(TargetFrameworkVersion)' == 'v4.5'">$(BondOptions) --using="Lazy=Lazy&lt;{0}&gt;" --using="OrderedSet=SortedSet&lt;{0}&gt;" --using="Decimal=decimal" --using="EnumString=Alias.EnumString&lt;{0}&gt;" --using="Array={0}[]" --using=ArrayBlob=byte[] --using="CustomList=UnitTest.Aliases.SomeCustomList&lt;{0}&gt;" --using="ConvertedBool=Box&lt;bool&gt;" --using="ConvertedInt8=Box&lt;System.SByte&gt;" --using="ConvertedInt16=Box&lt;System.Int16&gt;" --using="ConvertedInt32=Box&lt;System.Int32&gt;" --using="ConvertedInt64=Box&lt;System.Int64&gt;" --using="ConvertedUInt8=Box&lt;System.Byte&gt;" --using="ConvertedUInt16=Box&lt;System.UInt16&gt;" --using="ConvertedUInt32=Box&lt;System.UInt32&gt;" --using="ConvertedUInt64=Box&lt;System.UInt64&gt;" --using="ConvertedFloat=Box&lt;float&gt;" --using="ConvertedDouble=Box&lt;double&gt;" --using="ConvertedString=Box&lt;string&gt;" --using="ConvertedWString=Box&lt;string&gt;"</Options>
```
chwarr: There's really no reason to do this testing via codegen. If we just use aliases on a class in the unit test .cs file, we don't need changes in the .bond file or the .csproj file.

cs/test/core/TypeAliasTests.cs
chadwalters: We'd probably want to constrain T anyway but I don't think C#'s current generic constraints are expressive enough for what we need at the moment.

chadwalters:
It looks like there is a real test failure in InternalTests.cs at line 80:
Assert.AreEqual("default wstring value", Schema<StructWithDefaults>.Fields[29].default_value.wstring_value);

The value is showing up as null when it shouldn't.

chwarr:
Thanks. I'll take a look.

chwarr:
I found and fixed the wstring default value bug that I introduced.

chwarr:
All TODOs done.
This is ready to go, once CI passes and someone takes a look at the subsequent changes.

This PR needs to be squashed/rebased (and fixups applied) before merging.

chwarr:
K. Let me merge this in.

PR #307, opened 2017-01-18
Fix is_nullable for msvc 2015u3

### LINKED ISSUE #306: C++ is_nullable<T> fails (sometimes) for msvc 2015u3
I spent a few hours hunting a bug in our software. In the end, I could track it down to `bond::is_nullable` failing - sometimes - with MSVC 2015 Update 3.

Unfortunately, I cannot reproduce the issue in a small application, and the bug appears only in a rather complex piece of code where we use `mpl::foreach` to manually parse the Bond schema and use various visitors and dispatches to process the `FieldTemplate`s.

On the plus side, I managed to reliably reproduce the issue (at least in our code) by reimplementing `is_nullable` in our code base (by copying the original) and both pieces of code do not agree! (i.e. `static_assert(codeandroid::is_nullable<T>::value != bond::is_nullable<T>::value)`.

The compiler seems to be thrown off by the way default template parameters are deduced.

If we change the definition of `bond::is_nullable` from:

```
template <typename T, typename Allocator> struct
is_nullable<nullable<T, Allocator> >
    : true_type {};
```
to:
```
template <typename T, typename Allocator, bool useValue> struct
is_nullable<nullable<T, Allocator, useValue> >
    : true_type {};
```
then the compiler happily and reliably generates correct code.

Very unnerving. On the other hand, directly above `is_nullable` there is `is_list_container` which already does the same:

```
template <typename T, typename Allocator, bool useValue> struct
is_list_container<nullable<T, Allocator, useValue> >
    : true_type {};
```

I think it safe to make this change and will prepare a PR.

codeandroid:
This fixes #306 by helping msvc 2015u3 handle template parameters.

### DIFF
--- a/cpp/inc/bond/core/nullable.h
+++ b/cpp/inc/bond/core/nullable.h
@@ -726,8 +726,8 @@ is_list_container<nullable<T, Allocator, useValue> >
     : true_type {};
 
 
-template <typename T, typename Allocator> struct
-is_nullable<nullable<T, Allocator> >
+template <typename T, typename Allocator, bool useValue> struct
+is_nullable<nullable<T, Allocator, useValue> >
     : true_type {};
 
 } // namespace bond


chwarr:
Yes, this looks fine and makes sense. Thanks for the issue and the fix. Once the CI builds all come back clean, I'll merge this in.

For future reference, we try to document bug fixes like this in the CHANGELOG.md. I add entries when I merge if they weren't there, so no need to update this PR.

chwarr:
Re-running CI due to timeouts.

PR #308, opened 2017-01-19
[c++] Ensure that generated .cpp files always have an external symbol, Comm_types.hs rename

chwarr:
Two commits to review. They should be rebased when merging in to master. I'm not doing this as different PRs, as they would conflict with each other.

commit dfbdfbc7e5f58ea052faf59b75ae0cab34091e2d
Author: Christopher Warrington <chwarr@microsoft.com>
Date:   Wed Jan 18 19:38:33 2017 -0800

    [gbc] Rename Types_Comm_cpp to Comm_cpp
    
    Renamed the file, the module, and the functions.

commit 7d2ab6f2687b30b50fe49bd115ea809e737658d2
Author: Christopher Warrington <chwarr@microsoft.com>
Date:   Wed Jan 18 18:38:59 2017 -0800

    [c++] Ensure .cpp files always have a symbol
    
    Depending on the contents of the .bond file, the generated .cpp file can
    have no (new) external definitions/symbols. When mechanically creating
    static libraries, this can cause linker warnings to be emitted. (e.g.,
    LNK4221 from the Microsoft linker)
    
    To avoid this problem, all generated .cpp files now have a dummy
    definition with external linkage in an anonymous namespace.
    
    This symbol should end up in .bss and should be omitted entirely when
    finally linked into an application/shared library.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,12 +13,17 @@ different versioning scheme, following the Haskell community's
 
 ## Unreleased ##
 
-* `gbc` & compiler library: TBD (bug fix bump needed)
+* `gbc` & compiler library: TBD (B version bump needed)
 * IDL core version: TBD
 * IDL comm version: TBD
 * C++ version: TBD (minor bump needed [Boost dependencies changed])
 * C# NuGet version: TBD (minor bump needed)
 * C# Comm NuGet version: TBD (minor bump needed [dependencies changed])
+
+### bond compiler library ###
+
+* The C++ Comm .cpp template has been renamed to `comm_cpp` from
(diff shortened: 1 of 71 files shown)


chwarr:
The dummy symbol commit, 7d2ab6f, is a proposal to avoid these linker warnings. If someone has a better way, I'm all ears.

chwarr:
Looking at compiler errors from clang/gcc...

chwarr:
I'm abandoning this PR.

This approach won't work with Clang. Even with some changes to get past the original warning, Clang can figure out that the variable can't possible be used, so it then emits an unused variable warning.

PR #309, opened 2017-01-20
Remove non-ASCII characters from source files

chwarr:


cpp/inc/bond/core/stl_containers.h
```diff
@@ -63,49 +63,49 @@ require_modify_element<std::vector<bool, A> >
...
 
 
 template<typename C, typename T, typename A>
-inline 
+inline
 C* string_data(std::basic_string<C, T, A>& str)
 {
     // C++11 disallows COW string implementation (see [string.require] 21.4.1)
-    // however it was permitted in C++03. In order to support COW we can�t
+    // however it was permitted in C++03. In order to support COW we can't
```
chwarr: This had a curly apostrophe.

cs/src/core/io/IInputStream.cs
```diff
@@ -49,13 +49,13 @@ public interface IInputStream
         ulong ReadUInt64();
 
         /// <summary>
-        /// Read little-endian encoded single precision ‎IEEE 754 float
+        /// Read little-endian encoded single precision IEEE 754 float
```
chwarr: This had some fancy Unicode space between precision and IEEE.

Issue #314, opened 2017-01-27
C++ Comm services do not build when packaged in a Windows DLL


chadwalters:
C++ Comm is lacking appropriate declspec hooks to make this work. Codegen needs to be modified to allow this (along the lines of the apply-attribute functionality in gbc).

PR #318, opened 2017-02-01
[C++ Comm] Extend support for export declarations in comm.h 

chadwalters:
Allows C++ Comm services to be shared via DLLs on Windows.
Changes gbc apply-attribute option to export-attribute (with apply-attribute supported but documented as deprecated).
Also adds examples for DLL and static library support for C++ Core and Comm.
Addresses object lifetime issues in C++ Comm layers.
Fixes https://github.com/Microsoft/bond/issues/314


chadwalters:
This looks ready to go. Don't squash the commits.

PR #319, opened 2017-02-02
[C++ comm] Rephrase unit tests to match core.

tstein:


cpp/test/comm/CMakeLists.txt
```diff
@@ -1,74 +1,65 @@
...
+        -DBOND_FAST_BINARY_PROTOCOL)
+    target_link_libraries (${name} PRIVATE
+        bond
+        comm_test_common
+        ${Boost_UNIT_TEST_FRAMEWORK_LIBRARY}
+        ${Boost_CHRONO_LIBRARY})
+
+    cxx_target_compile_options (Clang ${name} PRIVATE -DBOOST_ASIO_HAS_STD_CHRONO)
+    cxx_target_compile_options (AppleClang ${name} PRIVATE -DBOOST_ASIO_HAS_STD_CHRONO)
+    # FIXME: Is MSVC the right identifier?
```
lalo: Is this FIXME comment still valid?

chwarr: MSVC is the right identifier. In [our code](https://github.com/Microsoft/bond/blob/fed300858c3cee70210a072d1ff84737a496dc45/cmake/Compiler.cmake#L3), we compare the argument, here `MSVC`, to [`CMAKE_CXX_COMPILER_ID`](https://cmake.org/cmake/help/v3.0/variable/CMAKE_LANG_COMPILER_ID.html).

Though, `cxx_target_compile_options` uses `target_compile_options` instead of `target_compile_definitions`, which means that I _think_ we might have be defining the wrong thing until this change...

cpp/test/comm/idl.cpp
```diff
@@ -0,0 +1,3 @@
+#include <unit_test_comm_apply.cpp>
```
chwarr: Minor preference for `#include "unit_test_comm_apply.cpp"` over `<...>`.

tstein: I was matching most of cpp/test/core/idl.cpp. Want me to change both files in this commit to dquotes?

cpp/test/comm/CMakeLists.txt
tstein: Yeah, I built and ran tests successfully on Windows after putting up the cr.

I'm pretty sure this works as phrased on macOS/Linux - there was a point where the build broke because these directives were still outside add_unit_test(). I'd like to leave it as-is.

cpp/test/comm/idl.cpp
chwarr: This commit. A different one. Not at all. I don't care.

cpp/test/comm/CMakeLists.txt
chwarr: Looks like all we need to do it remove the FIXME comment. I was more musing about whether/how this worked in the past... :-)

chwarr:
Fix up LGTM.

cpp/test/comm/idl.cpp
tstein: Done. Will merge after tests pass.

PR #320, opened 2017-02-07
Epoxy Transport shutdown

chwarr:
Make EpoxyTransport and EpoxyListener's StopAsync shutdown everything that they created.

TODO:

* [x] EpoxyConnection and EpoxyListener collection management is currently duplicated in both transport and listener. The code in the two is very similar. There's an opportunity to distill out this collection management. Fixed in a later commit in this PR.
* [x] Address TODO comment about bug in another test. Fixed in PR #321.

### DIFF
--- a/cs/src/comm/epoxy-transport/EpoxyListener.cs
+++ b/cs/src/comm/epoxy-transport/EpoxyListener.cs
@@ -83,21 +83,56 @@ namespace Bond.Comm.Epoxy
             return TaskExt.CompletedTask;
         }
 
-        public override Task StopAsync()
+        public override async Task StopAsync()
         {
+            // Request that the accept loop stop
             shutdownTokenSource.Cancel();
+            // Stop listening. This causes the accept loop's wait for an
+            // incomming socket to fail, which then causes the accept loop to
+            // look at the value of shutdownTokenSource.Token.
             listener.Stop();
 
-            return acceptTask;
+            // Wait for the accept loop to exit. Once it has exited, no new
+            // connections will be accepted, so we can close the outstanding
+            // ones.
+            await acceptTask;
+
+            // Collect the connections to stop.
+            HashSet<EpoxyConnection> connectionsToClose;
+            lock (connectionsLock)
+            {
+                connectionsToClose = new HashSet<EpoxyConnection>(connections);
+                connections.Clear();
+            }
+
+            // Stop all outstanding connections. EpoxyConnection shutdown is
+            // idempotent, so it's safe to call even if someone else called
+            // StopAsync.
+            var connectionShutdownTasks = new Task[connectionsToClose.Count];
+            int idx = 0;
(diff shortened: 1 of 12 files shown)


cs/src/comm/epoxy-transport/EpoxyListener.cs
```diff
@@ -83,21 +82,59 @@ public override Task StartAsync()
...
         }
 
-        public override Task StopAsync()
+        public override async Task StopAsync()
         {
+            // Request that the accept loop stop
             shutdownTokenSource.Cancel();
+
+            // Stop listening. This causes the accept loop's wait for an
+            // incomming socket to fail, which then causes the accept loop to
```
tstein: s/incomming/incoming

cs/src/comm/epoxy-transport/EpoxyListener.cs
```diff
@@ -83,21 +82,59 @@ public override Task StartAsync()
...
+            var connectionShutdownTasks = new Task[connectionsToClose.Count];
+            int idx = 0;
+            foreach (var connection in connectionsToClose)
+            {
+                logger.Site().Debug("{0} stopping connection {1}", this, connection);
+                connectionShutdownTasks[idx] = connection.StopAsync();
+                ++idx;
+            }
+
+            logger.Site().Debug("Waiting for all connection to stop.");
```
tstein: s/connection/connections

This message might deserve a higher level. I defer to you.

cs/src/comm/epoxy-transport/EpoxyTransport.cs
```diff
@@ -199,6 +206,13 @@ public struct TimeoutConfig
             logger = new Logger(logSink, enableDebugLogs);
             // Metrics sink may be null
             metrics = new Metrics(metricsSink);
+
+            shutdownLock = new ReaderWriterLockSlim(LockRecursionPolicy.NoRecursion);
+            isShutdown = false;
+
+            collectionsLock = new object();
```
tstein: It's a little odd that the shutdownLock is a dedicated *Lock type, but this is just an object. Any reason we'd be worse off doing it consistently?

cs/src/comm/epoxy-transport/EpoxyListener.cs
chwarr: Fixed.

cs/src/comm/epoxy-transport/EpoxyListener.cs
chwarr: Fixed.

cs/src/comm/epoxy-transport/EpoxyTransport.cs
chwarr: After switching everything over to a `CleanupCollection`, there is no longer even shutdown state in the `EpoxyTransport` that need guarding.

PR #321, opened 2017-02-07
[c# epoxy] Improve multiple shutdown test

chwarr:
Before, the `Connection_CanBeStoppedMultipleTimes` test was just asserting
that the connection had been shutdown once--not twice--before the
timeout.

Now, we assert that both shutdowns have happened.


chwarr:
AppVeyor failure is C++ build timeout. Kicking another build.

PR #323, opened 2017-02-08
Shatter the C++ Comm unit test monolith.

tstein:
This change includes a cleanup of names, resulting in a few conflicts between test suites and examples (since the examples are also registered as test suites). A very thorough review process determined that the test suites win that fight.


tstein:
On my Windows desktop with a hyperthreaded quad-core 3.6 GHz Xeon, the old build-and-test took 18:28. With this, it takes 17:57.

PR #327, opened 2017-02-10
Generate unique ports for each test and example.

tstein:


cmake/Bond.cmake
```diff
@@ -129,6 +130,18 @@ function (add_bond_test test)
...
             COMMAND ${test})
     endif()
+
+    get_property (lastTestPort GLOBAL PROPERTY LAST_TEST_PORT)
+    if (NOT DEFINED lastTestPort)
+        set (lastTestPort 25189)
+    endif()
+    math(EXPR testPort1 "${lastTestPort} + 1")
+    math(EXPR testPort2 "${lastTestPort} + 2")
+    target_compile_definitions (${test} PRIVATE -DTEST_PORT_1=${testPort1})
```
chwarr: `target_compile_definitions` takes multiple arguments, so we can simplify this to

`target_compile_definitions(${test} PRIVATE -DTEST_PORT_1=${testPort1} -DTEST_PORT_2=${testPort2})` and so on.

chwarr: Fun fact: `target_compiler_definition` handles [both](https://cmake.org/gitweb?p=cmake.git;a=blob;f=Source/cmTargetCompileDefinitionsCommand.cxx;h=008d1a210a9ec05fe9310794576d47267e68f46a;hb=HEAD#l46) `MACRO` and `-DMACRO`.

examples/cpp/comm/comm_dll/using_comm_dll.cpp
```diff
@@ -82,7 +82,7 @@ int main()
 
     // Exersize Comm facilities
 
-    bond::comm::SocketAddress loopback("127.0.0.1", 25188);
+    bond::comm::SocketAddress loopback("127.0.0.1", EXAMPLE_PORT_1);
```
chwarr: In the examples, do we want an `#ifndef EXAMPLE_PORT_1 #define EXAMPLE_PORT_1 25188 #endif` to make them easier to copy/paste into other build systems?

cmake/Bond.cmake
```diff
@@ -129,6 +130,18 @@ function (add_bond_test test)
...
+    target_compile_definitions (${test} PRIVATE -DTEST_PORT_2=${testPort2})
+    target_compile_definitions (${test} PRIVATE -DEXAMPLE_PORT_1=${testPort1})
```
chwarr: `TEST_PORT1` seems an OK name, even for examples. That way we only have two of these things, not four.

cmake/Bond.cmake
tstein: That is a fun fact.

Given that this function (and, really, cmake) has no way to express a structured map properly, I think I would prefer that the opaque strings we pass be the right opaque strings for the layer they're destined for. I prefer `-DMACRO`.

tstein: Done.

cmake/Bond.cmake
tstein: If you're happy, I'm happy. Done.

examples/cpp/comm/comm_dll/using_comm_dll.cpp
tstein: If we're talking specifically about examples, they only need to be given unique ports like this because we use our examples as tests. If this became a problem because we wanted to run them elsewhere, I'd rather strike at the root cause - get them out of the test framework entirely and let them use the default port.

PR #332, opened 2017-02-19
[c++ protocol] Add Read/Write/Skip type_alias methods to protocols

mirrada:
Read/Write/Skip methods are present in JSON protocol, but are missed in all other protocols. I copy-pasted them to Compact Binary, Fast Binary, Simple Binary and Random.

### DIFF
--- a/cpp/inc/bond/protocol/compact_binary.h
+++ b/cpp/inc/bond/protocol/compact_binary.h
@@ -347,6 +347,17 @@ public:
     }
 
 
+    // Read for type alias
+    template <typename T>
+    typename boost::enable_if<is_type_alias<T> >::type
+    Read(T& value)
+    {
+        typename aliased_type<T>::type x;
+        Read(x);
+        set_aliased_value(value, x);
+    }
+
+
     // Read for blob
     void Read(blob& value, uint32_t size)
     {
@@ -713,6 +724,14 @@ public:
 
         Write(length);
         detail::WriteStringData(_output, value, length);
+    }
+
+    // Write for type alias
+    template <typename T>
+    typename boost::enable_if<is_type_alias<T> >::type
+    Write(T& value)
+    {
+        Write(get_aliased_value(value));
     }
 
     // Write for blob
--- a/cpp/inc/bond/protocol/fast_binary.h
+++ b/cpp/inc/bond/protocol/fast_binary.h
@@ -148,7 +148,7 @@ public:
 
     // Read for primitive types
     template <typename T>
-    typename boost::disable_if<is_string_type<T> >::type
+    typename boost::disable_if_c<is_string_type<T>::value || is_type_alias<T>::value >::type
     Read(T& value)
     {
         _input.Read(value);
@@ -164,6 +164,17 @@ public:
 
         ReadVariableUnsigned(_input, length);
         detail::ReadStringData(_input, value, length);
+    }
+
+
+    // Read for type alias
+    template <typename T>
+    typename boost::enable_if<is_type_alias<T> >::type
+    Read(T& value)
+    {
+        typename aliased_type<T>::type x;
+        Read(x);
+        set_aliased_value(value, x);
     }
 
 
@@ -457,6 +468,14 @@ public:
 
         WriteVariableUnsigned(_output, length);
         detail::WriteStringData(_output, value, length);
+    }
+
+    // Write for type alias
+    template <typename T>
+    typename boost::enable_if<is_type_alias<T> >::type
+    Write(T& value)
+    {
+        Write(get_aliased_value(value));
     }
 
     // Write for blob
--- a/cpp/inc/bond/protocol/random_protocol.h
+++ b/cpp/inc/bond/protocol/random_protocol.h
@@ -177,6 +177,18 @@ public:
         }
     }
 
+
+    // Read for type alias
+    template <typename T>
+    typename boost::enable_if<is_type_alias<T> >::type
+    Read(T& value)
+    {
+        typename aliased_type<T>::type x;
+        Read(x);
+        set_aliased_value(value, x);
+    }
+
+
     // Read for blob
     void Read(blob& value, uint32_t size)
     {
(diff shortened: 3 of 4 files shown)


sapek:
Type aliases are handled before protocols are called (in `bond::Serializer` for serialization and in `bond::value` for deserialization). `SimpleJsonWriter` is an exception and handles type alias writing because it doesn't use default `Serializer`.

What problem are you trying to solve by adding these methods?

mirrada:
Without fixes in `CompactBinaryReader` I have compilation error:

> Error	C2664	'void bond::CompactBinaryReader<bond::InputBuffer>::Read(bond::blob &,uint32_t)': cannot convert argument 1 from 'const char *' to 'int8_t &'	PullRequest	c:\bond\cpp\inc\bond\core\value.h	284	

You can see full output in `error.txt`

I just want to serialize `const char*` in my way (`const_char_ptr.h` and `routeLinkData.bond` are attached). 

For serialization I do:

```
#include "routeLinkData_reflection.h"
#include <bond/protocol/simple_json_writer.h>
#include <bond/core/bond.h>
#include <bond/stream/stdio_output_stream.h>

void serialize(BondSerialization::RouteLinkData mode, FILE* f)
{
    using namespace BondSerialization;

    bond::StdioOutputStream stdoutput(f);
    bond::SimpleJsonWriter<bond::StdioOutputStream> stdwriter(stdoutput, true);
    bond::Serialize(mode, stdwriter);
}
```

I've tried to modify and compile your `time_alias` example (https://github.com/Microsoft/bond/tree/master/examples/cpp/core/time_alias). I changed:
```
template <> struct
aliased_type<boost::posix_time::ptime>
{
    typedef int64_t type;
};
```
to
```
template <> struct
aliased_type<boost::posix_time::ptime>
{
    typedef std::string type;
};
```

It falls into the same compile error:

> Error	C2664	'void bond::CompactBinaryReader<bond::InputBuffer>::Read(bond::blob &,uint32_t)': cannot convert argument 1 from 'boost::posix_time::ptime' to 'int8_t &'	PullRequest	c:\bond\cpp\inc\bond\core\value.h	284	

And I've noticed one more feature. I added breakpoints into your `int64_t get_aliased_value(const boost::posix_time::ptime& value)` and `void set_aliased_value(boost::posix_time::ptime& var, int64_t value)` functions. And they were not triggered. But with my fixes they were.

[zip.zip](https://github.com/Microsoft/bond/files/787229/zip.zip)

sapek:
The [interface for custom type mapping of strings](https://microsoft.github.io/bond/manual/bond_cpp.html#string-concept) is different than for basic scalar types. [Here's an example](https://github.com/Microsoft/bond/tree/939c298ff6d619dbc0bc9f5ca25c2088de86d83b/examples/cpp/core/string_ref) of custom string mapping (you can also refer to this [test](https://github.com/Microsoft/bond/blob/939c298ff6d619dbc0bc9f5ca25c2088de86d83b/cpp/test/core/container_extensibility.h#L140-L204)).

The intention of this interface is to allow de/serializaiton of custom strings w/o incurring an extra copy.

mirrada:
But `string concept` is not well-suited for my case. 

If you take a look at my `const_char_ptr.h`, you will see that I need to store all strings in common string pool (e.g. `std::vector<std::string> g_strings`). So when a string is deserialized I want to find it in a pool and store its pointer in field with type `const char*`. But `string concept` allows to overload only `T* string_data(...)`, which will be used in `std::copy` or `memcpy`. So they **will** incur an extra copy. And it is critical for my case.

Surely, I can try to overload `std::copy` and `memcpy` for my type `T*`, but this solution seems to be much more insecure than my attempts to fix `type_alias` de/serialization.

So, I really need a solution for my problem with `Microsoft.Bond`. And in my opinion I made the most trivial contribution to fix it. Give me a better idea on how to handle my case, please, and I would gladly implement it.

sapek:
You are right, the current interface doesn't let you implement string interning in a reasonable way. Adding support for aliasing `string` via `set_aliased_value`/`get_aliased_value` *might* be a good addition to consider. Some quick notes on what it would take to do it:
- It shouldn't be implemented in the protocols. It should be done in `bond::Serializer` for serialization and in `bond::value` for deserialization. Essentially today we [guard](https://github.com/Microsoft/bond/blob/bc4f7d881acbf5e8ed703e56302a3c77f3a3c522/cpp/inc/bond/core/transforms.h#L157) use of `is_type_alias` with `is_basic_type`. That's why you are getting compiler error when using `string`. The right solution, if we want to add that functionality, would be to relax this to include strings.
- It is not immediately obvious to me that there won't be any conflicts with existing string concept. We need to review where `is_string`/`is_wstring` traits are used today and make sure that we are testing these scenarios with the new string alias mechanism.
- If we add a new way to alias strings we need to add full set of test cases. See the tests I linked before as a guide.
- We would need to update documentation.
- It would be nice to have an example. String interning would be a great one.

msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

mirrada:
Thank you for understanding.

But I didn't figure out the status of my pull request.

Do you want me to make a new contribution with `type alias` fixes? Sounds like a big peace of work.

Or will you do it by yourself? How much time will it take you to manage with it?

May be you could merge this pull request until proper fixes are made?

chadwalters:
We probably won't merge the pull request as it stands since it doesn't address the problem in a way that we would like to support long term. We would like to get the long term solution in place -- we will have to look at it to see what that might take. String interning is definitely a feature we'd like to support. Can you manage your current changes as a patch set?

mirrada:
Yes, of course I can keep my changes to `Bond` only for our project. But then it will be hard for me to follow `Bond`'s improvements in `c++` and `c#` serialization.

Anyway, I am looking forward to seeing string interning feature in `Bond`. I would be glad to be informed about your final decision regarding the long term solution and the time needed for its implementation.

I just would like to thank you for the great work, many thanks!

chwarr:
Opened an issue specifically for figuring out how to support string aliased with interned strings.

I'm going to close this PR, as we'll want to come up with a design before code.

PR #333, opened 2017-02-22
[c# epoxy] Fix TLS example for IPv6 localhost

chwarr:
Before, we were only listening on IPv4 loopback. If "localhost" resolves
to [::1], we couldn't connect. Now, we listen on both of the default
IPv4 and IPv6 loopback addresses.

### DIFF
--- a/examples/cs/comm/tls/Tls.cs
+++ b/examples/cs/comm/tls/Tls.cs
@@ -13,7 +13,8 @@ namespace Bond.Examples.Tls
 
     public static class Tls
     {
-        static readonly IPEndPoint serviceEndpoint = new IPEndPoint(IPAddress.Loopback, EpoxyTransport.DefaultSecurePort);
+        static readonly IPEndPoint serviceEndpointIpv4 = new IPEndPoint(IPAddress.Loopback, EpoxyTransport.DefaultSecurePort);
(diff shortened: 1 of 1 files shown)


examples/cs/comm/tls/Tls.cs
```diff
@@ -13,7 +13,8 @@ namespace Bond.Examples.Tls
 
     public static class Tls
     {
-        static readonly IPEndPoint serviceEndpoint = new IPEndPoint(IPAddress.Loopback, EpoxyTransport.DefaultSecurePort);
+        static readonly IPEndPoint serviceEndpointIpv4 = new IPEndPoint(IPAddress.Loopback, EpoxyTransport.DefaultSecurePort);
+        static readonly IPEndPoint serviceEndpointIpv6 = new IPEndPoint(IPAddress.IPv6Loopback, EpoxyTransport.DefaultSecurePort);
```
lalo: It doesn't matter that the same port number is being used? I'm guessing one of them is going to fail either way since localhost can only map to one so its a moot point?

chwarr: I'm pretty sure that the IPv4 and IPv6 port spaces are distinct and can be bound independently. Let me validate.

lalo: We talked in person, this is a non-issue.

PR #334, opened 2017-02-22
[c# epoxy] Safely shutdown possibly null sockets

chwarr:
If the function that produces the socket itself throws, we may not have a
socket to call Shutdown and Close on. This fixes a regression introducted by
2f9aeae.

### DIFF
--- a/cs/src/comm/epoxy-transport/EpoxyNetworkStream.cs
+++ b/cs/src/comm/epoxy-transport/EpoxyNetworkStream.cs
@@ -369,8 +369,8 @@ namespace Bond.Comm.Epoxy
         {
             try
             {
-                socket.Shutdown(SocketShutdown.Both);
-                socket.Close();
+                socket?.Shutdown(SocketShutdown.Both);
+                socket?.Close();
             }
             catch (SocketException ex)
             {


cs/src/comm/epoxy-transport/EpoxyNetworkStream.cs
```diff
@@ -369,8 +369,8 @@ static void SafeShutdownSocket(Socket socket, Logger logger)
         {
             try
             {
-                socket.Shutdown(SocketShutdown.Both);
-                socket.Close();
+                socket?.Shutdown(SocketShutdown.Both);
+                socket?.Close();
```
lalo: Don't we need to add this also on lines 246 and 247 or only on the safe shutdown?

Also I just realized that this method and a couple of other ones are missing the summary comments.

chwarr: We only need this for SafeShutdownSocket. EpoxyNetworkStream.Shutdown can only be called on constructed object, which have non-null Socket instances. SafeShutdownSocket is used when we're creating the socket in the first place, to then pass to the EpoxyNetworkStream ctor.

I'd rather not add summary comments to these deep internal methods in this PR.

chwarr:
The failed AppVeyor build actually built and passed, but was right on the 1 hour limit.

PR #339, opened 2017-03-01
Add CMake option to install generated .cpp files

chwarr:
The CMake cache variable `BOND_LIBRARIES_INSTALL_CPP` controls whether the
generated .cpp files are installed under src/. This can be used to keep the
code around for debugging purposes or to create a self-contained copy of
Bond that can be built with something other than CMake.


chwarr:
CI builds failed, as we couldn't install Pandoc. I've opened issue https://github.com/Microsoft/bond/issues/341 for this.

PR #344, opened 2017-03-03
[c++] Avoid name shadowing warnings

chwarr:
Bond's warning.h header was suppressing warnings about names shadowing
each other. Those have been removed, and renames made.

TODO

* [x] Confirm that we're OK reserving the prefix "br6c5"
    * We're not. Instead, generate non-conflicting names.
* [x] <strike>Make gbc reject identifiers that start with this prefix</strike>
* [x] Test with other compilers. (This PR will start that process.)
* [x] Add QuickCheck property tests for `uniqueName`
* [ ] Determine whether this is a major change for C++ Bond

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,12 +12,18 @@ different versioning scheme, following the Haskell community's
 [package versioning policy](https://wiki.haskell.org/Package_versioning_policy).
 
 ## Unreleased ##
-* `gbc` & compiler library: TBD
+* `gbc` & compiler library: TBD (minor bump needed)
 * IDL core version: TBD
 * IDL comm version: TBD
-* C++ version: TBD
+* C++ version: TBD (major bump needed)
 * C# NuGet version: TBD
 * C# Comm NuGet version: TBD (minor bump needed)
+
+### C++ ###
+* **Breaking change:** Suppression of MSVC name shadowing warnings is no
+  longed needed, and Bond's warning.h header no longed supresses C4456,
+  C4458, or C4459. This may cause these warnings to now be triggered on
+  other code.
 
 ### C# Comm ###
 
--- a/compiler/bond.cabal
+++ b/compiler/bond.cabal
@@ -79,6 +79,7 @@ test-suite gbc-tests
   hs-source-dirs:   tests, .
   main-is:          Main.hs
   other-modules:    Tests.Codegen
+                    Tests.Codegen.Util
                     Tests.Syntax
   ghc-options:      -threaded -Wall
   build-depends:    bond,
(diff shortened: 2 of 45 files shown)


chwarr:
When we're generating the C++ structs (the _types code), we generate some member functions as well. If the name we've picked for the parameter (`other`, `left`, or `right`) matches the name of a field, we'll get a shadowing warning. To resolve this, the design I came up with was to prefix all the parameters in the generated member functions with "br6c5". This is short and unlikely to match any existing field name prefiexes in use today. The "br" stands for "Bond reserved" and the "6c5" is just some random chars. I'm open to other names. The following won't work:

* `_bond`, as there are schemas with fields like `_bond_payload` (Some people prefix members with an `_`
* `__anything` is reserved by C++
* `_CapitalAnything` is reserved by C++
* `p_` is likely to conflict with existing schemas

I _guess_ we could find three already reserved words and use those... Of the reserved identifiers, here's the ones that are likely to be safe in C++:

* bond_meta
* import
* optional
* required
* required_optional
* sealed (maybe not because C++/CLI?)
* var
* view_of

If we pick from these, we're going need a comment _in the generated code_ to explain what's going on...

Anyone else have a better solution to this problem? If not, a better prefix?

sapek:
This seems to me like a rather ugly solution: everyone will get less readable generated code (affecting not just reading code but also debugging, etc) in order to avoid what must be a pretty rare conflict causing a warning. If it was the case of invalid generated code that doesn't compile that would be another story, but as I understand it is merely a warning, and a warning that doesn't really indicate anything wrong with the code.

IMO it is fine to leave it to the user. They have a choice to either live with the warning, suppress it or avoid these identifiers in their schema.

If you really want to be heroic and avoid this warning, I think a cleaner solution would be to use non-conflicting names in the generated code *only* if there is a conflict. You could disambiguate with numeric suffix to the name.

chwarr:
`left0`, `left1` is a _much_ better idea. Thanks!

CHANGELOG.md
```diff
@@ -12,13 +12,19 @@ different versioning scheme, following the Haskell community's
...
+* `gbc` & compiler library: TBD (minor bump needed)
 * IDL core version: TBD
 * IDL comm version: TBD
-* C++ version: TBD
+* C++ version: TBD (major bump needed)
 * C# NuGet version: TBD
 * C# Comm NuGet version: TBD (minor bump needed)
 
+### C++ ###
+* **Breaking change:** Suppression of MSVC name shadowing warnings is no
```
chadwalters: Given this, I wonder if we want to bundle this together with more of the warning un-suppression work, Thoughts?

chwarr: That makes sense. Let me move this to a branch that we can both work in.

chadwalters: OK sounds good.

chwarr: Moved to https://github.com/Microsoft/bond/tree/clean-msvc-warnings. Let's clean up the warnings there and then merge the whole branch in at once.

PR #349, opened 2017-03-07
Update thread_service.h

zhenqinshang:
Dead bug


chwarr:
Chad, you made this change originally. Can you take a look?

chadwalters:
zhenqinshang, can you explain what problem you ran into? The point here is that you can't join the thread you are currently on. The change you are proposing will cause that to happen. It could be that there is some other problem here though so if you can provide more info or a small snippet that reproduces the problem, that would be helpful.

chadwalters:
I am going to close this pull request for now until there is more information about the problem being addressed.

Issue #342, opened 2017-03-02
gbc-tests does not build with derive 2.6.1 and ghc 8.0.2


chwarr:
As of derive 2.6.1, building the gbc-tests fails. I have not yet looked into this in detail. Is gbc-tests's use of derive incorrect? Has derive had a regression? Is derive a victim of some other package?

```
3/2/2017 11:42:59 AM :: chwarr@CHWARR-DESK :: S:\src\bond\compiler
  [master] [5] > cabal test -j
Preprocessing library bond-0.8.0.0...
Preprocessing test suite 'gbc-tests' for bond-0.8.0.0...
[5 of 6] Compiling Tests.Syntax     ( tests\Tests\Syntax.hs, dist\build\gbc-tests\gbc-tests-tmp\Tests\Syntax.o ) [Data.DeriveTH changed]

tests\Tests\Syntax.hs:1:1: error:
    Exception when trying to run compile-time code:
      Could not convert Decl () to Dec
InstDecl () Nothing (IRule () Nothing (Just (CxTuple () [])) (IHApp () (IHCon () (UnQual () (Ident () "Arbitrary"))) (TyCon () (UnQual () (Ident () "Attribute"))))) (Just [InsDecl () (PatBind () (PVar () (Ident () "arbitrary")) (UnGuardedRhs () (Do () [Generator () (PVar () (Ident () "x1")) (Var () (UnQual () (Ident () "arbitrary"))),Generator () (PVar () (Ident () "x2")) (Var () (UnQual () (Ident () "arbitrary"))),Qualifier () (App () (Var () (UnQual () (Ident () "return"))) (Paren () (App () (App () (Con () (UnQual () (Ident () "Attribute"))) (Var () (UnQual () (Ident () "x1")))) (Var () (UnQual () (Ident () "x2"))))))])) Nothing)])
src\Language\Haskell\Convert.hs:(124,5)-(132,57): Non-exhaustive patterns in function conv

CallStack (from HasCallStack):
  error, called at src\Language\Haskell\Convert.hs:21:87 in derive-2.6.1-GF6BDS0jI0b5Hi4GHVjtNF:Language.Haskell.Convert
    Code: derive makeArbitrary ''Attribute
```

Here's the full package list that's causing this for me:

```
3/2/2017 11:46:00 AM :: chwarr@CHWARR-DESK :: S:\src\bond\compiler
  [master] [7] > cabal sandbox hc-pkg list
[...]

PR #351, opened 2017-03-08
[gbc] Temporarily pin derive dependency to < 2.6

chwarr:
gbc-tests no longer builds with derive 2.6 or later. While we
investigate that problem, pin the dependency to < 2.6 to keep gbc-tests
building.

Mitigates https://github.com/Microsoft/bond/issues/342


chadwalters:
CI Failures unrelated. Looks good.

Issue #341, opened 2017-03-01
Documentation builds are failing


chwarr:
Recent documentation builds for pull requests have been failing.

An [example](https://ci.appveyor.com/project/sapek/bond/build/1.0.1249/job/j0j18kqw9cy3uri0) from PR #339:

```
Preprocessing library http-client-tls-0.3.4...
Network\HTTP\Client\TLS.hs:51:18:
    Ambiguous module name `Network.URI':
      it was found in multiple packages:
      network-uri-2.6.1.0 network-2.4.2.3
Installed texmath-0.8.6.7
cabal : cabal.exe: Error: some packages failed to install:
At line:15 char:5
+     cabal install -j pandoc --constraint='mtl<=2.1.3.1'
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (cabal.exe: Erro...led to install::String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
http-client-tls-0.3.4 failed during the building phase. The exception was:
ExitFailure 1
pandoc-1.17.2 depends on http-client-tls-0.3.4 which failed to install.
```

PR #353, opened 2017-03-09
Use precompiled pandoc for building documentation

chwarr:
Instead of building pandoc from source, we now just use Chocolately to
install a precompiled version of pandoc. This speeds up documentation
builds and prevents random breakage due to Hackage package changes.

NB: The Chocolately Pandoc 1.19.2.1 package is currently broken. It
appears to have a pre-release Pandoc 2.0 executable that doesn't
understand all the command line flags that 1.x does (e.g., --smart).

Fixes https://github.com/Microsoft/bond/issues/341

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -69,19 +69,29 @@
 
             choco install haskellplatform -version 2014.2.0.0 -y
 
-            $env:Path = [System.Environment]::GetEnvironmentVariable("Path","Machine")
+            # choco install updated the path, so re-read them from the registry and reset $env:path
+
+            $userPath = [System.Environment]::GetEnvironmentVariable("Path","User")
+
+            $machinePath = [System.Environment]::GetEnvironmentVariable("Path","Machine")
+
+            $env:Path = "$userPath;$machinePath"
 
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -69,19 +69,29 @@
...
 
             if ($env:BOND_BUILD -eq "Doc") {
 
-                cabal install -j pandoc --constraint='mtl<=2.1.3.1'
+                choco install pandoc --version 1.19.2  -y
 
                 choco install doxygen.install -y
 
-            }
+                $userPath = [System.Environment]::GetEnvironmentVariable("Path","User")
```
chadwalters: Should these lines move inside the if-block or does cabal update also change the path?

chwarr: I moved them inside the if block. `cabal update` doesn't change the path. It just downloads the Hackage package list.

chwarr:
I fixed a problem with the order the path was being constructed in. Please squash when mering now.

chadwalters:
CI failures are not related. Looks good.

PR #358, opened 2017-03-10
Extended bug fix in commit 3294b3 to SimpleJsonReader(Stream) constructor

maschram:


chwarr:
The fix looks fine to me, Can you also add a test case or two that would fail without your fix but pass with it?

maschram:
Added unit test

cs/test/core/JsonParsingTests.cs
```diff
@@ -703,5 +703,25 @@ static T ParseJson<T>(string json) where T : new()
...
         }
+
+        [Test]
+        public void JsonParsing_DateAsStringInStreamConstructor()
+        {
+            string timestamp = "2017-03-07T04:47:50.1145227Z";
+            string json = $"{{\"timestamp\":\"{timestamp}\"}}";
+
+            DateAsString target;
+            using (MemoryStream ms = new MemoryStream())
```
chwarr: How about the following instead?

```
var ms = new MemoryStream(Encoding.UTF8.GetBytes(json));
var reader = new SimpleJsonReader(ms);
var target = Deserialize<DataAsString>.From(reader);
Assert.AreEqual(timestamp, target.timestamp);
```

We can also skip disposing the MemoryStream. MemoryStream.Dispose doesn't do anything interesting.

chwarr: The Travis CI build failed for as the `reader` constructed here never gets used. My proposed change will fix this problem as well.

chwarr:
LGTM. I'll merge after the CI builds pass.

PR #363, opened 2017-03-15
Improve Bond's compliance with Microsoft's SDL

chwarr:
For projects that use Microsoft's SDL, these commits help minimize the amount of work needed to use Bond.

As a bonus, the changes are relatively contained, and don't materially affect Clang or GCC.

These commits should be rebased atop master, or a full merge should be done.

### DIFF
--- a/examples/cpp/core/bf/err.h
+++ b/examples/cpp/core/bf/err.h
@@ -1,0 +1,29 @@
+#pragma once
+
+#include <string.h>
+#include <string>
+
+// Convert an error number into a human-readable string.
+inline std::string ErrorString(int errnum)
+{
+    std::string result;
+
+#ifdef _MSC_VER
+    // strerrorlen_s hasn't made it into the Microsoft C runtime yet. Until
+    // it has, we're just going to reserve a buffer that's pretty big and
+    // get as much of the error message as we can.
+    result.reserve(240);
+
+    (void)strerror_s(&result[0], result.size(), errnum);
+
+    // strerror_s wrote an embedded NUL, so truncate to that size so the
+    // result doesn't have an embedded NUL.
+    result.resize(strlen(result.c_str()));
+#else
+    // strerror isn't guaranteed to be thread-safe, but bf is single
+    // threaded.
+    result = strerror(errnum);
+#endif
+
+    return result;
+}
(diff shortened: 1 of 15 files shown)


chwarr:
I've broken something somewhere. I'm taking a look.

chwarr:
Fixes for Clang & GCC and VC++ 2013 have been integrated. They've been rebased in, but no one has yet reviewed and left comments, so I force pushed. If you want to see older revisions, let me know the commit hash, and I'll resurrect it.

examples/cpp/core/multiprecision/CMakeLists.txt
```diff
@@ -8,3 +8,7 @@ add_bond_codegen (multiprecision.bond
 add_bond_test (multiprecision
     ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/multiprecision_types.cpp
     multiprecision.cpp)
+
+# Deep in the guts of Boost we get a warning about not using a "secure"
+# routine. Suppress them for this example.
+cxx_target_compile_options(MSVC multiprecision PRIVATE -D_SCL_SECURE_NO_WARNINGS)
```
chadwalters: We may need to track this down at some point if Bond is pulling this in for our customers. If there is some specific behavior in this example that pulls it in unrelated to the core of Bond then I am less concerned about it.

examples/cpp/core/bf/err.h
```diff
@@ -0,0 +1,29 @@
...
+// Convert an error number into a human-readable string.
+inline std::string ErrorString(int errnum)
+{
+    std::string result;
+
+#ifdef _MSC_VER
+    // strerrorlen_s hasn't made it into the Microsoft C runtime yet. Until
+    // it has, we're just going to reserve a buffer that's pretty big and
+    // get as much of the error message as we can.
+    result.reserve(240);
```
tstein: I feel like you can make this bigger if you want to.

examples/cpp/core/multiprecision/CMakeLists.txt
chadwalters: We discussed and this doesn't seem like a big concern.

chadwalters:
CI failure on VC14 build is not related to this change. Looks good to go.

examples/cpp/core/multiprecision/CMakeLists.txt
chwarr: Yeah, the Boost stuff emitting a warning is being pulled in by the example in question, not by Bond.

PR #370, opened 2017-03-20
[c++] Remove needless std::string creation in bf

chwarr:
A bare, observing `const char*` is fine for the OpenFile helper
function, as that's what the underlying `fopen` API uses.


chwarr:
CI failures are timeouts.

PR #371, opened 2017-03-20
[c++] Sort OutputBuffer init list and members uniformly

chwarr:
This make the init list order match the real order of initialization,
avoiding a warning from Clang/GCC.


chwarr:
Grumble... CI failures are timeouts...

PR #373, opened 2017-03-20
[c++] pre-generate missing protocols

veselinp:
- generated files *_apply.h/cpp are missing explicit Apply<>
  instantiations for CompactBinaryWriter<OutputCounter> and
  SimpleBinaryWriter<Null> protocols.

### DIFF
--- a/compiler/src/Language/Bond/Codegen/Cpp/ApplyOverloads.hs
+++ b/compiler/src/Language/Bond/Codegen/Cpp/ApplyOverloads.hs
@@ -16,10 +16,8 @@ import Language.Bond.Codegen.Util
 -- | Protocol data type is used to specify what protocols the @Apply@ function
 -- overloads should be generated for.
 data Protocol =
-    Protocol
-    { protocolReader :: String -- ^ Name of the class implementing the protocol reader.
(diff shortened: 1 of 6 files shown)


compiler/src/Language/Bond/Codegen/Cpp/ApplyOverloads.hs
```diff
@@ -16,10 +16,8 @@ import Language.Bond.Codegen.Util
 -- | Protocol data type is used to specify what protocols the @Apply@ function
 -- overloads should be generated for.
 data Protocol =
-    Protocol
-    { protocolReader :: String -- ^ Name of the class implementing the protocol reader.
-    , protocolWriter :: String -- ^ Name of the class implementing the protocol writer.
-    }
+    ProtocolReader String | -- ^ Name of the class implementing the protocol reader.
```
chwarr: For the CHANGELOG: this is a major version change for the "bond" Haskell library.

veselinp: I've added CHANGELOG statement and fixes the title.

PR #374, opened 2017-03-21
Remove FieldTemplate<> from field reflection type

veselinp:
- FieldTemplate<> params are part of field relfection type, this unnecessary increases  generated symbol names.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,9 @@ get a compiler error. To fix, remove the `<Writer>` part:
 
 ### `gbc` and Bond compiler library ###
 
-* Remove FieldTemplate details from field reflection type.
+* C++ codegen generates field reflection, that uses an explicit type 
+  derrived from reflection::FieldTemplate. This hides FieldTemplate 
+  details and, as result, shortens symbol names.
 * C++ codegen ensures that parameter names do not shadow field names.
 * When generating C++ apply files, there are now explicit `bond::Apply<>`
   instantiations for `CompactBinaryWriter<OutputCounter>` and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,9 +45,7 @@ get a compiler error. To fix, remove the `<Writer>` part:
 
 ### `gbc` and Bond compiler library ###
 
-* C++ codegen generates field reflection, that uses an explicit type 
-  derrived from reflection::FieldTemplate. This hides FieldTemplate 
-  details and, as result, shortens symbol names.
+* C++ codegen hides FieldTemplate details, shortening symbol names.
 * C++ codegen ensures that parameter names do not shadow field names.
 * When generating C++ apply files, there are now explicit `bond::Apply<>`
   instantiations for `CompactBinaryWriter<OutputCounter>` and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ get a compiler error. To fix, remove the `<Writer>` part:
 
 ### `gbc` and Bond compiler library ###
 
+* Remove FieldTemplate details from field reflection type.
 * C++ codegen ensures that parameter names do not shadow field names.
 * When generating C++ apply files, there are now explicit `bond::Apply<>`
   instantiations for `CompactBinaryWriter<OutputCounter>` and
--- a/cpp/inc/bond/core/reflection.h
+++ b/cpp/inc/bond/core/reflection.h
@@ -93,7 +93,7 @@ struct FieldTemplate
     static const Metadata& metadata;
 
     /// @brief Static data member representing the field pointer
-    static const field_pointer field;
+    static constexpr field_pointer field = field_ptr;
 
     /// @brief Static data member equal to the field ordinal
     BOND_STATIC_CONSTEXPR uint16_t id = field_id;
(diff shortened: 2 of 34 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ different versioning scheme, following the Haskell community's
 * C# NuGet version: TBD (minor bump needed)
 * C# Comm NuGet version: TBD (minor bump needed)
 
+### C++ & `gbc` ###
+
+* Remove FieldTemplate details from field reflection type.
+
```
chwarr: Minor: there's a `gbc` section right below where this would make more sense. Consider also duplicating the entry in the C++ section as well.

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ different versioning scheme, following the Haskell community's
 * C# NuGet version: TBD (minor bump needed)
 * C# Comm NuGet version: TBD (minor bump needed)
 
+### C++ & `gbc` ###
+
+* Remove FieldTemplate details from field reflection type.
```
chwarr: Minor: can you expand on this and say why we're making the change (to shorten the symbol names)?

veselinp:
I've pushed a fixup, clang should be happy now.

chadwalters:
clang build is happy. Looks like VC12 is still broken though.

CHANGELOG.md
veselinp: Done!

CHANGELOG.md
veselinp: Done!

veselinp:
vc12 doesn't understand constexpr :)

veselinp:
now, clang is broken, I'm looking again

cpp/test/core/container_extensibility.cpp
```diff
@@ -123,8 +123,8 @@ TEST_CASE_BEGIN(MultiIndexTest)
                 SimpleStructView, 
                 boost::multi_index::indexed_by<
                     boost::multi_index::sequenced<>,
-                    ordered_non_unique_field<SimpleStructView::Schema::var::m_str>,
-                    ordered_non_unique_field<SimpleStructView::Schema::var::m_uint64>
+                    ordered_non_unique_field<SimpleStructView::Schema::var::m_str, get_field_template<SimpleStructView::Schema::var::m_str>::type>,
+                    ordered_non_unique_field<SimpleStructView::Schema::var::m_uint64, get_field_template<SimpleStructView::Schema::var::m_uint64>::type>
```
chadwalters: Can you explain this part of the change?

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ different versioning scheme, following the Haskell community's
 
 ### `gbc` and Bond compiler library ###
 
+* C++ codegen generates field reflection, that uses an explicit type 
+  derrived from reflection::FieldTemplate. This hides FieldTemplate 
+  details and, as result, shortens symbol names.
```
chadwalters: Minor: Change to "C++ codegen hides FieldTemplate details, shortening symbol names."

veselinp: Done.

cpp/test/core/container_extensibility.cpp
veselinp: I simplified this, and moved all logic into ordered_non_unique_field<>, please check cpp/test/core/multi_index.h

vc12 doesn't support constexp - ordered_non_unique_field requires member field address in compile time, which can be resolved via FieldTemplate.

chwarr:
I will look again later today.

cpp/inc/bond/core/reflection.h
```diff
@@ -101,24 +101,21 @@ struct FieldTemplate
     /// @brief Static data member describing field metadata
     static const Metadata& metadata;
 
-    /// @brief Static data member representing the field pointer
-    static const field_pointer field;
-
```
chwarr: This looks like a breaking change (removal of public member of public interface). Is it intentional? If so, what's motivating making this breaking change?

It appears that the replacement for this functionality is `&theFieldType.GetVariable(someStruct)`.

python/inc/bond/python/struct.h
```diff
@@ -26,14 +26,14 @@ class def_readwrite_property
         : c(c)
     {}
 
-    template <typename Field>
-    void operator()(const Field&) const
+    template <uint16_t field_id, typename ModifierTag, typename Struct, typename FieldType, FieldType Struct::*field_ptr, const bond::Metadata* metadata_ptr>
```
chwarr: Was this change required to get this to build?

* If so, then the introduction of the anonymous struct is not working as I'd expect it to (the anonymous struct should have an _is a_ relationship with `FieldTemplate`), and is a large breaking change.
* If this was just done to clean up the code, that's fine, but I need to get confirmation from you.

cpp/inc/bond/core/reflection.h
veselinp: Good point, let me revisit this.

chwarr: If this materially reduces the lib and PDB size, then it's something we can consider. The next release for C++ will have some other breaking changes (albeit minor ones) like this.

python/inc/bond/python/struct.h
veselinp: This change is related to FieldTemplate<>::field change.

cpp/inc/bond/core/reflection.h
veselinp: I added it back for now, we may want to handle vc12 explicitly to remove this symbol from generated .obj files.

Issue #305, opened 2017-01-18
Alignment issues in serializers


codeandroid:
We began using Bond in a project compiled with emscripten. [1]

Running deserializers for CompactBinary, FastBinary and SimpleJson always resulted in an "alignment fault" when executing the resulting asm.js code.

The following exemplary piece of code demonstrates the issue and a possible fix. (There are other similar pieces of code here and there.)

It is located in `cpp/inc/bond/stream/input_buffer.h`:

```
template <typename T>
void Read(T& value)
{
    if (sizeof(T) > _blob.length() - _pointer)
    {
        EofException(sizeof(T));
    }
    
    // ORIGINAL
    //value = *reinterpret_cast<const T*>(_blob.content() + _pointer);

    // FIX
    std::memcpy(&value, _blob.content() + _pointer, sizeof(T));

    _pointer += sizeof(T);
}
```

Emscripten is a platform requiring loads and stores to be aligned (like MIPS, PowerPC, etc. but not like x86 which allows unaligned access though not without impact).

Also, I think, the original code violates the C++ strict aliasing rules...

Using `std::memcpy` should be safe and most compilers should detect the use of `std::memcpy` and generate properly optimized code.

I do not understand this part of C++ deep enough to be a 100% sure that this is the right approach as I'm developing for x86 and x86_64 where unaligned access _just works_. Otherwise I would've prepared a PR already. :)

I also tested SimpleJson (when the others failed) and also ran into alignment issues but this time most likely due to RapidJson itself. A fix for RapidJson ( https://github.com/miloyip/rapidjson/pull/353 ) has been merged into `master` and should be included in the next release ( http://rapidjson.org/md__c_h_a_n_g_e_l_o_g.html ).

->

If you could provide some guidance on how to deal with this issue, whether that's the right direction or whether there's a different solution... that'd be much appreciated.

[1] This is part of an experimental web app sharing code and data between our desktop applications and existing services talking Bond.Comm.

chwarr:
`std::memcpy` is the C++ standards compliant way to do these sort of type punning conversions. We've made similar fixes elsewhere in the code when we've found them. In particular, I in commit bc4f7d881acbf5e8ed703e56302a3c77f3a3c522 I made a very similar change in [random_protocol.h](https://github.com/Microsoft/bond/commit/bc4f7d881acbf5e8ed703e56302a3c77f3a3c522#diff-bcc17120306854587c556e415c080e0aR95).

[...]

PR #375, opened 2017-03-22
[c++] Use memcpy to type pun; fix alias violations

chwarr:
We now use `memcpy` to pun from one type to another, avoiding aliasing
violations as well as alignment errors. x86-like chips never failed due
to the alignment errors, but others, like some ARM chips, did. This
should also fix emscripten cross-compilation.

Fixes https://github.com/Microsoft/bond/issues/305

### DIFF
--- a/cpp/inc/bond/stream/input_buffer.h
+++ b/cpp/inc/bond/stream/input_buffer.h
@@ -5,7 +5,9 @@
 
 #include <bond/core/blob.h>
 #include <bond/core/exception.h>
+#include <bond/core/traits.h>
 #include <boost/static_assert.hpp>
+#include <cstring>
 
 namespace bond
 {
@@ -21,7 +23,7 @@ struct VariableUnsignedUnchecked
     {
         T byte = *p++;
         value |= (byte & 0x7f) << Shift;
-        
+
         if (byte >= 0x80)
             VariableUnsignedUnchecked<T, Shift + 7>::Read(p, value);
     }
@@ -36,7 +38,7 @@ struct VariableUnsignedUnchecked<T, 0>
     {
         T byte = *p++;
         value = (byte & 0x7f);
-        
+
         if (byte >= 0x80)
             VariableUnsignedUnchecked<T, 7>::Read(p, value);
     }
@@ -84,14 +86,14 @@ struct VariableUnsignedUnchecked<uint64_t, 56>
 }
 
 /// @brief Memory backed input stream
-class InputBuffer 
+class InputBuffer
 {
 public:
     /// @brief Default constructor
     InputBuffer()
         : _pointer()
     {}
-    
+
     /// @brief Construct from a blob
     ///
     /// InputBuffer makes a copy of the blob. Assuming that the blob was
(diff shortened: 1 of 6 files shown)


cpp/inc/bond/stream/input_buffer.h
```diff
@@ -126,38 +128,42 @@ class InputBuffer
...
+
         if (sizeof(T) > _blob.length() - _pointer)
         {
             EofException(sizeof(T));
         }
-        
-        value = *reinterpret_cast<const T*>(_blob.content() + _pointer);
+
+        const void* const src = _blob.content() + _pointer;
+        std::memcpy(&value, src, sizeof(T));
```
chwarr: TODO: see whether on x86 there's any performance regression due to this change in reading. If there is, guard the unaligned access like is done in output_buffer.

cpp/inc/bond/stream/output_buffer.h
```diff
@@ -19,18 +21,19 @@ struct VariableUnsignedUnchecked
 {
     BOOST_STATIC_ASSERT(N < 10);
 
-    static uint32_t Write(uint8_t* p, T value)
+    static uint32_t Write(char* p, T value)
```
chwarr: There was no measurable performance implication due to these changes when I tested.

cpp/inc/bond/stream/output_buffer.h
```diff
@@ -165,15 +171,16 @@ class OutputMemoryStream
...
     template<typename T>
     void Write(const T& value)
     {
+#if defined(__x86_64__) || defined(_M_X64) || defined(__i386) || defined(_M_IX86)
         //
-        // Performance tweak: for small types don't go into generic Write(),
-        // which results in call to memcpy() and additional memory access.
-        // The direct copy of the value (which is likely on register already)
-        // is faster.
+        // x86/x64 performance tweak: for small types don't go into generic
```
chwarr: I re-validated that this tweak still results in better performance--at least with MSVC.

cpp/inc/bond/stream/output_buffer.h
```diff
@@ -298,7 +314,7 @@ class OutputMemoryStream
     {
         if (sizeof(T) * 8 / 7 + _rangeSize + _rangeOffset < _bufferSize)
         {
-            uint8_t* ptr = reinterpret_cast<uint8_t*>(_rangePtr + _rangeSize);
+            char* ptr = _rangePtr + _rangeSize;
```
chwarr: Minor: can be const

chwarr:
@glenfe, if you have time, I would appreciate if you could take a look at these changes. You're knowledge of the C++ standards is far better than mine. Thanks in advance.

glenfe:
Looks fine to me.

When reviewing the changes (but unrelated to your change) I noticed some things with the use of allocators: I'll create a proper Issue about it tomorrow.

cpp/inc/bond/stream/output_buffer.h
chwarr: The _pointer_ can be const, but the thing it points to cannot. Not worth the potential confusion. Leaving as-is.

cpp/inc/bond/stream/input_buffer.h
chwarr: I did some testing, and the `reinterpret_cast` code path is faster for amd64 MSVC, so I've added it back.

chwarr:
Official AppVeyor builds are completely backed up due to some work @tstein is doing to improve the coverage we have on Travis for versions of Boost. (He's going to remove appveyor.yml in a subsequent change so that this doesn't keep happening.)

This change passed on [my AppVeyor](https://ci.appveyor.com/project/chwarr/bond/build/1.0.322) config, with the usual timeout in the VC14 C++ build during tests. I ran all of these tests on my local machine.

Going to merge.

PR #378, opened 2017-03-22
Add boost-builder.

tstein:


tools/boost-builder/build_boosts.zsh
```diff
@@ -0,0 +1,30 @@
...
+  wget https://downloads.sourceforge.net/project/boost/boost/$VERSION_DOTTED/boost_$VERSION.tar.bz2
+  tar -xf boost_$VERSION.tar.bz2
+  cd boost_$VERSION
+
+  # --with-python needed for boost < 1.62
+  ./bootstrap.sh --exec-prefix=$INSTALL_ROOT/boost_$VERSION \
+                 --includedir=$INSTALL_ROOT/boost_$VERSION/include \
+                 --with-libraries=$BOOST_LIBRARIES \
+                 --with-python=python2.7
+  ./b2 -j $NUMPROCS install
```
chwarr: Does this build 32- or 64-bit? (Or both?) We care more about 64-bit.

tools/boost-builder/Dockerfile
```diff
@@ -0,0 +1,9 @@
+# Does this match Travis's trusty?
+FROM ubuntu:trusty
+
+ADD build_boosts.zsh /root/
+
+RUN apt-get update
+RUN apt-get -y install build-essential python2.7-dev wget zsh
```
chwarr: What version of the compilers does `build-essential` pull in for trusty?

tools/boost-builder/Dockerfile
```diff
@@ -0,0 +1,9 @@
+# Does this match Travis's trusty?
+FROM ubuntu:trusty
```
tstein: Changed tags from trusty to 14.04.5, which does match travis exactly. (Confirmed based on output at the beginning of the CI run for this PR.)

tools/boost-builder/Dockerfile
tstein: It pulls in gcc-4.8. Our actual CI uses clang.

I don't think it's particularly important that these match, but this container should try to use the same compiler that Ubuntu uses to build Trusty packages. It's very likely that they are using whatever build-essential provides, but I will look for proof of that.

tools/boost-builder/Dockerfile
tstein: Changed from tag trusty to 14.04.5, which matches Travis exactly.

tools/boost-builder/build_boosts.zsh
tstein: Host architecture. Docker only supports x64, so that's all this container is going to produce.

Don't know about optimization. Let me take a look at the source of the Trusty boost deb.

tstein:
Planning to bounce these tars through Azure blob storage. Credential management is slightly easier if we don't do it inside the container, so I think I'll just write another script for that.

tools/boost-builder/build_boosts.zsh
tstein: Don't see any optimization in the deb recipe. We'll have whatever b2's default it, which appears to be what Ubuntu is shipping.

tools/boost-builder/Dockerfile
tstein: Checked a few random system libs and they're compiled with the exact gcc that build-essential ships. I'm satisfied; happy to keep digging if you aren't.

chwarr: Good enough for me.

chwarr:
OK. We can use Travis's secret management to plumb an Azure storage key into the CI builds.

tstein:
This is actually, finally ready for final review. In summary, this change:
* adds a docker container, which builds boost binaries for versions 1.58 through 1.63 and spits out tarballs locally
* adds an upload script, which puts those binaries into an Azure blob storage bucket
* modifies our travis config to pull down those binaries and to build core against all of them
* builds comm once against the latest boost
* removes explicit references to python in the travis config - that's covered by C++
* cleans up unused parts of the travis config

chwarr:
It looks like some generated files didn't get generated or put where we expected them. From the [Comm build](https://travis-ci.org/Microsoft/bond/jobs/216091094):

```
Scanning dependencies of target generic_service
[ 90%] Building CXX object cpp/test/comm/CMakeFiles/generic_type.dir/generic_type.cpp.o
[ 91%] Building CXX object cpp/test/comm/CMakeFiles/generic_service.dir/generic_service.cpp.o
/home/travis/build/Microsoft/bond/cpp/test/comm/generic_type.cpp:8:10: fatal error: 
      'generic_reflection.h' file not found
#include <generic_reflection.h>
         ^
/home/travis/build/Microsoft/bond/cpp/test/comm/generic_service.cpp:8:10: fatal error: 
      'generic_reflection.h' file not found
#include <generic_reflection.h>
         ^
1 error generated.
make[3]: *** [cpp/test/comm/CMakeFiles/generic_type.dir/generic_type.cpp.o] Error 1
```

tstein:
This was working not two hours ago. 🤔 https://travis-ci.org/Microsoft/bond/builds/216089373

Taking a look.

tstein:
It worked on a re-run. Unfortunately, travis doesn't seem to preserve the log if you re-run?

That failure looks like a transient travis thing. I'd like to push as-is, both because it doesn't appear to be on our side and so we can get more runs with this config to look at.

chwarr:
Thanks for looking. Sounds reasonable. I'm reviewing right now.

tools/boost-builder/upload_boosts.zsh
```diff
@@ -0,0 +1,10 @@
+#!/bin/zsh
```
chwarr: Is the lack of `set -ex` intentional here?

.travis.yml
```diff
@@ -10,81 +10,75 @@ cache:
...
-    - BOND_GHC_VERSION=-7.8.4
-    - BOND_GHC_VERSION=-7.6.3
+    - BOND_LANG=cpp BOOST="1.63.0"
+    - BOND_LANG=cpp BOOST="1.62.0"
+    - BOND_LANG=cpp BOOST="1.61.0"
+    - BOND_LANG=cpp BOOST="1.60.0"
+    - BOND_LANG=cpp BOOST="1.59.0"
+    - BOND_LANG=cpp BOOST="1.58.0"
+    - BOND_LANG=cpp COMM="true"
+    - BOND_LANG=hs GHC_VERSION=7.8.4
```
chwarr: Is the difference in whether the values are quoted or not intentional?

.travis.yml
```diff
@@ -10,81 +10,75 @@ cache:
...
+        - wget
 
 before_install:
     - if [ "$TRAVIS_OS_NAME" == "linux" ]; then echo "Hardware:"; grep model\ name /proc/cpuinfo | uniq -c; free -m; fi
-    # select C++ compiler
+    # linux prereqisite packages
+    - if [ "$TRAVIS_OS_NAME" == "linux" ]; then wget --no-check-certificate https://www.cmake.org/files/v3.2/cmake-3.2.3-Linux-x86_64.tar.gz; fi
+    - if [ "$TRAVIS_OS_NAME" == "linux" ]; then tar -xf cmake-3.2.3-Linux-x86_64.tar.gz; fi
+    - if [ "$TRAVIS_OS_NAME" == "linux" ]; then export PATH=$PWD/cmake-3.2.3-Linux-x86_64/bin:$PATH; fi
+    # Defaults.
```
chwarr: Minor: "set defaults" matches the phrasing nearby.

.travis.yml
```diff
@@ -10,81 +10,75 @@ cache:
...
-    - if [ "$BOND_GHC_VERSION" != "" ]; then make gbc-tests; fi
-    - cd ..
     - if [ "$BOND_LANG" == "cs" ]; then xbuild /p:Configuration=Debug cs/cs.sln; fi
     - if [ "$BOND_LANG" == "cs" ]; then xbuild /p:Configuration=Fields cs/cs.sln; fi
     - if [ "$BOND_LANG" == "cs" ]; then mono NUnit.Runners.2.6.4/tools/nunit-console.exe -framework=mono-4.5 -labels cs/test/core/bin/debug/net45/Properties/Bond.UnitTest.dll cs/test/core/bin/debug/net45/Fields/Bond.UnitTest.dll cs/test/internal/bin/debug/net45/Bond.InternalTest.dll; fi
-    - if [ "$BOND_GHC_VERSION" != "" ]; then cd compiler; fi
-    - if [ "$BOND_GHC_VERSION" != "" ]; then ../build/compiler/build/gbc-tests/gbc-tests; fi
-
-before_cache:
-    - if [ "$BOND_GHC_VERSION" != "" ]; then rm -rf compiler/.cabal-sandbox; fi
```
chwarr: Did you intend to remove this `before_cache` section? (It looks like a `cd ..` is missing anyway...)

tools/boost-builder/upload_boosts.zsh
tstein: No - upload_boosts was simple enough that I never had a problem that prompted me to add it. Added.

.travis.yml
tstein: No. Let's get more quotes.

.travis.yml
tstein: Done.

.travis.yml
tstein: Yes. It dates back to a 2015 commit that adds that directory to caching and also empties it before caching, which doesn't make a lot of sense. I don't know of a reason not to cache the sandbox.

It works like this, so... I think there's no `cd` missing? :)

tstein:
Works, but comm times out. :(

chwarr:
LGTM. I merged this in.

Issue #379, opened 2017-03-23
Correct or improve C++11 and higher allocator support


glenfe:
To correctly support the C++ allocator model, or otherwise simplify allocator-aware code.

1. [x] **1.** When the C++11 allocator model is available (i.e. `BOND_NO_CXX11_ALLOCATOR` is not defined) to rebind an allocator type `A` for type `T`, should always use `std::allocator_traits<A>::template rebind_alloc<T>` as I did in the recent #174 nullable changes. This is because a valid C++ allocator is no longer required to provide a `rebind` member template. 
2. [x] **2.** The integral constant expression for the detection of the Default Allocator can be simplified by a writing trait `is_default_allocator<A>` that is the integral constant `true_type` when the given allocator is the Default Allocator, and `false_type` otherwise.

PR #380, opened 2017-03-23
Support C++11 and above allocator model for rebind

glenfe:
Address two issues in #379


chwarr:
AppVeyor failure is due to an occasional segfault in a Comm test that @veselinp is looking into.
The Travis failure appears to have been a timeout. Re-running.

Issue #384, opened 2017-03-28
Travis build fails "missing cabal"


lalo:
Travis builds with config with ghc 7.6.3 fail due to

```
cabal: Error: some packages failed to install:
cmdargs-0.10.16 failed during the configure step. The exception was:
user error (The package 'cmdargs' requires Cabal library version -any &&
>=1.18 but no suitable version is installed.)
CMake Error at cabal_build.cmake:33 (message):
```

For example: https://travis-ci.org/Microsoft/bond/builds/216078128

Solution:

http://stackoverflow.com/questions/35565955/cabal-install-refuses-to-install-module-because-of-missing-cabal-library

PR #385, opened 2017-03-28
[travis] Install Cabal library

lalo:
Fixes issue #384


chwarr:
LGTM. The commit authorship is a bit weird...

tstein:
I had to do the same thing in https://github.com/Microsoft/bond/pull/378. Nothing bad's gonna happen if we merge both, though.

lalo:
This is the commit message I'll put after squashing

[travis] Install Cabal library

- CI builds using GHC 7.6.3 will fail since appropriate Cabal library is missing
- This change makes sure to install Cabal
- Closes https://github.com/Microsoft/bond/issues/384

PR #394, opened 2017-04-04
boost-builder: Set -u in scripts.

tstein:
Was advising a friend on shell scripting and realized I missed an important option here. -u causes the shell script to terminate if you try to expand an undefined variable.


chadwalters:
AppVeyor timeout for VC12 build unrelated. Looks good.

PR #398, opened 2017-04-05
Remove unused methods

veselinp:
This affects custom fast-path transforms.


veselinp:
I am working on a bigger change that would simplify this change.

PR #404, opened 2017-04-08
[C++] Adding GetBuffer overloads

ara-ayvazyan:
This change adds non-`const` overloads of `GetBuffer` function in protocol readers and writers which helps make it possible to refactor the handling of buffers to remove the hard-coded built-in buffers.

### DIFF
--- a/cpp/inc/bond/protocol/compact_binary.h
+++ b/cpp/inc/bond/protocol/compact_binary.h
@@ -178,7 +178,7 @@ public:
     }
 
 
-    /// @brief Access to underlaying buffer
+    /// @brief Access to underlying buffer
     typename boost::call_traits<Buffer>::const_reference
     GetBuffer() const
     {
@@ -186,7 +186,7 @@ public:
     }
 
 
-    /// @brief Access to underlaying buffer
+    /// @brief Access to underlying buffer
     typename boost::call_traits<Buffer>::reference
     GetBuffer()
     {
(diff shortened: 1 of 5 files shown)


cpp/inc/bond/protocol/compact_binary.h
```diff
@@ -186,6 +186,14 @@ class CompactBinaryReader
     }
 
 
+    /// @brief Access to underlaying buffer
```
chwarr: Minor: "underlying buffer" here and elsewhere

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -108,6 +108,13 @@ class SimpleJsonReader
     {
         return _value == rhs._value;
     }
+
+    /// @brief Access to underlaying buffer
+    typename boost::call_traits<Buffer>::reference
```
chwarr: Add a const overload as well, to match the other protocols? Similalrly for simple_json_writer.

ara-ayvazyan: Didn't realize that `SimpleJsonReader` doesn't have `const` version of `GetBuffer`. Do you mean to add them for _all_ writers as well (none of them has now)?

chwarr: Just for the reader.

chwarr:
LGTM. Will wait for CI builds before merging.

PR #406, opened 2017-04-08
[C++] Marking throw-only functions with BOND_NORETURN

ara-ayvazyan:


cpp/inc/bond/core/transforms.h
```diff
@@ -377,7 +377,7 @@ class RequiredFieldValiadator
     {}
 
 private:
-    void MissingFieldException() const;
+    BOND_NORETURN void MissingFieldException() const;
```
chwarr: Consider marking both the declaration and the definition, in case some compilers need both or report a mismatch.

cpp/inc/bond/core/maybe.h
```diff
@@ -9,7 +9,7 @@ namespace bond
 {
 
 
-void NothingException();
+BOND_NORETURN void NothingException();
```
chwarr: Do you know why we declare this instead of `#include "exception.h"`?

ara-ayvazyan: Not sure (maybe there was no exception.h at that time?). I can fix that.

cpp/inc/bond/core/transforms.h
ara-ayvazyan: Are you sure that is needed specifically for `[[noreturn]]`? It should not impact the definition/implementation.

chwarr: Looks like Clang, GCC, and MSVC only need `[[noreturn]]` or `__declspec(noreturn)` or `__attribute__((noreturn))` on the declaration. I was confusing these with SAL annotations, which get weird if the declaration and definition are different.

cpp/inc/bond/core/maybe.h
ara-ayvazyan: I'll check if this is done to avoid recursive includes.

ara-ayvazyan: Verified. This is done to avoid recursive includes.

chwarr: Got it. Thanks.

chwarr:
Going to close and re-open to re-run the CI builds. They seems to have gotten lost.

PR #407, opened 2017-04-08
[C++] Minor fixes.

ara-ayvazyan:


chwarr:
In the future, can include a few more details in the commit message. For example, this on is more than just "minor fixes" It's about forward declarations.

LGTM. Waiting on CI builds.

chwarr:
Merged in commit e844e5d5

PR #409, opened 2017-04-10
[c++] Refactoring to use constexpr in more places

ara-ayvazyan:


chwarr:
There's a conflict now with master (as of 3db58380a83de4184945d81074baa94b3e52028a) that needs to be resolved.

cpp/inc/bond/protocol/compact_binary.h
```diff
@@ -503,7 +503,7 @@ class CompactBinaryReader
 };
 
 template <typename Buffer>
-const uint16_t CompactBinaryReader<Buffer>::magic = COMPACT_PROTOCOL;
+BOND_CONSTEXPR_OR_CONST uint16_t CompactBinaryReader<Buffer>::magic;
```
chwarr: Minor: we're using `BufferT` above, but `Buffer` here. Consider `BufferT` here as well.

chwarr: Do we need storage for these integral constants? Similarly for the other protocols.

ara-ayvazyan: Yes, we do for _clang_ (I guess that is why it was commented out initially).

cpp/inc/bond/core/bond_version.h
```diff
@@ -16,12 +18,10 @@ namespace bond
     template <typename BufferT>
     class CompactBinaryReader;
 
-    static const uint16_t v1 = 0x0001;
-    static const uint16_t v2 = 0x0002;
+    BOND_STATIC_CONSTEXPR uint16_t v1 = 0x0001;
```
ara-ayvazyan: I believe global `const` variables have [internal linkage](http://en.cppreference.com/w/cpp/language/storage_duration) implicitly (which is needed here, since these are _defined_ in a header). The `static` here makes it more explicit. The `const`/`constexpr` should be enough. I'll try to change and see if I hit any compiler issues.

PR #410, opened 2017-04-11
[c++] Fixing broken custom buffer support

ara-ayvazyan:
Currently custom input and output buffer support is limited. Some scenarios like transcoding or the use of `bonded<T>` is broken. This change addresses those issues and provides full support of [custom streams](https://microsoft.github.io/bond/manual/bond_cpp.html#custom-streams).
- Fixes the places where built-in `InputBuffer` and `OutputBuffer` are used by relaxing the return type rules for existing `GetCurrentBuffer` function and introducing several new ones: `CreateInputBuffer`, `CreateOutputBuffer` and `GetBufferRange`.
- Fixes the `Protocols<Buffer>` type by removing its type dependency on a `Buffer` which allows it to contain protocols of different input buffers, and instead introduces `FilteredProtocols<Buffer>` helper type that will filter out the ones for a specific buffer type.


chwarr:
Don't forget to update the changelog, especially with a note about what other custom buffers will now need to support.

chwarr:
Can you resolve the conflicts in custom_protocols?

chwarr:
Merging now...

chwarr:
Actually, I never reviewed this. Reviewing now. :-)

cpp/inc/bond/core/protocol.h
```diff
@@ -148,27 +149,44 @@ struct ValueReader
...
+
+template <typename InputBuffer>
+struct FilteredProtocols
+{
+    template <typename Reader, typename Enable = void> struct
+    check_buffer
+        : std::false_type {};
+
+    template <typename Reader> struct
+    check_buffer<Reader, typename boost::enable_if<std::is_same<typename Reader::Buffer, InputBuffer> >::type>
```
chwarr: Can `check_buffer` be private, the minimize the size of the public API?

cpp/inc/bond/protocol/encoding.h
```diff
@@ -118,12 +118,15 @@ ReadVariableUnsigned(Buffer& input, T& value)
...
-inline void WriteRawBlob(Writer& writer, const blob& data)
-{
-    writer.Write(data);
+// Returns an object that represents a buffer range. The input arguments are
+// determined by what the GetCurrentBuffer returns for the given input buffer
+// implementation (i.e. not necessarily a blob). The GetBufferRange may return
+// a type different from blob as long as it is understood by the corresponding
+// output buffer associated with the writer (i.e. there is an overloaded Write
+// function that accepts it).
+inline blob GetBufferRange(const blob& begin, const blob& end)
```
chwarr: Should there be a generic definition/declaration for `GetBufferRange`, `CreateInputBuffer`, and `CreateOutputBuffer` to a) provide a good failure message when they are unimplemented and b) show the concept that each needs to implement?

ara-ayvazyan: I like the idea. Added bond/stream/stream_interface.h as discussed offline.

cpp/inc/bond/core/protocol.h
ara-ayvazyan: Yep, marked as private.

cpp/inc/bond/stream/stream_interface.h
```diff
@@ -14,7 +14,8 @@ namespace detail
 {
 
 template <typename T>
-using always_false = std::false_type;
+struct always_false
+    : std::false_type {};
```
chwarr: Cool. Let's wait for CI builds to check this fix.

PR #411, opened 2017-04-11
[c++] Use simplified uniform signature for bond::Apply

ara-ayvazyan:
The `bond::Apply` function has multiple overloads with [SFINAE](http://en.cppreference.com/w/cpp/language/sfinae) checks which are invoked from multiple places. This change moves them to `bond::detail::ApplyTransform` and replaces with only two simple ones without those checks. It reduces compilers' memory consumption and improves build time and also avoids some compiler crashes (primarily on MSVC).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ different versioning scheme, following the Haskell community's
     * Clang 3.4 or newer
     * GNU C++ 4.7 or newer
     * Microsoft Visual C++ 2013 or newer
+* The `bond::Apply` now has a uniform signature. Users who were calling
+  the overload for `Marshaler<Writer>` transform and were _mistakenly_
+  passing `Writer` explicitly (i.e. `bond::Apply<Writer>(marshaler, value)`)
+  they will get a compile error. To fix, please remove the `<Writer>` part.
 
 ## 5.3.0: 2017-04-12 ##
 * `gbc` & compiler library: 0.9.0.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,10 +26,11 @@ different versioning scheme, following the Haskell community's
     * Clang 3.4 or newer
     * GNU C++ 4.7 or newer
     * Microsoft Visual C++ 2013 or newer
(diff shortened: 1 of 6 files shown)


cpp/inc/bond/core/apply.h
```diff
@@ -44,34 +46,48 @@ Apply(const Transform& transform, const T& value)
...
+        return DoublePassApply(transform, value);
     else
         return StaticParser<const T&>(value).Apply(transform, typename schema<T>::type());
 }
 
+} // namespace detail
+
+
+template <typename Transform, typename T>
+bool inline Apply(const Transform& transform, T& value)
```
chwarr: Let's drop a note in the changelog to help people migrate if they accidentally had explicit types in their call to `bond::Apply<whoops>(...)`.

ara-ayvazyan: I think the only ones that won't be backward compatible are the overloads for `Marshaler<Writer>` where the first template argument is a `Writer` and not a `Transform`. Will put this in change log.

CHANGELOG.md
```diff
@@ -26,6 +26,10 @@ different versioning scheme, following the Haskell community's
     * Clang 3.4 or newer
     * GNU C++ 4.7 or newer
     * Microsoft Visual C++ 2013 or newer
+* The `bond::Apply` now has a uniform signature. Users who were calling
+  the overload for `Marshaler<Writer>` transform and were _mistakenly_
+  passing `Writer` explicitly (i.e. `bond::Apply<Writer>(marshaler, value)`)
+  they will get a compile error. To fix, please remove the `<Writer>` part.
```
chwarr: Here's a copy edited version:

The `bond::Apply` function now has a uniform signature. Call sites for 
the `Marshaler<Writer>` transform overload that were _mistakenly_
passing `Writer` explicitly (e.g. `bond::Apply<Writer>(marshaler, value)`)
will now get a compiler error. To fix, remove the `<Writer>` part:
`bond::Apply(marshaler, value)`.

chwarr:
Looks good to merge after CI builds pass.

PR #419, opened 2017-04-18
[gbc c++] Generate extern templates of bond::Apply instead of overloads

ara-ayvazyan:
Currently code-generated *_apply.h/.cpp files are achieving build time improvements by declaring additional overloads of `bond::Apply` function for predefined set of transforms and a given user-defined struct(s) which are then selected, thanks to [ADL](http://en.cppreference.com/w/cpp/language/adl). With C++11 the same can be achieved in a more idiomatic way by using [extern templates](http://en.cppreference.com/w/cpp/language/function_template).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,8 @@ different versioning scheme, following the Haskell community's
     * Microsoft Visual C++ 2013 or newer
 * **Breaking change** The generated apply.h/.cpp files now contain
   [extern templates](http://en.cppreference.com/w/cpp/language/function_template)
-  of `bond::Apply` instead of overload implementations.
+  of `bond::Apply` instead of overload implementations. Calls to bare `Apply`
+  or `TypeNamespace::Apply` must be changed to `bond::Apply`.
 * The `bond::Apply` function now has a uniform signature. Call sites for
 the `Marshaler<Writer>` transform overload that were _mistakenly_ passing
 `Writer` explicitly (e.g. `bond::Apply<Writer>(marshaler, value)`) will now
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ the `Marshaler<Writer>` transform overload that were _mistakenly_ passing
 `Writer` explicitly (e.g. `bond::Apply<Writer>(marshaler, value)`) will now
 get a compiler error. To fix, remove the `<Writer>` part:
 `bond::Apply(marshaler, value)`.
+* C++ codegen now generates [extern templates](http://en.cppreference.com/w/cpp/language/function_template) of `bond::Apply` instead of overloads.
 
 ## 5.3.0: 2017-04-12 ##
 * `gbc` & compiler library: 0.9.0.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ different versioning scheme, following the Haskell community's
 * C# NuGet version: TBD
 * C# Comm NuGet version: TBD
 
+### `gbc` and Bond compiler library ###
+
+* C++ codegen now generates [extern templates](http://en.cppreference.com/w/cpp/language/function_template) of `bond::Apply` instead of overloads.
+
 ### C++ ###
 
 * **Breaking change** A C++11 compiler is now required. The minimum
@@ -31,7 +35,8 @@ the `Marshaler<Writer>` transform overload that were _mistakenly_ passing
 `Writer` explicitly (e.g. `bond::Apply<Writer>(marshaler, value)`) will now
 get a compiler error. To fix, remove the `<Writer>` part:
 `bond::Apply(marshaler, value)`.
-* C++ codegen now generates [extern templates](http://en.cppreference.com/w/cpp/language/function_template) of `bond::Apply` instead of overloads.
+* **Breaking change** Generated *_apply.h/.cpp files now contain [extern templates](http://en.cppreference.com/w/cpp/language/function_template)
+  of `bond::Apply` instead of overload implementations.
 
 ## 5.3.0: 2017-04-12 ##
 * `gbc` & compiler library: 0.9.0.0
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ the `Marshaler<Writer>` transform overload that were _mistakenly_ passing
 `Writer` explicitly (e.g. `bond::Apply<Writer>(marshaler, value)`) will now
 get a compiler error. To fix, remove the `<Writer>` part:
 `bond::Apply(marshaler, value)`.
+* C++ codegen now generates [extern templates](http://en.cppreference.com/w/cpp/language/function_template) of `bond::Apply` instead of overloads.
```
chadwalters: Is this technically a breaking change since we are changing the namespace of the Apply functions?

compiler/src/Language/Bond/Codegen/Cpp/Apply_cpp.hs
```diff
@@ -21,15 +20,13 @@ apply_cpp protocols cpp file _imports declarations = ("_apply.cpp", [lt|
...
 |])
   where
-    body = [lt|
-    {
-        return ::bond::Apply<>(transform, value);
-    }|]
-
     attr = [lt||]
 
+    extern = [lt|template|]
```
chadwalters: Wouldn't it be more appropriate to have the literal 'template' appear directly in the applyOverloads function and have extern defined as 'extern = [lt||]' here and as 'extern = [lt|extern|}' here?

compiler/tests/generated/apply/basic_types_apply.h
```diff
@@ -6,208 +6,257 @@
 #include <bond/stream/output_buffer.h>
 
 
-namespace tests
+namespace bond
```
chadwalters: Since we are moving namespaces anyway, would it makes sense to move these inside bond::details?

cpp/inc/bond/core/apply.h
```diff
@@ -78,13 +78,13 @@ ApplyTransform(const Transform& transform, const T& value)
 
 
 template <typename Transform, typename T>
-bool inline Apply(const Transform& transform, T& value)
+bool Apply(const Transform& transform, T& value)
```
chadwalters: What is the impact of removing these inlines?

compiler/tests/generated/apply/basic_types_apply.h
ara-ayvazyan: The extern template must appear in the namespace where it is declared.

cpp/inc/bond/core/apply.h
ara-ayvazyan: The extern template declaration is not allowed for explicit inline functions: 

> Explicit instantiation of a function template or of a member function of a class template cannot use inline or constexpr.

(see _Expicit instantiation_ section in [here](http://en.cppreference.com/w/cpp/language/function_template), because it makes different requirements with regards to internal/external linkage. Note, that this does not mean the function will/may not be inlined in regular cases (in fact, I never use explicit `inline` with template functions because they are already implicitly declared that way, unless it is a full specialization in a header). It simply allows external linkage as well.

CHANGELOG.md
ara-ayvazyan: Even though we are removing function overloads from user namespace, which means a breaking change if they are calling with a fully qualified name (e.g. `MyNamespace::Apply(transform, value)`), I doubt anyone is doing that because it simply defeats the purpose of those functions. In fact, any call to `Apply` must be done with unqualified name in order the ADL to work. I wouldn't consider this as a breaking change, but let me know if you have a different opinion.

chadwalters: Even though it is unlikely to impact folks, we should mark it as breaking. Since we have other breaking changes pending, this won't be a blocker to getting this particular one released.

ara-ayvazyan: Added a separate note for c++ as breaking and gbc as non-breaking.

CHANGELOG.md
```diff
@@ -19,13 +19,22 @@ different versioning scheme, following the Haskell community's
...
+  of `bond::Apply` instead of overloads.
+
 ### C++ ###
 
 * **Breaking change** A C++11 compiler is now required. The minimum
   supported C++ compiler versions are now:
     * Clang 3.4 or newer
     * GNU C++ 4.7 or newer
     * Microsoft Visual C++ 2013 or newer
+* **Breaking change** The generated apply.h/.cpp files now contain
```
chwarr: For this note about the breaking change, we should describe what manual action needs to be taken to fix the break.

compiler/src/Language/Bond/Codegen/Cpp/Apply_cpp.hs
```diff
@@ -21,15 +20,13 @@ apply_cpp protocols cpp file _imports declarations = ("_apply.cpp", [lt|
...
 |])
   where
-    body = [lt|
-    {
-        return ::bond::Apply<>(transform, value);
-    }|]
-
     attr = [lt||]
 
+    extern = [lt||]
```
chwarr: Minor: no need to use quasi-quote syntax here. Just `""` is fine.

```
attr = ""
extern = ""
```

We only need to use the `[lt|...|]` when there are `#{expressions}` that we want to splice in.

compiler/src/Language/Bond/Codegen/Cpp/Apply_h.hs
```diff
@@ -30,15 +29,16 @@ apply_h protocols export_attribute cpp file imports declarations = ("_apply.h",
...
+} // namespace bond
 |])
   where
     includeImport (Import path) = [lt|#include "#{dropExtension path}_apply.h"|]
 
     export_attr = optional (\a -> [lt|#{a}
     |]) export_attribute
 
-    semi = [lt|;|]
+    extern = [lt|extern |]
```
chwarr: Minor: `"extern "` is enough here too.

chwarr:
I pushed a copy edit commit, so be sure to fetch before making more changes.

PR #422, opened 2017-04-20
[c++] Use /permissive- for Visual C++ 2017

chwarr:
Visual C++ 2017 has a new compiler switch to enable a more
standards-compliant mode. Since we build cross-compiler and
cross-platform, this flag can help us catch certain errors faster, so we
now enable it if the compiler supports it.


chwarr:
Closing and re-opening to re-run CI.

chwarr:
CI failure is MSVC 2013 timeout...

PR #424, opened 2017-04-21
[c++] Fixing a regression in pre-compiled Apply functions

ara-ayvazyan:
After introducing the two simpler overloads of `Apply` in #411 the case when passing a non-`const` reference to `bonded` selects an overload that takes non-`const T&` while corresponding `_apply.cpp` compiles it with `const` argument. This change fixes it by making sure that `Apply` accepting `T&` participates in overload resolution only for modifying transforms.

### DIFF
--- a/cpp/test/core/apply_tests.cpp
+++ b/cpp/test/core/apply_tests.cpp
@@ -1,6 +1,6 @@
 #include "precompiled.h"
 #include "apply_tests.h"
-#include "apply_test_reflection.h"
+#include "apply_test_apply.h"   // The _reflection.h should not be included.
 
 void Init(unittest::apply::Struct& obj)
 {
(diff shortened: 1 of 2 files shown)


cpp/test/core/apply_tests.cpp
```diff
@@ -1,6 +1,6 @@
 #include "precompiled.h"
 #include "apply_tests.h"
-#include "apply_test_reflection.h"
+#include "apply_test_apply.h"   // The _reflection.h should not be included.
```
chwarr: Minor: consider expanding the comment a bit indicating that we don't want to include _reflection so that we _only_ see the pre-generated apply overloads, instead of potentially expanding better ones based on what's in _reflection.h

chwarr:
LGTM. Thanks.

PR #425, opened 2017-04-23
[c++] Removing customize<protocols>

ara-ayvazyan:
As we see [here](https://microsoft.github.io/bond/manual/bond_cpp.html#understanding-bondedt) `bonded<T>` is using `ProtocolReader` in order to hold serialized data for any of the _built-in_ protocols. It maintains a global type list of those protocols which are also templated on input stream. In order to use a custom protocol, or existing protocol with custom input stream the `customize<protocols>` structure must be specialized where you can alter that predefined list (see [bf example](https://github.com/Microsoft/bond/blob/master/examples/cpp/core/bf/input_file.h#L135-L155)). This approach makes it less useful and a bit dangerous (anyone had [ODR](https://en.wikipedia.org/wiki/One_Definition_Rule) violation issues?) because the user must make sure that any static library they are using which also depends on Bond must have exactly same list of protocols, including possibly pre-compiled *_apply.cpp files. This becomes a problem, especially with large scale projects.

The proposed solution is to remove the global protocol list and ability to alter it via `customize<protocols>`, and let the user supply the actual list of desired protocols to the API-s that really need them (e.g. `bond::Marshal(obj, writer)` will use the default `bond::BuiltInProtocols` while `bond::Marshal<MyProtocols>(obj, writer)` will use the ones described by `MyProtocols`). The `ProtocolReader` will be holding serialized data in a structure similar to [`boost::any`](http://www.boost.org/doc/libs/1_64_0/doc/html/any.html) (`bond::detail::any`) with a small object optimization which will accommodate at least built-in protocols.


CHANGELOG.md
```diff
@@ -52,6 +52,10 @@ different versioning scheme, following the Haskell community's
       and
       [the bf example](https://github.com/Microsoft/bond/commit/11beaf5319639e4bdee96a25f95154e4fed93a75#diff-bdda0f39d99280d4858b4453906eea17)
       for more details.
+* **Breaking change** The `bond::customize<protocols>` is removed. All the
+  public API-s that require a protocol list (e.g. `bond::Marshal`) are now
+  accepting an extra template argument `Protocols` which is defaulted to
+  `bond::BuiltInProtocols`. For more details please see [the bf example](https://github.com/Microsoft/bond/commit/11beaf5319639e4bdee96a25f95154e4fed93a75#diff-bdda0f39d99280d4858b4453906eea17).
 * The `bond::Apply` function now has a uniform signature. Call sites for
```
chadwalters: "are now accepting" -> "now accept"
"is defaulted to" -> "defaults to"

I don't think the linked bf example diffs are correct or should even be applied until after the commit lands in master.

cpp/inc/bond/core/apply.h
```diff
@@ -16,26 +16,26 @@ namespace detail
 {
 
 /// @brief Apply transform to serialized struct wrapped in bonded<T>
-template <typename Transform, typename T, typename Reader>
+template <typename Protocols, typename Transform, typename T, typename Reader>
```
chadwalters: Any motivation for having Protocols as first template parameter rather than, say, last?

cpp/inc/bond/core/bond.h
```diff
@@ -13,116 +13,116 @@ namespace bond
```
chadwalters: Ordering of template parameters in this file seems oddly inconsistent to me.

cpp/inc/bond/core/detail/any.h
```diff
@@ -0,0 +1,297 @@
...
+
+
+namespace bond
+{
+namespace detail
+{
+
+template <template <typename> class TypeId, std::size_t Size>
+class any
+{
```
chadwalters: Provide some comment about why this implementation vs boost::any.

cpp/inc/bond/core/detail/mpl.h
```diff
@@ -0,0 +1,114 @@
+#pragma once
+
+#include <boost/static_assert.hpp>
+#include <type_traits>
+
+
```
chadwalters: Provide some comment about why this implementation vs boost::mpl.

cpp/inc/bond/core/apply.h
ara-ayvazyan: That is intentional, so that the user can explicitly pass a custom protocol list while arguments are still deduced implicitly (e.g. `bond::Apply<MyProtocols>(transform, value)`).

cpp/inc/bond/core/bond.h
ara-ayvazyan: Specifically which functions are you referring to? The order is usually determined based on the reasons I replied in above comment.

cpp/inc/bond/core/bonded.h
```diff
@@ -6,20 +6,23 @@
...
 #include "protocol.h"
 #include "runtime_schema.h"
-#include "detail/protocol_visitors.h"
+#include "bond_fwd.h"
 #include "detail/double_pass.h"
+#include "detail/protocol_visitors.h"
 #include "detail/marshaled_bonded.h"
 
+
+
```
chadwalters: Minor: excess spaces

ara-ayvazyan:
Given that the change is not short, here are some links to important parts in the code that may help to review:
- Important additions are [detail/mpl.h](https://github.com/Microsoft/bond/pull/425/files#diff-abe3cdc65b1fd3ac15ec1a844c642a33), [detail/any.h](https://github.com/Microsoft/bond/pull/425/files#diff-5a0c1cc4d6c5bff424d0023bae7b3dda) and [detail/visit_any.h](https://github.com/Microsoft/bond/pull/425/files#diff-f32d265e13cc17ac106ac4308f321127).
- Most of the important changes are in [protocol.h](https://github.com/Microsoft/bond/pull/425/files#diff-dc467f6158a80f542dc696525a88604d), [select_protocol.h](https://github.com/Microsoft/bond/pull/425/files#diff-3efbd97a6dddb2a5165056ec93d9033a), [detail/protocol_visitors.h](https://github.com/Microsoft/bond/pull/425/files#diff-88e08c8ad17e5bbd390a4581661c71e7) and [customize.h](https://github.com/Microsoft/bond/pull/425/files#diff-95fac4d0ea8015afc866555d11789442).
- The [bond_fwd.h](https://github.com/Microsoft/bond/pull/425/files#diff-73e6de22503cc8f351ff5e9984e2f282) has some more declarations.
- The [SimpleBinaryReader](https://github.com/Microsoft/bond/pull/425/files#diff-f96b70b8c7c899f190f192a5712149b6) has acquired a new `MarshaledBondedProtocolsT` type argument.
- The [`bonded<T>::Deserialize`](https://github.com/Microsoft/bond/pull/425/files#diff-2f54e46ff6f57497c9972a089e010606L171) combines two overloads in one implementation.
- The `detail::is_reader` is moved to [traits.h](https://github.com/Microsoft/bond/pull/425/files#diff-2e1711623c008d51a42cacee249681b3) and there is a new `buffer_magic`.
- The [bf/input_file.h](https://github.com/Microsoft/bond/pull/425/files#diff-bdda0f39d99280d4858b4453906eea17) is changed to use new APIs.
- The [unit_test_framework.h](https://github.com/Microsoft/bond/pull/425/files#diff-459ca020dcac6dfa0cbe348b741c6882) now uses variadic templates.
- The [test/core/equal.h](https://github.com/Microsoft/bond/pull/425/files#diff-80b583762e5c6a4201856a3a85025661) replaces the `operator==` with `Compare` function so that it is easier to call with extra `Protocols` type argument.
- All other changes are either adding the extra `Protocols` type argument or using it.

cpp/inc/bond/core/detail/any.h
```diff
@@ -0,0 +1,307 @@
...
+        template <typename T, bool IsSmall>
+        struct impl;
+
+        template <typename T>
+        struct impl<T, true>
+        {
+            static T* unsafe_cast(any& x) BOND_NOEXCEPT
+            {
+                BOOST_ASSERT(TypeId<T>::value == x._id);
+                return reinterpret_cast<T*>(x.data());
```
chwarr: Any reason we can't use `static_cast<T*>` here and below? `x.data()` returns a `void*`.

cpp/inc/bond/core/detail/any.h
```diff
@@ -0,0 +1,307 @@
...
+
+    void* data() BOND_NOEXCEPT
+    {
+        return &_storage;
+    }
+
+    template <typename T>
+    void emplace(const T& value)
+    {
+        functions::template table<T>::emplace(*this, value);
```
chwarr: Can we add comments saying that these `emplace` functions can only be called if this instance is effectively empty?

cpp/inc/bond/core/detail/visit_any.h
```diff
@@ -0,0 +1,125 @@
...
+template <typename T, typename Visitor, typename Any>
+typename boost::disable_if<std::is_void<std::result_of_t<Visitor(T&)> >, boost::optional<std::result_of_t<Visitor(T&)> > >::type
+try_visit_any(Visitor&& visitor, Any& x)
+{
+    if (auto value = any_cast<T>(&x))
+    {
+        return std::forward<Visitor>(visitor)(*value);
+    }
+
+    return {};
```
chwarr: `boost::none` may be clearer here than `{}`. Below as well for the newer implementation.

cpp/inc/bond/core/detail/mpl.h
```diff
@@ -0,0 +1,120 @@
...
+template <typename List, template <typename> class C>
+using filter_t = typename filter<List, C>::type;
+
+template <template <typename> class C> struct
+filter<list<>, C>
+    : identity<list<> > {};
+
+template <typename T, typename... U, template <typename> class C> struct
+filter<list<T, U...>, C>
+    : append< typename std::conditional<C<T>::value, list<T>, list<> >::type, filter_t<list<U...>, C> > {};
```
chwarr: Did you do any performance testing to compare this implementation, which appends a dummy list in the false case to code like this? (Syntax may not be correct, but in the false case, skip the append and just recur.)

```
template <typename T, typename... U, template <typename> class C> struct
filter<list<T, U...>, C>
    : std::conditional<C<T>::value, append<list<T>, filter_t<list<U...>, C> >>, filter_t<list<U...>, C> > >::type, {};
```

cpp/inc/bond/core/select_protocol.h
```diff
@@ -8,203 +8,224 @@
...
+                static_cast<ProtocolType>(Reader::magic),
+                Apply<Protocols>(transform, bonded<T, ProtocolReader>(reader)));
+        }
 
-    if (reader.ReadVersion())
-    {
-        return std::make_pair(
-            static_cast<ProtocolType>(Reader::magic), 
-            Apply(transform, bonded<T, ProtocolReader>(reader)));
+        return {};
```
ara-ayvazyan: Trying to avoid `boost::none` because it is going to change with `std::optional` (it uses `nullopt_t`) and usually preferring `{}` as it is more universal. But will change to `boost::none` if you still think it is cleaner.

cpp/test/core/list_tests.cpp
```diff
@@ -1,35 +1,33 @@
...
-        && left.s_uint64 == right.s_uint64
-        && left.s_string == right.s_string
-        && left.m_int8_string == right.m_int8_string
-        && left.m_string_bool == right.m_string_bool;
+    return Compare<Protocols>(left.l_bool, right.l_bool)
+        && Compare<Protocols>(left.l_int64, right.l_int64)
+        && Compare<Protocols>(left.l_float, right.l_float)
+        && Compare<Protocols>(left.l_string, right.l_string)
+        && Compare<Protocols>(left.v_int16, right.v_int16)
+        && Equal<Protocols>(left.v_double, right.v_double)
```
ara-ayvazyan: No, I actually changed this by hand, one by one, because it was a bit dangerous :) The `Compare` is a replacement for `operator=` and you can see that old code here used `Equal` for `v_double` only.

cpp/inc/bond/core/detail/mpl.h
ara-ayvazyan: Wasn't worried about perf here because this part is supposed to deal with only a handful of types (<10?), otherwise I actually did try some testing on this (as part of my other change to use variadics instead of mpl) and for relatively large number of types (~100+ or so) MSVC simply "breaks" :) and the best approach is to divide the set into 2 parts so that the recursion depth would become logN. I can change to use that, but I think here it will complicate with no benefits.

cpp/inc/bond/core/select_protocol.h
chwarr: Nope, you're reasoning makes sense. Carry on.

cpp/inc/bond/core/detail/visit_any.h
chwarr: Discussed above. `{}` to make it easier to switch to `std::optional`.

cpp/inc/bond/core/detail/mpl.h
chwarr: No, we don't need the complicated tree here.

cpp/inc/bond/core/detail/any.h
```diff
@@ -0,0 +1,307 @@
...
+    {
+        if (!empty())
+        {
+            _functions.destroy(*this);
+            _id = {};
+        }
+    }
+
+
+    std::uint32_t _id;
```
chwarr: `_storage` then `_functions` then `_id` to minimize slack space added for alignment?

cpp/inc/bond/core/traits.h
```diff
@@ -168,6 +175,40 @@ template <typename T> struct
...
     : is_object<typename aliased_type<T>::type> {};
 
+
+// is_reader
+template <typename Input, typename T = void, typename Enable = void> struct
+is_reader
+    : false_type {};
+
+template <typename Input, typename T> struct
+is_reader<Input&, T>
```
ara-ayvazyan: You cannot use any such traits in partial specializations, because compiler won't be able to deduce the type (e.g. `is_reader<typename remove_reference<T>::type, T>`). So either (1) you specialize for individual `T` which is not a reference or `const/volatile` and consumers are explicitly using `remove_reference/remove_cv`, or (2) we add such partial specializations for `T&`, `const/volatile T` which falls back to `T` like that is done above so that users do not have to worry about that (assuming this trat should not behave differently for `T` vs `T&` etc.).

cpp/inc/bond/core/select_protocol.h
```diff
@@ -8,203 +8,224 @@
...
-    const Iter&,
-    const EndIter& end,
-    Buffer& input,
-    const Transform& transform)
+template <typename T, typename Protocols>
+struct NextProtocolFunctor
 {
-    typedef typename boost::mpl::deref<Iter>::type Reader;
+    template <typename Buffer, typename Transform, typename Reader>
+    inline boost::optional<std::pair<ProtocolType, bool> >
```
ara-ayvazyan: No, it is not required. I somehow followed the existing _style_ here.

cpp/inc/bond/core/detail/any.h
ara-ayvazyan: Should I instead rename it to something like `emplace_uninit`?

chadwalters: Sure, I'd go with `emplace_uninitialized` but also have a comment to that effect.

cpp/test/core/equal.h
```diff
@@ -35,110 +103,120 @@ class Compare
...
 
-inline bool Equal(double left, double right)
+template <typename Protocols = bond::BuiltInProtocols, typename T1, typename T2>
+bool Compare(const T1& lhs, const T2& rhs)
+{
+    return lhs == rhs;
+}
+
+
+template <typename Protocols = bond::BuiltInProtocols, typename T>
```
ara-ayvazyan: From what I understand, the `Compare` (former `operator==`) is doing a simple value to value comparison, with some custom logic in some structs, while `Equal` compares the structures being tested in a way test is concerned (e.g. deserialize a `bonded<T>` instead of calling `bonded<T>::operator==`).

chwarr:
AppVeyor failed due to a network communication error downloading the Haskell compiler. I'm attempting to [re-run the build](https://ci.appveyor.com/project/MicrosoftBond/bond/build/1.0.196).

PR #440, opened 2017-05-03
Clarify string encoding in binary protocol specs

chwarr:
* Use Unicode jargon when it helps.
* Clarify that strings are UTF-8 and that wstrings are UTF16-LE, if
  you're strictly following the spec.
* Copy simple binary spec from the .cs file to the .h file.


chwarr:
For C++, I am intentionally avoiding talking about the fact that the C++ library doesn't validate that the correct encoding is being used.

cpp/inc/bond/protocol/simple_binary.h
```diff
@@ -9,6 +9,59 @@
...
+                           count            variable encoded uint32 count of 1-byte (for
+                                            string) or 2-byte (for wstring) Unicode code
+                                            units (variable encoded in v2)
+
+                           characters       1-byte UTF-8 code units (for string) or 2-byte
+                                            UTF-16LE code units (for wstring)
+
+
+                                           .-------. .-------.
+                    blob, list, set,       | count | | items |...
```
lalo: nit: should be item? followed by item item. Right?

PR #450, opened 2017-05-17
Remove documentation publish from AppVeyor build

chwarr:
The current documentation build process can expose the value of
BOND_TOKEN to authors of pull requests. We will need to find a different
way to build and publish the Bond documentation.

Until then, BOND_TOKEN used has been invalidated, and the publishing
step has been removed from the AppVeyor build.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,8 +18,6 @@
     clone_depth: 15
 
     environment:
-        BOND_TOKEN:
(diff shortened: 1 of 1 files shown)


chwarr:
AppVeyor's secure environment variables [are not exposed to pull requests](http://help.appveyor.com/discussions/questions/2615-secure-variables-for-pull-request). This change is not needed, but we still need to roll the token, as it was exposed--but never used--via other testing.

PR #452, opened 2017-05-24
java core: Implement placeholder bonded.

tstein:


compiler/src/Language/Bond/Codegen/Java/Util.hs
```diff
@@ -50,11 +50,7 @@ defaultValue :: MappingContext -> Field -> Maybe Text
 defaultValue java Field {fieldDefault = Nothing, ..} = implicitDefault fieldType
   where
     newInstance t = Just [lt|new #{getInstanceTypeName java t}()|]
-    implicitDefault (BT_Bonded t) = Just [lt|new Bonded<#{getTypeName java t}>()|]
-    -- We can't really initialize a default object for generic type
-    -- field due to the limitation of both Java and Kotlin.
-    -- I'm still looking for a solution. Please let me know if you has
-    -- any suggestions.
+    implicitDefault (BT_Bonded t) = Just [lt|new Bonded<#{getTypeName java t}>(null)|]
```
chwarr: It seems like the implicit default should be a `Bonded` wrapper around a `new T` instead of around `null`. (The default value for a struct is not null. It's a default-initialized struct.)

compiler/src/Language/Bond/Codegen/Java/SerializationMethods.hs
```diff
@@ -278,3 +280,13 @@ readMap java keyType valueType fieldName depth =
         countLocal = "count" ++ show depth
         keyLocal = "k" ++ show depth
         valueLocal = "v" ++ show depth
+
+readBonded :: MappingContext -> Type -> String -> Text
+readBonded java bondedType fieldName =
```
chwarr: `bondedType` matches any type, not just `BT_Bonded`. We've got a patten that we often use in cases like this:

```
readBonded java bondedType@(BT_Bonded _) fieldName = ...
readBonded _ _ _ = error "Impossible..."
```

jvm/src/main/java/com/microsoft/bond/IBonded.java
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package com.microsoft.bond;
+
+import com.microsoft.bond.protocol.ProtocolWriter;
+
+import java.io.IOException;
+
+public interface IBonded<T extends BondSerializable> {
+    <W extends ProtocolWriter> void Serialize(W writer) throws IOException;
```
sigansky: ><W extends ProtocolWriter> [](start = 4, length = 26)

There is no need for generic protocol writer, as we discussed yesterday.

compiler/src/Language/Bond/Codegen/Java/Util.hs
tstein: That was intentional for the placeholder, but you have encouraged me to Do The Right Thing(tm). 

The fix for that created some weirdness that led me to a small refactor of `defaultValue`. The C# `defaultValue` returns `Maybe Text` to omit an initializer when the Bond default matches the C# implicit default, but the Java one always returns an explicit default. I dropped the `Maybe`.

compiler/src/Language/Bond/Codegen/Java/SerializationMethods.hs
tstein: `bondedType` refers to the type inside the bonded. This is used as `BT_Bonded t -> readBonded java t varName`.

If you found the name misleading, I'm happy to change it.

jvm/src/main/java/com/microsoft/bond/IBonded.java
tstein: This is based on the structure of the C# version. I have a sneaking feeling it'll be back, but changed for now.

compiler/src/Language/Bond/Codegen/Java/SerializationMethods.hs
chwarr: If I'm reading the implementation, `readBonded` is a partial function over `Type` that only works when the `Type` is `BT_Bonded`. However, `readBonded` is written as if it were total.

chwarr: `innerType` instead of `bondedType`, perhaps?

chwarr: Got it. `innerType`, perhaps?

PR #453, opened 2017-05-25
Java: Update code generator to emit raw integer constants for enum values

sigansky:


compiler/src/Language/Bond/Codegen/Java/Enum_java.hs
```diff
@@ -30,37 +30,73 @@ package #{javaPackage};
...
-    }
+    public boolean equals(Object other) { return (other instanceof #{declName}) && (this.value == ((#{declName}) other).value); }
+ 
+    @Override
+    public int hashCode() { return this.value; }
 
     @Override
-    public int hashCode() {
-        return this.value;
+    public String toString() { return String.valueOf(this.value); }
```
tstein: I would expect `Color.RED.toString()` to return `"RED"`, not `0`. This is what true Java enums do.

You'll need a new member for this, but I think it's in scope for this change.

compiler/src/Language/Bond/Codegen/Java/Enum_java.hs
```diff
@@ -30,37 +30,73 @@ package #{javaPackage};
...
+        -- constant object
+        constantObjectDecl c@Constant {..} =
+            [lt|public static final #{declName} #{constantName} = #{enumObjectAssigmentValue c};|]
 
-        -- Process constants to make sure all values can be converted
-        -- to integer.
-        enumConstantsWithInt = fixEnumWithInt 0 enumConstants []
+        -- constant int
+        constantIntValueDecl Constant {..} = let value x = [lt|#{x}|] in
+            [lt|public static final int #{constantName} = #{optional value constantValue};|]
```
tstein: What happens if `constantValue` is `Nothing`?

compiler/src/Language/Bond/Codegen/Java/Enum_java.hs
```diff
@@ -30,37 +30,73 @@ package #{javaPackage};
...
+            [lt|case Values.#{constantName}: return #{constantName};|]
 
+        -- Process constants to make sure every constant value is set (either explicit or auto-generated)
+        enumConstantsWithInt = fixEnumWithInt 0 enumConstants []
+ 
         fixEnumWithInt :: Int -> [Constant] -> [Constant] -> [Constant]
         fixEnumWithInt _ [] result = reverse result
         fixEnumWithInt nextInt (h:r) result = case constantValue h of
           Just n -> fixEnumWithInt (n + 1) r (h:result)
           _ -> fixEnumWithInt (nextInt + 1) r ((Constant (constantName h) (Just nextInt)):result)
```
tstein: While you're in here, could you change this to an explicit `Nothing`?

compiler/src/Language/Bond/Codegen/Java/Enum_java.hs
tstein: I see - we rewrite them down there so they all have a value. If my understanding is correct, this is fine, though I'd be more jazzed if we errored out right here if a `Nothing` ever made it through.

compiler/src/Language/Bond/Codegen/Java/Enum_java.hs
```diff
@@ -30,37 +30,73 @@ package #{javaPackage};
...
+    public int hashCode() { return this.value; }
 
     @Override
-    public int hashCode() {
-        return this.value;
+    public String toString() { return String.valueOf(this.value); }
+
+    public static #{declName} get(int value) {
+        switch (value) {
+            #{newlineSep 3 switchCaseConstantMapping enumConstantsWithIntDistinct}
```
tstein: The use of the `Distinct` flavor is to avoid two case labels for ints with the same value, right? Is it a compile error to have two equivalent labels? My gut says to emit both labels for human clarity, let the compiler ignore the redundant ones, and let the `@generated` annotation suppress the warning I would think this causes.

compiler/src/Language/Bond/Codegen/Java/Enum_java.hs
```diff
@@ -30,37 +30,73 @@ package #{javaPackage};
...
+            then findEnumConstantsDistinct r keys
+            else h : findEnumConstantsDistinct r ((constantValue h):keys)
+
+        -- The RHS value to assign to an enum object
+        enumObjectAssigmentValue :: Constant -> Text
+        enumObjectAssigmentValue enumConstant =
+            let firstDeclaredEnumConstant = head (filter (\c -> (constantValue c) == (constantValue enumConstant)) enumConstantsWithIntDistinct) in
+                if firstDeclaredEnumConstant == enumConstant
+                    then [lt|new #{declName}(Values.#{constantName enumConstant})|]
+                    else[lt|#{constantName firstDeclaredEnumConstant}|]
```
tstein: ~ s/else[lt/else [lt

compiler/src/Language/Bond/Codegen/Java/Enum_java.hs
```diff
@@ -30,37 +30,73 @@ package #{javaPackage};
...
+        -- Process constants to make sure every constant value is distinct
+        enumConstantsWithIntDistinct = findEnumConstantsDistinct enumConstantsWithInt []
+
+        findEnumConstantsDistinct :: [Constant] -> [Maybe Int] -> [Constant]
+        findEnumConstantsDistinct [] _ = []
+        findEnumConstantsDistinct (h:r) keys = if elem (constantValue h) keys
+            then findEnumConstantsDistinct r keys
+            else h : findEnumConstantsDistinct r ((constantValue h):keys)
+
+        -- The RHS value to assign to an enum object
```
tstein: I would like a beefier comment here, too. How about this?

```The RHS value to assign to an enum object. If an enum element is the first instance with a particular value, this function will return an instantiation. If it isn't, this function will return a reference to the first enum object with the same value.```

File: compiler/src/Language/Bond/Codegen/Java/Enum_java.hs [hunk 7]
```diff
@@ -30,37 +30,73 @@ package #{javaPackage};
...
+        -- Process constants to make sure every constant value is set (either explicit or auto-generated)
+        enumConstantsWithInt = fixEnumWithInt 0 enumConstants []
+ 
         fixEnumWithInt :: Int -> [Constant] -> [Constant] -> [Constant]
         fixEnumWithInt _ [] result = reverse result
         fixEnumWithInt nextInt (h:r) result = case constantValue h of
           Just n -> fixEnumWithInt (n + 1) r (h:result)
           _ -> fixEnumWithInt (nextInt + 1) r ((Constant (constantName h) (Just nextInt)):result)
 
+        -- Process constants to make sure every constant value is distinct
```
tstein: I would like a beefier comment here. How about this?

```Filter a list of constants, leaving a list of constants with distinct values. If several constants in the input list share a value, the first one that appears will be the one that appears in the output list.```
?

File: compiler/src/Language/Bond/Codegen/Java/Enum_java.hs [hunk 8]
```diff
@@ -30,37 +30,73 @@ package #{javaPackage};
...
+        findEnumConstantsDistinct :: [Constant] -> [Maybe Int] -> [Constant]
+        findEnumConstantsDistinct [] _ = []
+        findEnumConstantsDistinct (h:r) keys = if elem (constantValue h) keys
+            then findEnumConstantsDistinct r keys
+            else h : findEnumConstantsDistinct r ((constantValue h):keys)
+
+        -- The RHS value to assign to an enum object
+        enumObjectAssigmentValue :: Constant -> Text
+        enumObjectAssigmentValue enumConstant =
+            let firstDeclaredEnumConstant = head (filter (\c -> (constantValue c) == (constantValue enumConstant)) enumConstantsWithIntDistinct) in
```
tstein: This might be better expressed with [`any`](http://hackage.haskell.org/packages/archive/base/latest/doc/html/Prelude.html#v:any).

compiler/src/Language/Bond/Codegen/Java/Enum_java.hs
sigansky: Returning the integer applies uniformly to all values whereas returning a literal doesn't. Specifically we need to consider what to do for aliases (multiple constants mapping to the same value) and unknown values. For the latter the int value is the only reasonable choice.

compiler/src/Language/Bond/Codegen/Java/Enum_java.hs
sigansky: We emit both but they all reference the same object. The intent here is to declare only the first occurrence and let others reference it.

compiler/src/Language/Bond/Codegen/Java/Enum_java.hs
sigansky: Any evaluates to Bool, so I am not sure what you mean. Can you please clarify?

compiler/src/Language/Bond/Codegen/Java/Enum_java.hs
sigansky: Not sure I understand this.

compiler/src/Language/Bond/Codegen/Java/Enum_java.hs
sigansky: Not sure about this, Can you clarify?

compiler/src/Language/Bond/Codegen/Java/Enum_java.hs
tstein: Was thinking about this as a general switch - my bad. This is good as-is.

compiler/src/Language/Bond/Codegen/Java/Enum_java.hs
tstein: Got a bit tripped up - this suggestion doesn't actually make sense. With the comment I suggested, or something similar, I think this is fine.

PR #454, opened 2017-05-25
Address issue with Mono .mdb -> .pdb change

chadwalters:
Fixes issues with Mono build on Travis.


chadwalters:
See Travis run here: https://travis-ci.org/chadwalters/bond/builds/235906891

PR #456, opened 2017-05-25
[c++ grpc] Implement events and generic services

lalo:
- codegen enabled for events and void types
- codegen for generic services

### DIFF
--- a/compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
+++ b/compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
@@ -160,8 +160,8 @@ inline #{className}::#{proxyName}<TThreadPool>::#{proxyName}(
         serviceMethodsWithIndex :: [(Integer,Method)]
         serviceMethodsWithIndex = zip [0..] serviceMethods
 
-        publicProxyMethodDecl Function{methodInput = Nothing, ..} = [lt|void Async#{methodName}(::grpc::ClientContext* context, std::function<void(const #{response methodResult}&, const ::grpc::Status&)> cb);|]
-        publicProxyMethodDecl Function{..} = [lt|void Async#{methodName}(::grpc::ClientContext* context, const #{request methodInput}& request, std::function<void(const #{response methodResult}&, const ::grpc::Status&)> cb);|]
+        publicProxyMethodDecl Function{methodInput = Nothing, ..} = [lt|void Async#{methodName}(::grpc::ClientContext* context, const std::function<void(const #{response methodResult}&, const ::grpc::Status&)>& cb);|]
+        publicProxyMethodDecl Function{..} = [lt|void Async#{methodName}(::grpc::ClientContext* context, const #{request methodInput}& request, const std::function<void(const #{response methodResult}&, const ::grpc::Status&)>& cb);|]
         publicProxyMethodDecl Event{methodInput = Nothing, ..} = [lt|void Async#{methodName}(::grpc::ClientContext* context);|]
         publicProxyMethodDecl Event{..} = [lt|void Async#{methodName}(::grpc::ClientContext* context, const #{request methodInput}& request);|]
 
@@ -175,7 +175,7 @@ inline #{className}::#{proxyName}<TThreadPool>::#{proxyName}(
 inline void #{className}::#{proxyName}<TThreadPool>::Async#{methodName}(
     ::grpc::ClientContext* context,
     #{voidParam methodInput}
-    std::function<void(const #{response methodResult}&, const ::grpc::Status&)> cb)
+    const std::function<void(const #{response methodResult}&, const ::grpc::Status&)>& cb)
 {
     #{voidRequest methodInput}
     auto calldata = new ::bond::ext::gRPC::detail::client_unary_call_data< #{request methodInput}, #{response methodResult}, TThreadPool >(
@@ -193,7 +193,7 @@ inline void #{className}::#{proxyName}<TThreadPool>::Async#{methodName}(
 
         methodDecl Event{..} = [lt|#{template}template <typename TThreadPool>
 inline void #{className}::#{proxyName}<TThreadPool>::Async#{methodName}(
-    ::grpc::ClientContext* context,
+    ::grpc::ClientContext* context
     #{voidParam methodInput})
 {
     #{voidRequest methodInput}
@@ -207,7 +207,7 @@ inline void #{className}::#{proxyName}<TThreadPool>::Async#{methodName}(
             voidRequest Nothing = [lt|auto request = ::bond::bonded< ::bond::Void>{ ::bond::Void()};|]
             voidRequest _ = mempty
             voidParam Nothing = mempty
-            voidParam _ = [lt|const #{request methodInput}& request|]
+            voidParam _ = [lt|, const #{request methodInput}& request|]
 
         serviceAddMethod Function{..} = [lt|this->AddMethod("/#{getDeclTypeName idl s}/#{methodName}");|]
         serviceAddMethod Event{..} = [lt|this->AddMethod("/#{getDeclTypeName idl s}/#{methodName}");|]
--- a/compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
+++ b/compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
@@ -69,7 +69,7 @@ grpc_h _ cpp file imports declarations = ("_grpc.h", [lt|
 
     request mt = request' (payload mt)
       where
-        payload = maybe "void" cppType
+        payload = maybe "::bond::Void" cppType
         request' params =  [lt|::bond::bonded<#{padLeft}#{params}>|]
           where
             paramsText = toLazyText params
@@ -77,14 +77,14 @@ grpc_h _ cpp file imports declarations = ("_grpc.h", [lt|
 
     response mt = response' (payload mt)
       where
-        payload = maybe "void" cppType
+        payload = maybe "::bond::Void" cppType
         response' params =  [lt|::bond::bonded<#{padLeft}#{params}>|]
           where
             paramsText = toLazyText params
             padLeft = if L.head paramsText == ':' then [lt| |] else mempty
 
     grpc s@Service{..} = [lt|
-class #{declName} final
+#{template}class #{declName} final
 {
 public:
     template <typename TThreadPool>
@@ -135,8 +135,8 @@ public:
     using Service = #{serviceName}< ::bond::ext::gRPC::thread_pool>;
 };
 
-template <typename TThreadPool>
-inline #{declName}::#{proxyName}<TThreadPool>::#{proxyName}(
+#{template}template <typename TThreadPool>
+inline #{className}::#{proxyName}<TThreadPool>::#{proxyName}(
     const std::shared_ptr< ::grpc::ChannelInterface>& channel,
     std::shared_ptr< ::bond::ext::gRPC::io_manager> ioManager,
     std::shared_ptr<TThreadPool> threadPool)
@@ -149,6 +149,8 @@ inline #{declName}::#{proxyName}<TThreadPool>::#{proxyName}(
 #{doubleLineSep 0 methodDecl serviceMethods}
 |]
       where
+        className = CPP.className s
+        template = CPP.template s
         proxyName = "ClientCore" :: String
         serviceName = "ServiceCore" :: String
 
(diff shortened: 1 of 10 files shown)


compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
```diff
@@ -158,29 +160,54 @@ inline #{declName}::#{proxyName}<TThreadPool>::#{proxyName}(
...
+        proxyMethodMemberInit Event{..} = [lt|, rpcmethod_#{methodName}_("/#{getDeclTypeName idl s}/#{methodName}", ::grpc::RpcMethod::NORMAL_RPC, channel)|]
 
-        methodDecl Function{..} = [lt|template <typename TThreadPool>
-inline void #{declName}::#{proxyName}<TThreadPool>::Async#{methodName}(
+        methodDecl Function{..} = [lt|#{template}template <typename TThreadPool>
+inline void #{className}::#{proxyName}<TThreadPool>::Async#{methodName}(
     ::grpc::ClientContext* context,
-    const #{request methodInput}& request,
+    #{voidParam methodInput}
     std::function<void(const #{response methodResult}&, const ::grpc::Status&)> cb)
```
ara-ayvazyan: Pass as `const&`?

compiler/tests/generated/generic_service_grpc.h
```diff
@@ -35,7 +35,8 @@
 namespace tests
 {
 
-class Foo final
+template <typename Payload>
+    class Foo final
 {
 public:
     template <typename TThreadPool>
```
ara-ayvazyan: I see that here we anyway deal with a template. Wondering why are the below callbacks use specifically `std::function` and not an arbitrary function object (template)?

compiler/tests/generated/service_grpc.h
```diff
@@ -178,12 +178,42 @@ class Foo final
...
+                0,
+                cq0,
+                tp,
+                std::bind(&ServiceCore::foo11, this, std::placeholders::_1));
+            _rd_foo12.emplace(
+                this,
+                1,
+                cq0,
+                tp,
+                std::bind(&ServiceCore::foo12, this, std::placeholders::_1));
```
ara-ayvazyan: Any reason for preferring `std::bind` over lambdas?

examples/cpp/grpc/pingpong/pingpong.cpp
```diff
@@ -0,0 +1,273 @@
...
+    ClientContext pingVoidContext;
+    ClientContext pingShouldThrowContext;
+    ClientContext pingGenericContext;
+
+    PingRequest request;
+    request.name = user;
+    bond::bonded<PingRequest> req(request);
+
+    {
+        std::function<void(const bond::bonded<PingReply>&, const Status&)> f_print =
```
ara-ayvazyan: Are do we need explicit `std::function` here?

lalo: no, `auto f_print` would suffice

ara-ayvazyan: Actually why do we need a variable here? :)

examples/cpp/grpc/pingpong/pingpong.cpp
```diff
@@ -0,0 +1,273 @@
...
+    bool pingVoidIsCorrectResponse;
+    bool pingShouldThrowIsCorrectResponse;
+    bool pingGenericIsCorrectResponse;
+
+    ClientContext context;
+    ClientContext pingNoPayloadContext;
+    ClientContext pingNoResponseContext;
+    ClientContext pingVoidContext;
+    ClientContext pingShouldThrowContext;
+    ClientContext pingGenericContext;
```
ara-ayvazyan: I feel like this example shows too many options at the same time?

compiler/tests/generated/generic_service_grpc.h
lalo: Well, this service in particular is a generic service (https://github.com/lalo/bond/blob/7329aaa47da678289122d13cfa1e3811ea048515/examples/cpp/grpc/pingpong/pingpong.bond#L26) so that is why we have a template here. 

I stuck with std::function from the beginning, I think we could do an arbitrary object since our thread_pool supports it (https://github.com/Microsoft/bond/blob/grpc%2B%2B/cpp/inc/bond/ext/grpc/thread_pool.h#L91)

examples/cpp/grpc/pingpong/pingpong.cpp
lalo: you are correct, this example is not much of an example and more of some unit tests so maybe I should move them over to the test dir.

compiler/tests/generated/generic_service_grpc.h
ara-ayvazyan: Assuming we do not want to waste any CPU cycles, then function object should be better, unless it adds significant complexity that you want to avoid. The `std::function` comes with a cost.

examples/cpp/grpc/pingpong/pingpong.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+    DoublePingServiceImpl double_ping_service;
+    PingPongServiceImpl ping_pong_service;
+
+    bond::ext::gRPC::server_builder builder;
+    builder.SetThreadPool(threadPool);
+    const std::string server_address("127.0.0.1:50051");
+    builder.AddListeningPort(server_address, grpc::InsecureServerCredentials());
+    builder.RegisterService(&double_ping_service);
+    builder.RegisterService(&ping_pong_service);
```
lalo: was not aware of this. thanks

compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
lalo: fixed

compiler/tests/generated/service_grpc.h
lalo: @chwarr may have some more background on this decision

cpp/inc/bond/ext/grpc/detail/client_call_data.h
```diff
@@ -57,7 +57,7 @@ struct client_unary_call_data : io_manager_tag
         std::shared_ptr<grpc::ChannelInterface> channel,
         std::shared_ptr<io_manager> ioManager,
         std::shared_ptr<TThreadPool> threadPool,
-        CallbackType cb)
+        CallbackType cb = CallbackType())
```
ara-ayvazyan: Simply do `= {}` unless you are hitting compiler bugs somewhere.

examples/cpp/grpc/pingpong/pingpong.cpp
lalo: we don't need it, but I liked how it read better separated. I will replace it to auto though

cpp/inc/bond/ext/grpc/detail/client_call_data.h
lalo: what is the name for this notation?

ara-ayvazyan: I think it is [aggregate_initialization](http://en.cppreference.com/w/cpp/language/aggregate_initialization).

compiler/tests/generated/service_grpc.h
```diff
@@ -452,25 +518,92 @@ inline Foo::ClientCore<TThreadPool>::ClientCore(
...
+        _ioManager,
+        _threadPool);
+    calldata->dispatch(rpcmethod_foo11_, context, request);
+}
 
-/* TODO: stub implementation for event foo12 */
+template <typename TThreadPool>
+inline void Foo::ClientCore<TThreadPool>::Asyncfoo12(
+    ::grpc::ClientContext* context,
+    )
```
ara-ayvazyan: Does `,` compile fine? :)

lalo: you are right, good catch

compiler/tests/generated/generic_service_grpc.h
lalo: I'll consider this change but will not address it in this PR.

compiler/tests/generated/service_grpc.h
chwarr: `std::function` was the easiest way _for me_ to get everything to work. Refactoring everything that today uses a `std::function` to take callable object hasn't happened yet. Feel free to delay as many of those refactorings as you need to to a separate commit and PR.

compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
```diff
@@ -135,8 +135,8 @@ public:
     using Service = #{serviceName}< ::bond::ext::gRPC::thread_pool>;
 };
 
-template <typename TThreadPool>
-inline #{declName}::#{proxyName}<TThreadPool>::#{proxyName}(
+#{template}template <typename TThreadPool>
```
chwarr: Minor `CPP.template` is causing odd indentation.

examples/cpp/grpc/pingpong/pingpong.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+// countdown_event.h and event.h needed for test purposes
+#include <bond/ext/detail/countdown_event.h>
+#include <bond/ext/detail/event.h>
+
+#include <bond/ext/grpc/io_manager.h>
+#include <bond/ext/grpc/server.h>
+#include <bond/ext/grpc/server_builder.h>
+#include <bond/ext/grpc/unary_call.h>
+#include <bond/ext/grpc/thread_pool.h>
```
chwarr: Minor: alpha sort

examples/cpp/grpc/pingpong/pingpong.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+using namespace pingpong;
+
+event pingNoResponse_event;
+
+class DoublePingServiceImpl final : public DoublePing::Service
+{
+    void Ping(
+        bond::ext::gRPC::unary_call<
+        bond::bonded<PingRequest>,
```
chwarr: Minor: let's indent the parameters to `unary_call` one level.

examples/cpp/grpc/pingpong/pingpong.cpp
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    void PingNoResponse(
+        bond::ext::gRPC::unary_call<
+        bond::bonded<PingRequest>,
+        bond::bonded<bond::Void>> call) override
+    {
+        PingRequest request = call.request().Deserialize();
+
+        call.Finish(bond::bonded<bond::Void>{bond::Void()}, Status::OK);
```
chwarr: We should add a comment like this here:
```
// TODO: the current implementation requires that we respond with dummy data. This will be fixed in a later release.
```

examples/cpp/grpc/pingpong/pingpong.cpp
```diff
@@ -0,0 +1,276 @@
...
+    {
+        call.Finish(bond::bonded<bond::Void>{bond::Void()}, Status::OK);
+    }
+
+    void PingEventVoid(
+        bond::ext::gRPC::unary_call<
+        bond::bonded<bond::Void>,
+        bond::bonded<bond::Void>> call) override
+    {
+        call.Finish(bond::bonded<bond::Void>{bond::Void()}, Status::OK);
```
chwarr: Similar comment for `PingNoResponse` here.

examples/cpp/grpc/pingpong/pingpong.cpp
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    void PingShouldThrow(
```
chwarr: "Throw" is not the right word to use. "ReturnsError"?

examples/cpp/grpc/pingpong/pingpong.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+    DoublePingServiceImpl double_ping_service;
+    PingPongServiceImpl ping_pong_service;
+
+    bond::ext::gRPC::server_builder builder;
+    builder.SetThreadPool(threadPool);
+    const std::string server_address("127.0.0.1:50051");
+    builder.AddListeningPort(server_address, grpc::InsecureServerCredentials());
+    builder.RegisterService(&double_ping_service)
+        .RegisterService(&ping_pong_service);
```
chwarr: Probably clearer to repeat `builder` on line 154 or to write this like

```
builder
    .RegisterService(...)
    .RegisterService(...);
```

examples/cpp/grpc/pingpong/pingpong.cpp
```diff
@@ -0,0 +1,276 @@
...
+                    pingVoidIsCorrectResponse = true;
+                    ping_event.set();
+                }
+                else
+                {
+                    pingVoidIsCorrectResponse = false;
+                }
+            };
+
+        doublePing.AsyncPingVoid(&pingVoidContext, f_print);
```
chwarr: Is it intentional that this invocation isn't in its own scope block?

examples/cpp/grpc/pingpong/pingpong.cpp
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    {
+        auto f_print = [&ping_event, &pingShouldThrowIsCorrectResponse](bond::bonded<PingReply> response, Status status)
+            {
+                pingShouldThrowIsCorrectResponse = false;
+
+                if(!status.ok())
```
chwarr: We can skip the `.ok()` check and just compare the `error_code()`. (`.ok()` is a helper for `error_code() == StatusCode::OK`)

examples/cpp/grpc/pingpong/pingpong.cpp
```diff
@@ -0,0 +1,276 @@
...
+                {
+                    if(status.error_code() == StatusCode::CANCELLED)
+                    {
+                        pingShouldThrowIsCorrectResponse = true;
+                        ping_event.set();
```
chwarr: Should this be set even if we don't get the right result?

examples/cpp/grpc/pingpong/pingpong.cpp
```diff
@@ -0,0 +1,276 @@
...
+    ClientContext pingVoidContext;
+    ClientContext pingShouldThrowContext;
+    ClientContext pingGenericContext;
+
+    PingRequest request;
+    request.name = user;
+    bond::bonded<PingRequest> req(request);
+
+    {
+        auto f_print = [&ping_event, &isCorrectResponse](bond::bonded<PingReply> response, Status status)
```
chwarr: This is kind of unwieldy, but it gets better once we can switch to wait_callback.

examples/cpp/grpc/pingpong/pingpong.cpp
chwarr: Are we missing an event like we have for `PingNoResponse`?

examples/cpp/grpc/pingpong/pingpong.cpp
lalo: i wanted parity with the c# one, which does throw an error

examples/cpp/grpc/pingpong/pingpong.cpp
lalo: done

examples/cpp/grpc/pingpong/pingpong.cpp
lalo: I'm not exercising this code, just wanted to get the codegen to compile.

lalo: done

examples/cpp/grpc/pingpong/pingpong.cpp
lalo: done

examples/cpp/grpc/pingpong/pingpong.cpp
lalo: done

examples/cpp/grpc/pingpong/pingpong.cpp
lalo: done

examples/cpp/grpc/pingpong/pingpong.cpp
lalo: it is on its own scope with the corresponding f_print

examples/cpp/grpc/pingpong/pingpong.cpp
lalo: right also fixed above

examples/cpp/grpc/pingpong/pingpong.cpp
lalo: right

compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
lalo: I'm aware of that, but I'll tackle it at the end since its also used by _comm

examples/cpp/grpc/pingpong/pingpong.cpp
chwarr: You should be able to check that it compiles with an empty method body.

lalo:
Merged in with commit 35b6b85b7466a7d3a5d992f63f125f06cc52e554

PR #457, opened 2017-05-25
More optimization of AppVeyor config

chadwalters:
Only load git submodules when they are needed. Remove redundant .NET Core build.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -83,12 +83,18 @@
             - BOND_BUILD: C#
               BOND_OUTPUT: Fields
               BOND_CONFIG: Fields
-            - BOND_BUILD: "C# .NET Core"
-              BOND_CONFIG: Debug
-
     install:
         - ps: >-
-            git submodule update --init --recursive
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -83,12 +83,18 @@
...
-
     install:
         - ps: >-
-            git submodule update --init --recursive
+            if (($env:BOND_BUILD -eq 'C++') -or ($env:BOND_BUILD -eq 'Python')) {
+
+                git submodule update --init thirdparty\rapidjson
+
+                # If gRPC is not disabled, init gRPC and its dependencies
+                if (($env:BOND_BUILD -eq 'C++') -and (-not ($env:BOND_CMAKE_FLAGS -match '-DBOND_ENABLE_GRPC=FALSE'))) {
```
chwarr: Minor: I think the lack of a blank line between the comment and the condition is OK because we're inside of another condition, but for robustness under future editing, I think we should have a blank line.

lalo: line 176 is also culprit of this (not part of this commit)

chwarr:
Merged to grpc++ in commit 2be617f1.

PR #459, opened 2017-05-26
[c++ grpc] Add wait_callback helper

chwarr:
The wait_callback helper can be used with the generated gRPC proxies to
synchronously wait for the proxy callback to be invoked.

* helloworld updated to use wait_callback
* gRPC compat client updated to use wait_callback
* unit tests added

TODO:

* [x] add unit test cases for copying wait_callback


cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,175 @@
...
+    {
+        return _impl->status();
+    }
+
+private:
+    /// The interesting guts of wait_callback. We use an impl class so that
+    /// we can handout weak_ptr instances to be captured by the callbacks
+    /// without forcing clients of wait_callback to deal with creating a
+    /// shared_ptr.
+    struct impl
```
ara-ayvazyan: Any reason for not using std::promise/future?

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,175 @@
...
+
+    /// @brief Creates a callback that will signal this wait_callback
+    /// instance.
+    ///
+    /// If multiple callbacks are created and invoked, only the results of
+    /// the first one that is invoked will be recorded.
+    ///
+    /// It is safe for callbacks to outlive the wait_callback that created
+    /// them. When they are invoked their results will be discarded.
+    CallbackType callback()
```
ara-ayvazyan: It would probably be better to return the lambda itself (auto). The caller can always decide to store it in std::function, but it won't be enforced by this implementation.

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,175 @@
...
+        mutable std::mutex _m;
+        /// condition variable used to signal anyone waiting
+        mutable std::condition_variable _cv;
+        /// The response, but more importantly, doubles as a flag indicating
+        /// whether a callback has been invoked yet. If this is empty, no
+        /// callback has been invoked yet. If non-empty, a callback has
+        /// already been invoked.
+        boost::optional<bond::bonded<TResponse>> _response;
+        /// The status.
+        boost::optional<grpc::Status> _status;
```
ara-ayvazyan: Do we really need two separate `boost::optional`-s instead of a single one for the aggregate type (they seem to be updated together)?

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,175 @@
...
+
+        impl() = default;
+        impl(const impl&) = delete;
+        impl(impl&&) = delete;
+        impl& operator=(const impl&) = delete;
+        impl& operator=(impl&&) = delete;
```
ara-ayvazyan: The `delete`s seem to be redundant.

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,175 @@
...
+
+#include <chrono>
+#include <condition_variable>
+#include <functional>
+#include <memory>
+#include <mutex>
+
+namespace bond { namespace ext { namespace gRPC {
+
+/// @brief A callback type that can be manually waited upon.
```
lalo: interesting use of upon instead of on

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,175 @@
...
+///
+/// The \ref wait member function can be used to wait until the callback
+/// has been called. Then, the \ref status and \ref response member
+/// functions can be called to inspect the results.
+///
+/// @example wait_callback.cpp
```
lalo: will it automatically find the file on the test dir?

doc/CMakeLists.txt
```diff
@@ -66,7 +66,11 @@ endfunction()
 find_haskell_program (pandoc)
 find_program (Doxygen_EXECUTABLE doxygen)
 
+message(STATUS "here1")
```
lalo: this is on purpose for some appveyor test or what?

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,175 @@
...
+    template <typename Rep, typename Period>
+    bool wait(std::chrono::duration<Rep, Period> timeout) const
+    {
+        return _impl->wait(timeout);
+    }
+
+    /// @brief Gets the response.
+    ///
+    /// @warning Blocks until a callback is invoked.
+    const bond::bonded<TResponse>& response() const
```
lalo: nit: result could make a little more sense or maybe I'm relating too much with Tasks from c#

cpp/inc/bond/ext/grpc/wait_callback.h
lalo: is it worthwhile to mention something regarding 'sync' or 'sync-ifying' the rpc calls?

cpp/inc/bond/ext/grpc/wait_callback.h
ara-ayvazyan: Even `std::packaged_task` might be helpful here.

cpp/inc/bond/ext/grpc/wait_callback.h
chwarr: I prefer to be explicit for types that cannot be copied or moved.

cpp/inc/bond/ext/grpc/wait_callback.h
chwarr: Renamed. It's a different file from the test cases.

cpp/inc/bond/ext/grpc/wait_callback.h
chwarr: I don't like the name Result: there are two pieces of information here: the status code and the response.

doc/CMakeLists.txt
chwarr: Removed. This was left over from debugging CMake.

cpp/inc/bond/ext/grpc/wait_callback.h
chwarr: Visual C++ 2013 doesn't support auto return types, and `-> decltype(/* the lambda expression*/)` doesn't work as lambda cannot appear unevaluated.

cpp/inc/bond/ext/grpc/wait_callback.h
chwarr: For the initial release, we're skipping promises/futures/tasks.

cpp/inc/bond/ext/grpc/wait_callback.h
chwarr: Good suggestion. Switched to a tuple for storage and flag. There are still two accessors.

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,180 @@
...
+    CallbackType callback()
+    {
+        // Create a weak_ptr to the implementation for the callback. If all
+        // the wait_callback instances sharing the implementation go away,
+        // their callbacks can still safely be invoked: they just won't be
+        // able to lock the weak_ptr and thus won't do anything.
+        std::weak_ptr<impl> wpImpl(_impl);
+
+        return std::function<void(const bond::bonded<TResponse>&, const grpc::Status&)>(
+            [wpImpl](const bond::bonded<TResponse>& response, const grpc::Status& status)
```
ara-ayvazyan: The `if (auto impl = wpImpl.lock) { ...  }` is better in terms of locality.

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,180 @@
...
+    {
+        return _impl->status();
+    }
+
+private:
+    /// The interesting guts of wait_callback. We use an impl class so that
+    /// we can handout weak_ptr instances to be captured by the callbacks
+    /// without forcing clients of wait_callback to deal with creating a
+    /// shared_ptr.
+    class impl
```
ara-ayvazyan: Any reason for not using `std::packaged_task`/`std::future`/`std::promise`?

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,180 @@
...
+/// The wait() member function can be used to wait until the callback has
+/// been called. Then, the status() and response() member functions can be
+/// called to inspect the results.
+template <typename TResponse>
+class wait_callback
+{
+public:
+    wait_callback() : _impl(std::make_shared<impl>()) { }
+
+    using CallbackType = std::function<void(const bond::bonded<TResponse>&, const grpc::Status&)>;
```
ara-ayvazyan: Is that typedef really needed? It would be harder to remove later.

cpp/inc/bond/ext/grpc/wait_callback.h
ara-ayvazyan: Oh, forgot that VC12 needs to be supported. In that case maybe you could have a check and return `auto` when possible? Also in the return statement you can just return the lambda, I don't think you need to explicitly cast to `std::function`.

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,180 @@
...
+    ///
+    /// It is safe for callbacks to outlive the wait_callback that created
+    /// them. When they are invoked their results will be discarded.
+    CallbackType callback()
+    {
+        // Create a weak_ptr to the implementation for the callback. If all
+        // the wait_callback instances sharing the implementation go away,
+        // their callbacks can still safely be invoked: they just won't be
+        // able to lock the weak_ptr and thus won't do anything.
+        std::weak_ptr<impl> wpImpl(_impl);
```
ara-ayvazyan: Init directly in the capture list (i.e. `[wpImpl = std::weak_ptr<impl>{ _impl }]`)?

cpp/inc/bond/ext/grpc/wait_callback.h
chwarr: We're skipping tasks/futures/promises completely for the initial release.

cpp/test/grpc/wait_callback.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include <bond/protocol/compact_binary.h>
+
+#include <atomic>
+#include <thread>
+
+using wait_callbackBox = bond::ext::gRPC::wait_callback<bond::Box<int>>;
+
+class wait_callback_tests
+{
+    static const int ANY_INT_VALUE = 100;
```
ara-ayvazyan: Why is everything `static` here instead of simply instantiating once?

cpp/inc/bond/ext/grpc/wait_callback.h
chwarr: Doesn't work in Visual C++ 2013.

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,180 @@
...
+        impl& operator=(impl&&) = delete;
+
+        /// Attempts to set _results. Only the first call to set that
+        /// completes will actuall set _results. The rest will do nothing.
+        void set(const bond::bonded<TResponse>& response, const grpc::Status& status)
+        {
+            std::unique_lock<std::mutex> lock(_m);
+            if (!_results)
+            {
+                _results.emplace(std::make_tuple(response, status));
```
ara-ayvazyan: I don't think you need `make_tuple` with `emplace`.

cpp/test/grpc/wait_callback.cpp
chwarr: I don't follow the question.

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        void wait() const
+        {
+            std::unique_lock<std::mutex> lock(_m);
+            return _cv.wait(lock, [this]() { return static_cast<bool>(_results); });
+        }
+
+        template <typename Rep, typename Period>
+        bool wait(std::chrono::duration<Rep, Period> timeout) const
```
ara-ayvazyan: I would probably accept a `const duration&` (at least std does like that, must be for a reason).

cpp/inc/bond/ext/grpc/wait_callback.h
chwarr: Yeah. I can make it private.

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,180 @@
...
+                    impl->set(response, status);
+                }
+            });
+    }
+
+    /// @brief Implicitly creates a callback.
+    operator CallbackType()
+    {
+        return callback();
+    }
```
ara-ayvazyan: Wouldn't be simpler and more natural to make the `wait_callback` a function object (callback) itself by adding `operator()`?

doc/CMakeLists.txt
```diff
@@ -135,6 +131,7 @@ message(STATUS "here2")
         doxygen/bond_layout.xml
         doxygen/bond_reference.css
         doxygen/mainpage.md
+        doxygen/examples/wait_callback.cpp
```
lalo: wait_callback_example.cpp

cpp/test/grpc/wait_callback.cpp
ara-ayvazyan: Why all members need to be marked as `static`?

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,180 @@
...
+
+    /// @brief Waits at least \p timeout for a callback created by
+    /// callback() to be invoked.
+    ///
+    /// @param timeout the minimum amount of time to wait.
+    ///
+    /// @return \p true if a callback was invoked. \p false if the timeout
+    /// occured.
+    template <typename Rep, typename Period>
+    bool wait(std::chrono::duration<Rep, Period> timeout) const
```
ara-ayvazyan: Accept `const duration&`.

cpp/inc/bond/ext/grpc/wait_callback.h
chwarr: Yes, yes it would. Thanks!

cpp/inc/bond/ext/grpc/wait_callback.h
chwarr: Done.

doc/CMakeLists.txt
chwarr: Fixed.

cpp/inc/bond/ext/grpc/wait_callback.h
chwarr: Done.

cpp/inc/bond/ext/grpc/wait_callback.h
chwarr: No longer applicable after switching to `operator ()(response, status)`.

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /// @brief Waits at least \p timeout for this to have been invoked.
+    ///
+    /// @param timeout the minimum amount of time to wait.
+    ///
+    /// @return \p true if a callback was invoked. \p false if the timeout
+    /// occured.
+    template <typename Rep, typename Period>
+    bool wait(const std::chrono::duration<Rep, Period>& timeout) const
```
ara-ayvazyan: Rename to `wait_for`? Might be easier to move to `std::future` later.

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,121 @@
...
+        impl() = default;
+        impl(const impl&) = delete;
+        impl(impl&&) = delete;
+        impl& operator=(const impl&) = delete;
+        impl& operator=(impl&&) = delete;
+
+        /// mutex to lock the shared state
+        mutable std::mutex _m;
+        /// condition variable used to signal anyone waiting
+        mutable std::condition_variable _cv;
```
ara-ayvazyan: Do you still need `mutable`s here?

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -0,0 +1,121 @@
...
+template <typename TResponse>
+class wait_callback
+{
+public:
+    wait_callback() : _impl(std::make_shared<impl>()) { }
+
+    /// @brief Records the response and status.
+    ///
+    /// May be invoked multiple times and from different threads, but only the first
+    /// invocation will be recorded.
```
ara-ayvazyan: Is this design approach intentional and required? I agree with `std::promise::set_value` which throws exception on second call and does not hide issues from user.

cpp/inc/bond/ext/grpc/wait_callback.h
chwarr: Good idea. In a separate PR I'll make a similar change to the other synchronization primitives.

cpp/inc/bond/ext/grpc/wait_callback.h
chwarr: I like it. Switched.

cpp/inc/bond/ext/grpc/wait_callback.h
chwarr: Nope. Removed.

cpp/inc/bond/ext/grpc/wait_callback.h
ara-ayvazyan: Mmm, then isn't it more important to make it explicit for a public type (`wait_callback` in this case) rather than a private struct? :)

cpp/test/grpc/wait_callback.cpp
```diff
@@ -0,0 +1,167 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+// TODO: move unit_test_framework.h to cpp/test/inc
+#include "../core/unit_test_framework.h"
```
ara-ayvazyan: Maybe better to adjust the build script to make, let's say `../core`, as an additional include directory?

chwarr: This isn't something I'm going to fix in this pull request. It's in a TODO comment, and there are many other copies of the same TODO comment. They should be fixed together.

chwarr:
Merged in commit 9b384148749388163da428.

PR #463, opened 2017-05-27
[c++ grpc] Take durations by const reference

chwarr:
For wait()-like functions we now take std::chrono::duration arguments by
const reference, just like std::condition_variable does.


chadwalters:
Looks good. CI failure due to known issue not related to this pull request.

PR #466, opened 2017-05-30
Address race condition in gRPC compat tests

chadwalters:
Ensure that the server has shutdown once all the expected requests have
been received and response have been queued up.

### DIFF
--- a/cpp/test/compat/grpc/pingpong_server.cpp
+++ b/cpp/test/compat/grpc/pingpong_server.cpp
@@ -54,9 +54,9 @@ public:
                 response.Payload = request.Payload;
 
                 NumRequestsReceived++;
-                Countdown.set();
 
                 call.Finish(bond::bonded<PingResponse>{ response }, Status::OK);
+                Countdown.set();
                 break;
             }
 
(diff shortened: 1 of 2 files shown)


cpp/test/compat/grpc/pingpong_server.cpp
```diff
@@ -106,6 +105,9 @@ int main()
 
     bool countdownSet = Countdown.wait(std::chrono::seconds(30));
 
+    server->Shutdown();
```
chwarr: Consider using the overload that takes a timeout after which it will forcefully fail the in-flight RPC operations.

cs/test/compat/grpc/server/Program.cs
```diff
@@ -76,6 +76,8 @@ static void Main(string[] args)
 
             bool countdownSet = Countdown.Wait(30000);
 
+            pingServer.ShutdownAsync().GetAwaiter().GetResult();
```
chwarr: Consider using[`Task.Wait(TimeSpan)`](https://msdn.microsoft.com/en-us/library/dd235606(v=vs.110).aspx) so the test doesn't block here forever if the RPC methods never return for some reason.

chadwalters:
Addressed issues with potential shutdown hangs.

chwarr:
Merged into grpc++ in commit 12164849 .

PR #467, opened 2017-05-30
[c++ grpc] Upgrade to gRPC v1.3.4

chwarr:
ClientAsyncResponseReader had a breaking API change that we needed to
adjust to.

There also appears to be a bug with ClientAsyncResponseReader's
placement new and placement delete operators, so we're consuming a
slightly patched version (see
https://github.com/chwarr/grpc/commit/8bd0fb92ec for the delta).
Upstream issue https://github.com/grpc/grpc/issues/11301 has been opened
about this problem.

### DIFF
--- a/cpp/inc/bond/ext/grpc/detail/client_call_data.h
+++ b/cpp/inc/bond/ext/grpc/detail/client_call_data.h
@@ -78,7 +78,7 @@ struct client_unary_call_data : io_manager_tag
         const TRequest& request)
     {
         _responseReader = std::unique_ptr<grpc::ClientAsyncResponseReader<TResponse>>(
-            new ::grpc::ClientAsyncResponseReader<TResponse>(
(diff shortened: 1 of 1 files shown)


lalo:
There seems to be an error on travis-ci when compiling grpc dependencies

chwarr:
Looking at Travis failures. They're weird... Nothing appears to have changed in the CMakeLists.txt for zlib between the versions...

lalo:
https://stackoverflow.com/questions/30794859/xcode-throws-compile-errors-in-zlib-h

https://github.com/madler/zlib/pull/7/files

chwarr:
Looks like this may have something to do with the zlib v1.2.8 to v.1.2.11 change.

chwarr:
I have a local repro and am starting to bisect to figure out what change introduced this...

lalo:
https://github.com/madler/zlib/pull/132

the last comment could be helpful, in the case it is grabbing the local zlib files

PR #468, opened 2017-05-31
Simplify bonded from client and server

lalo:


cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -157,7 +157,16 @@ class unary_call final
...
+    void Finish(const TResponse& msg, const grpc::Status& status = grpc::Status::OK)
+    {
+        _impl->Finish(bond::bonded<TResponse>{msg}, status);
+    }
+
+    /// @brief Responds to the client with the given message and status.
+    ///
+    /// Only the first call to \p Finish or \p FinishWithError will be
+    /// honored.
+    void Finish(const bond::bonded<TResponse>& msg, const grpc::Status& status = grpc::Status::OK)
```
chwarr: Good idea giving the status a default value of OK.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -157,7 +157,16 @@ class unary_call final
     ///
     /// Only the first call to \p Finish or \p FinishWithError will be
     /// honored.
-    void Finish(const TResponse& msg, const grpc::Status& status)
+    void Finish(const TResponse& msg, const grpc::Status& status = grpc::Status::OK)
+    {
+        _impl->Finish(bond::bonded<TResponse>{msg}, status);
```
chwarr: I have a slight preference to have this Finish overload delegate to the bonded overload in this class instead of directly to the impl. It minimizes the number of entry points to impl, if we need to add sandwich code for some reason.

chwarr:
Merged into grpc++ in commit 84dfc3cb.

Issue #464, opened 2017-05-27
The documentation should state that structs cannot be default to nothing


ramon-garcia:
I was confused when I learned that Bond structure fields cannot be default to nothing.

While I think that this should be allowed (issue #226 ) , meanwhile the documentation should match the behavior.

In the "A Young Person's Guide to C++ Bond", /doc/src/bond_cpp.md , the documentation says:

> Since a nullable type must represent the additional value of null, its serialized representation necessarily incurs some overhead compared to the base type. Often it is more efficient to avoid using nullable type and instead designate one of normal values to handle the special case that otherwise would be represented by null. For example empty is usually a good choice for string and container types and 0 for arithmetic types. Another option that may sometimes be appropriate is setting default value of a field to nothing.
> 
> The canonical scenario where a nullable type is the right choice is recursive structures. For example here's how Bond TypeDef struct is defined:

The sense of the text and the example indicate that default nothing should be preferred to nullable, except for special cases like recursive structures. Therefore, non recursive structure fields, for instance should be default noting.

This text should say that for structure fields, default of nothing is not posible and nullable must be chosen.

(In the c# guide, /doc/src/bond_cs.md , I could not find any reference to the choice between default nothing and nullable)

chwarr:
You're understanding of how the type system works is correct. Take a look at the refinements in https://github.com/Microsoft/bond/pull/469. I think it reads better now.

(As an aside, there's rendered documentation ([C++](https://microsoft.github.io/bond/manual/bond_cpp.html) and [C#](https://microsoft.github.io/bond/manual/bond_cs.html)) that I find easier to read.)

See also https://github.com/Microsoft/bond/issues/115

ramon-garcia:
The documentation is still confusing, because it seems to intend to say that nullabe types are not essential for optional structure fields:

"The canonical scenario where a nullable type is the right choice is recursive structures. For example here's how Bond `TypeDef` struct is defined"


>  struct TypeDef
>   {
>         ...
>         2: nullable<TypeDef> element;
>         ...
>   }

with this example, the choice of nullable does not have anything to do with recursion. It has to do with the element being a structure field, and being optional.

This is what I find confusing.

chwarr:
[...]

PR #469, opened 2017-05-31
[doc] Refine nullable, nothing, and default values

chwarr:
* Add an explicit section about default values
* Be explicit about what field types can have explicit vs. implicit vs.
  nothing default values.
* Improve cross-references.
* Copy relevant sections for C++ manual to C# manual. (This is the best
  we can do until we switch to something that let's us import section
  into multiple pages.

Closes https://github.com/Microsoft/bond/issues/464

### DIFF
--- a/doc/src/bond_cpp.md
+++ b/doc/src/bond_cpp.md
@@ -242,19 +242,44 @@ Some best practices and other considerations to keep in mind:
   break text-based protocols like [SimpleJsonProtocol](#simple-json)
 - `required` should be used sparingly and only with careful consideration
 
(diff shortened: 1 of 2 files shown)


chwarr:
AppVeyor build failure is due to the known failure with ansi-terminal building on Windows.
Commit https://github.com/Microsoft/bond/commit/26cb3046ffa944f7936a35291d7a968bcc4e4b34 works around the issue, and upstream issue https://github.com/feuerbach/ansi-terminal/issues/22 is almost fixed.

PR #476, opened 2017-06-01
[c++] Mark custom stream interface BOND_NORETURN

chwarr:
The unspecialized custom stream inferface functions that exist solely to
describe the interface have static asserts that are always false.
However, they are also non-void functions that have no return
statements. g++ 5.4.0 emits a warning about the lack of a return
statement. It apparently cannot see that the static assert is always
false. BOND_NORETURN is used to indicate that these methods don't
actually ever return anything.


chwarr:
CC @ara-ayvazyan, if you're interested.

PR #477, opened 2017-06-01
[c++ grpc] Upgrade to gRPC v1.3.4

chwarr:
ClientAsyncResponseReader had a breaking API change that we needed to
adjust to.

The gRPC zlib submodule is now conflicting with the system-wide zlib.
When using CMake inside of Travis, we prefer the system-wide zlib
package over the submodule.

There also appears to be a bug with ClientAsyncResponseReader's
placement new and placement delete operators, so we're consuming a
slightly patched version (see
https://github.com/chwarr/grpc/commit/8bd0fb92ec for the delta).
Upstream issue https://github.com/grpc/grpc/issues/11301 has been opened
about this problem.

This PR supersedes #467, which was closed with the underlying grpc++ branch was merged into master and deleted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ In the root `bond` directory run:
 ```bash
 mkdir build
 cd build
-cmake ..
+cmake .. -DBOND_ENABLE_COMM=FALSE -DBOND_ENABLE_GRPC=FALSE
 make
 sudo make install
 ```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -99,6 +100,9 @@ sudo apt-get install \
     python2.7-dev
 
 cabal install happy
+
+cd build # or wherever you ran CMake before
+cmake .. -DBOND_ENABLE_COMM=TRUE -DBOND_ENABLE_GRPC=TRUE -DgRPC_ZLIB_PROVIDER=package
```
lalo: well this is only the case if it is previously installed, since in windows it would work without

chwarr: The instructions for building Bond Code include installing the package "zlib1g-dev", so there should be a system-wide zlib available for gRPC to use.

chwarr:
The gRPC compat tests are failing due to some races. @chadwalters is looking into these problems in more detail.

PR #481, opened 2017-06-02
ClientContext lifetime improvements (WIP)

chwarr:
Here's a checkpoint of some work to improve the ClientContext lifetime management that @chadwalters  and I've been working on. This PR is here to get Travis and AppVeyor runs and as a place to collect comments. I expect it will be reworked heavily before being merged.

Current improvements:

* take ClientContext via shared pointer in the proxy methods so the context can be kept alive for the duration of the client call. Before, this was the client's responsibility, and failure to do so resulted in crashes.
* include the ClientContext in what's passed back to the callback so that the caller can "forget" about the context if they don't need it in the interim
* minimize copies by reading the response and status into a client_callback_args that we then move into the callback

Potential future improvements:

* add overloads that create a client context so callers who don't need to muck with the context don't even need to worry about one at all
* adjust naming to reflect that this is only what unary client calls will use

### DIFF
--- a/cpp/inc/bond/ext/grpc/client_callback.h
+++ b/cpp/inc/bond/ext/grpc/client_callback.h
@@ -1,0 +1,63 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#pragma once
+
+#include <grpc++/impl/codegen/status.h>
+#include <grpc++/client_context.h>
+
+#include <bond/core/bonded.h>
+
+#include <memory>
+#include <utility>
+
+namespace bond { namespace ext { namespace gRPC {
+
+    /// @brief The client-side results of a unary call.
+    template <typename TResponse>
+    struct unary_call_result
+    {
+        /// @brief The response received from the service.
+        ///
+        /// @note Depending on the implementation of the service, this may or
+        /// may not contain an actual response. Consult the documentation for
+        /// the service to determine under what conditions it sends back a
+        /// response.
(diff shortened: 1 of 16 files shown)


cpp/inc/bond/ext/grpc/client_callback.h
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#pragma once
+
```
chadwalters: #include <bond/core/bonded.h> (got removed during some intermediate iteration)

cpp/inc/bond/ext/grpc/client_callback.h
```diff
@@ -0,0 +1,41 @@
...
+namespace bond { namespace ext { namespace gRPC {
+
+    template <typename TResponse>
+    struct client_callback_args
+    {
+        bond::bonded<TResponse> response;
+        grpc::Status status;
+        std::shared_ptr<grpc::ClientContext> context;
+
+        explicit client_callback_args(std::shared_ptr<grpc::ClientContext> context)
```
chadwalters: Why is this form needed?

chwarr: When the proxy is creating its `client_unary_call_data` instance, which has a `client_callback_args` inside of it, it only has a ClientContext. Later, when reading the response, the `response` and `status` members will be updated in-place.

chwarr:
Yes, I'll be fixing up the commit before merging.

PR #486, opened 2017-06-05
java compat: Initial commit.

tstein:
This decides whether to run Java compat and gets it set up properly on
Win/Lin/Mac. It does this based on the existence of bond.jar and
compat.jar, meaning that, on a fresh tree, you need to run cmake again
after building gbc and the two jars.

Currently, all implemented protocols fail with EOFExceptions, and all
unimplemented protocols report themselves as failures.


chwarr:
Looks like the JVM branch needs the [C# on Travis mdb/pdb fix in master](https://github.com/Microsoft/bond/commit/107e3facd0b66c97967f99e05265233ddaa74441).

tstein:
Yeah. Let's just do a merge from master after the rapidjson fix goes through.

cpp/test/compat/core/compat.cmake
```diff
@@ -20,3 +20,20 @@ if (CSHARP_COMPAT)
...
     endif()
 endif()
+
+if (JAVA_COMPAT)
+    if (WIN32)
+        set (PATHSEP "\\\;")
+    else()
+        set (PATHSEP ":")
+    endif()
+
```
chwarr: Consider making this error out if `JAVE_CORE` isn't set, as `JAVA_COMPAT` implies that `JAVA_CORE` is also set.

cpp/test/compat/core/compat.cmake
```diff
@@ -20,3 +20,20 @@ if (CSHARP_COMPAT)
...
+    run (java -classpath ${JAVA_CORE}${PATHSEP}${JAVA_COMPAT} com.microsoft.bond.compat.CompatDriver
```
chwarr: Does the classpath need to be "quoted", or will CMake do that for us?

java/compat/gradlew
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env sh
```
chwarr: We already have one of these. Can we keep them in some shared location and have each directories just invoke the shared ones?

java/compat/src/main/java/com/microsoft/bond/compat/CompatDriver.java
```diff
@@ -0,0 +1,65 @@
...
+        final String toProtocol;
+        if (args.length == 4) {
+            toProtocol = args[3];
+        } else {
+            toProtocol = fromProtocol;
+        }
+
+        final FileInputStream input = new FileInputStream(inputFile);
+        final FileOutputStream output = new FileOutputStream(outputFile);
+        final Compat compat = new Compat();
```
chwarr: Minor: can we move this closer to its first use?

java/compat/src/main/java/com/microsoft/bond/compat/CompatDriver.java
```diff
@@ -0,0 +1,65 @@
...
+
+        if (fromProtocol.equals("fast")) {
+            reader = new FastBinaryReader(input, (short) 1);
+        } else if (fromProtocol.equals("compact")) {
+            reader = new CompactBinaryReader(input, (short) 1);
+        } else if (fromProtocol.equals("compact2")) {
+            reader = new CompactBinaryReader(input, (short) 2);
+        } else {
+            System.err.println("Unsupported input protocol: " + fromProtocol);
+            System.exit(255);
```
sigansky: What is 255? Can you please add a comment and/or make it a constant?

sigansky:
:shipit:

java/compat/gradlew
tstein: ```gradlew```, ```gradlew.bat```, and the jars in ```gradle/``` are meant to allow you to check out and build without installing the build tool globally. It's common in gradle projects, but I'm not a huge fan. I'd like to investigate getting rid of them in the near future in favor of assuming you have gradle installed, but I have some questions about how that interacts with IDEA that I need to answer first.

I initially symlinked these files to the ones in core, but I couldn't get git on Windows to handle that properly. I've got a branch going on the side to try to have a single gradle entry point for all of Java, but I think that may depend on getting the gbc hook working.

In short, I would like to accept the redundancy for the moment.

java/compat/.gitignore
```diff
@@ -0,0 +1,4 @@
+.gradle/
```
tstein: It shouldn't appear anywhere else, but changed anyway in core and compat.

java/compat/.idea/uiDesigner.xml
```diff
@@ -0,0 +1,124 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="Palette2">
+    <group name="Swing">
+      <item class="com.intellij.uiDesigner.HSpacer" tooltip-text="Horizontal Spacer" icon="/com/intellij/uiDesigner/icons/hspacer.png" removable="false" auto-create-binding="false" can-attach-label="false">
```
tstein: Auto-generated. I was a little worried about IDEA bringing it back if I deleted it, but it hasn't tried yet. Removed from core and compat.

java/compat/src/main/java/com/microsoft/bond/compat/CompatDriver.java
tstein: Done.

java/compat/src/main/java/com/microsoft/bond/compat/CompatDriver.java
tstein: A number that isn't zero. I have a soft expectation that 255 indicates a usage error and failures following correct invocations start from 1, but don't intend to formalize that until we want a machine to get any info other than success/failure from this program.

Lifted to `static int STATUS_FAILURE`.

cpp/test/compat/core/compat.cmake
tstein: I just tried doing a build on a Linux desktop on a folder with a space in it... we have a lot of these problems.

I was building out of a directory called ```bond space```, my build failed very early because cpps weren't getting copied to the build directory, and I ended up with a directory called ```bond\\ space``` parallel to my root. That's where they went.

As-is, this is probably going to fail, but it's deficient in exactly the way the C# equivalent is, and we may need to change a lot more to even get to those failures. I'd like to leave it.

cpp/test/compat/core/compat.cmake
tstein: Done.

java/compat/gradlew
chwarr: Symlinks under Git on Windows are currently an experimental feature, so we can't rely on them working. I'm ok with the redundancy for now.

cpp/test/compat/core/compat.cmake
chwarr: Hmmm... Sounds like we've got some bugs elsewhere in the CMake file. C# codegen and build [works with spaces](https://github.com/Microsoft/bond/tree/e81d51b33b9329642860f02b7fba5733f1c5b96f/cs/test/core/import%20dir%20with%20spaces). Can you open an issue? I'm not able to quickly reproduce the issue on Windows with MSBuild.

Issue #485, opened 2017-06-05
fix C++/rapidjson build under clang 3.9


tstein:
Clang 3.9 added [`-Wexpansion-to-defined`](https://www.mail-archive.com/cfe-commits@lists.llvm.org/msg10996.html), which breaks the 1.0.0 release of rapidjson that we have. They fixed it in [this commit](https://github.com/miloyip/rapidjson/commit/ee4207b3f0d8f6d646ada8657f0b6976a2fb8ed8), which was included in their 1.1.0 release.

PR #487, opened 2017-06-05
Update rapidjson to v1.1.0.

tstein:
v1.0.0 triggers a new warning in clang 3.9. v1.1.0 includes a fix for
this.

Changelogs for new releases brought in by this change:
* [v1.1.0](https://github.com/miloyip/rapidjson/releases/tag/v1.1.0)
* [v1.0.2](https://github.com/miloyip/rapidjson/releases/tag/v1.0.2)
* [v1.0.1](https://github.com/miloyip/rapidjson/releases/tag/v1.0.1)

resolves https://github.com/Microsoft/bond/issues/485


chwarr:
Looks like something in RapidJSON has changed such that MSVC is now detecting unreachable code where is wasn't before. Can you take care of the new warnings too? Also, check that the warning fix doesn't cause problems with RapidJSON v1.0.0, so that we can sort-of support v1.0.0 if possible.

Can you add a note to the changelog?

tstein:
That's weird - the appveyor warnings point to two spots [like this](https://github.com/Microsoft/bond/blob/master/cpp/inc/bond/core/select_protocol.h#L73) in our code, and they appear to be correct - we have code after an unconditional throw. I wonder why we didn't see this previously.

chadwalters:
The warning from select_protocol seems like it might be indicating a real issue. I'm digging into it.

chadwalters:
Closing this in favor of https://github.com/Microsoft/bond/pull/491.

PR #490, opened 2017-06-07
Extend IDL syntax to support service inheritance

sapek:
Add syntax for inheritance in service definitions, e.g.:

    service Foo
    {
        Result foo(Param);
    }

    service Bar : Foo
    {
        Result bar(Param);
    }

In the AST a service base is represented as `Maybe Type`, the same way
as struct bases are represented. The parser validates that only a
concrete service or an instance of a generic service is used as a
base.


sapek:
Done. [Diff ignoring whitespaces](https://github.com/Microsoft/bond/pull/490/commits/7be68d87e6b8c9551aacd7fcf797b7dea61617fc?w=1) for easier review.

chadwalters:
We agreed previously to put support for this experimental syntax behind a switch that people have to explicitly opt into, in part because we don't have a full design yet for how service inheritance gets exposed in comm/gRPC. I realize that in practice someone is unlikely to get bitten by this but I think that we should stick with what the group decided as the best practice for this kind of experimental syntactic IDL extension.

sapek:
Added `--enable-service-inheritance` switch to enable support for the new syntax. Support for codegen of services with inheritance is separately guarded by RPC codegen switches based on which templates support inheritance (initially neither gRPC nor Comm do).

```
E:\oss\bond\compiler (service-inheritance)
λ dist\build\gbc\gbc.exe c# tests\schema\service.bond
gbc.exe: user error (Use --enable-service-inheritance to enable service inheritance syntax.)

E:\oss\bond\compiler (service-inheritance)
λ dist\build\gbc\gbc.exe c# --comm tests\schema\service.bond
gbc.exe: user error (Use --enable-service-inheritance to enable service inheritance syntax.)

E:\oss\bond\compiler (service-inheritance)
λ dist\build\gbc\gbc.exe c# --enable-service-inheritance --comm tests\schema\service.bond
gbc.exe: user error (Service inheritance is not supported in Comm codegen.)

E:\oss\bond\compiler (service-inheritance)
λ dist\build\gbc\gbc.exe c# --enable-service-inheritance --grpc tests\schema\service.bond
gbc.exe: user error (Service inheritance is not supported in gRPC codegen.)
```

compiler/Main.hs
```diff
@@ -116,6 +118,12 @@ csCodegen options@Cs {..} = do
                         ]
 csCodegen _ = error "csCodegen: impossible happened."
 
+anySeriveInheritnace :: [Declaration] -> Bool
```
chadwalters: Spelling: "anyServiceInheritance"

compiler/tests/generated/exports/service_comm.h
```diff
@@ -1162,4 +1162,164 @@ namespace tests
     };
     
 
+    class Bar
```
chadwalters: This isn't actually doing anything with the base service is it? Perhaps you could isolate out the tests around the experimental syntax into a separate .bond file that doesn't impact the mainline tests?

compiler/tests/schema/generic_service.bond
```diff
@@ -7,3 +7,7 @@ service Foo<Payload>
     Payload foo33(Payload);
 }
 
+service Bar<Payload> : Foo<Payload>
```
chadwalters: As I mentioned above, I'd prefer to see a different .bond file used for the experimental syntax.

compiler/Main.hs
sapek: Fixed.

compiler/tests/generated/exports/service_comm.h
sapek: Moved inheritance test cases to `service_inheritance.bond` and run only syntax/AST tests, not codegen test since there is no codegen support for services with base yet.

compiler/Main.hs
```diff
@@ -116,6 +118,12 @@ csCodegen options@Cs {..} = do
                         ]
 csCodegen _ = error "csCodegen: impossible happened."
 
+anySeriveInheritance :: [Declaration] -> Bool
+anySeriveInheritance = getAny . F.foldMap serviceWithBase
```
chadwalters: Spelling: "anyServiceInheritance" (weird, I thought I'd already made this comment).

sapek: There were two spelling errors, one in `Service` and one in `Inheritance`.

sapek:
Chad, is there anything else you want me to change?

PR #494, opened 2017-06-09
ByteArrayInputStream cloning + misc

tstein:
9f07440 adds functionality that should be sufficient for `bonded` in memory streams.
1badec2 fixes a mistake from my earlier reorg into `java/core`.
bda0607 adds a shared test runner (configured for line coverage), which will make it a little easier to get new copies of the repo up and running.

### DIFF
--- a/java/core/src/main/java/com/microsoft/bond/protocol/Cloning.java
+++ b/java/core/src/main/java/com/microsoft/bond/protocol/Cloning.java
@@ -4,7 +4,6 @@ import java.io.ByteArrayInputStream;
 import java.lang.reflect.Field;
 
 class Cloning {
-    private static final Object byteArrayInputStreamLock = new Object();
     private static final Field byteArrayInputStream_buf;
     private static final Field byteArrayInputStream_pos;
     private static final Field byteArrayInputStream_count;
@@ -19,6 +18,11 @@ class Cloning {
             // fields since JDK 1.0.
             throw new RuntimeException(e);
         }
+
+        // These changes are scoped to our Field instances and need not be cleaned up.
+        byteArrayInputStream_buf.setAccessible(true);
+        byteArrayInputStream_pos.setAccessible(true);
+        byteArrayInputStream_count.setAccessible(true);
     }
 
     static ByteArrayInputStream clone(ByteArrayInputStream stream) {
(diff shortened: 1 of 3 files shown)


java/core/src/main/java/com/microsoft/bond/protocol/Cloning.java
```diff
@@ -0,0 +1,54 @@
...
+        try {
+            buf = (byte[]) bufField.get(stream);
+            pos = posField.getInt(stream);
+            count = countField.getInt(stream);
+        } catch (final IllegalAccessException e) {
+            // This should never happen. We just made those fields accessible.
+            throw new RuntimeException(e);
+        }
+
+        bufField.setAccessible(false);
```
chwarr: Does this restoration need to be done inside a `finally`?

chwarr: What if two clones are happening concurrently across threads and the first one starts resetting accessibility after the second one has gotten past line 32?

java/core/src/main/java/com/microsoft/bond/protocol/Cloning.java
```diff
@@ -0,0 +1,54 @@
...
+        posField.setAccessible(false);
+        countField.setAccessible(false);
+
+        final ByteArrayInputStream cloned = new ByteArrayInputStream(buf, 0, count);
+        cloned.skip(pos);
```
chwarr: A comment about why `skip` is being used instead of passing `pos` to the ctor would be helpful.

java/core/src/main/java/com/microsoft/bond/protocol/Cloning.java
```diff
@@ -0,0 +1,54 @@
...
+        //
+        // There's one more piece of state inside that class - mark. We don't
+        // use mark()/reset(), and cloned streams shouldn't leak out of the
+        // library, so we ignore it here.
+
+        final Field bufField;
+        final Field posField;
+        final Field countField;
+        try {
+            bufField = ByteArrayInputStream.class.getDeclaredField("buf");
```
chwarr: Can/should the results of `getDeclaredField` be looked up once and cached?

tstein: That's a good question. `Field.get()` doesn't promise that it can be used concurrently, and I'm having a hard-time convincing myself it can. Found a few random conversations on the internet where people state that they believe it's safe to cache your `Field` (or your `Method`, from the same package), but I can't find proof or a promise.

(see one discussion [from SO](https://stackoverflow.com/questions/5349383/is-java-lang-reflect-method-thread-safe) and another [from 2006](https://forums.manning.com/posts/list/10293.page).)

I could convince myself to do what you're asking if I dug deep enough into `java.lang.reflect`, but I'm feeling like it's not worth the time. What I have in this PR is more obviously safe and is only invoked once per `bonded`. I can imagine a large container of `bonded` or a machine-generated struct with a lot of them, but these seem like niche cases.

Your call - want me to go deep on this now, or make an issue to check it out later?

java/core/src/main/java/com/microsoft/bond/protocol/Cloning.java
tstein: See below.

tstein: You're right - this is going to be a problem. I'm not seeing a way to do this without locking, so I'm gonna rewrite it with a lock, at which point there's no reason not to cache `Field`s. Standby for i2.

java/core/src/main/java/com/microsoft/bond/protocol/Cloning.java
tstein: You got it.

tstein: I wrote a whole paragraph justifying this and then realized it isn't necessary. No more `skip()`. 🤔

PR #496, opened 2017-06-12
[C++] Suppress Comm codegen when Comm not enabled

chadwalters:


cpp/CMakeLists.txt
```diff
@@ -5,9 +5,13 @@ set (core_schemas
...
 list (APPEND schemas ${core-schemas})
 
+if (BOND_ENABLE_COMM)
+    set(comm-flag COMM)
+endif()
+
 add_bond_codegen (${core_schemas}
     ENUM_HEADER
-    COMM
+    comm-flag
```
chwarr: This is currently passing [a literal `comm-flag` to the function](https://ci.appveyor.com/project/MicrosoftBond/bond/build/1.0.194/job/jlv1647ysrs0kjbr#L1454). I think we want `${comm-flag}`. Hence, CI builds are broken.

chadwalters: Yep. Fixed.

PR #499, opened 2017-06-14
java core: Implement file stream cloning.

tstein:


java/core/src/main/java/com/microsoft/bond/protocol/Cloning.java
```diff
@@ -1,9 +1,24 @@
...
 import java.io.ByteArrayInputStream;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
 import java.lang.reflect.Field;
+import java.nio.ByteBuffer;
+import java.nio.MappedByteBuffer;
+import java.nio.channels.FileChannel;
 
 class Cloning {
```
sigansky: >Cloning [](start = 6, length = 7)

Please declare all classes final, it is a useful hint to compiler optimizations.

tstein: Not sure it matters, since static methods are static dispatch (and sorta implicitly final), but done.

PR #502, opened 2017-06-16
Java: added support for generic types, and updated existing serialization framework to use it for all types.

sigansky:


java/core/src/main/java/com/microsoft/bond/BondSerializable.java
```diff
@@ -3,99 +3,16 @@
...
-import com.microsoft.bond.protocol.ProtocolWriter;
-import com.microsoft.bond.protocol.TaggedProtocolReader;
-
-import java.io.IOException;
-import java.io.InputStream;
-
-
 /**
- * Enables serialization of Bond classes.
+ * Denotes a Bond enumeration type.
```
chadwalters: "Enumeration": copy-paste?

java/core/src/main/java/com/microsoft/bond/Bonded.java
```diff
@@ -19,7 +19,8 @@ public Bonded(T instance) {
 
     @Override
     public void serialize(ProtocolWriter writer) throws IOException {
-        value.serialize(writer);
+        //value.serialize(writer);
```
chadwalters: Mark with TODO to indicate that implementation will come?

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -0,0 +1,514 @@
...
+    // Registry of enum types; entries are added by static class initializers
+    private static final ConcurrentHashMap<
+            Class<? extends BondEnum>,
+            EnumBondType<? extends BondEnum>> enumTypeRegistry =
+            new ConcurrentHashMap<
+                    Class<? extends BondEnum>,
+                    EnumBondType<? extends BondEnum>>();
+
+    // precomputed hash code helps to speed up object retrieval from the type cache
+    private final int precomputedHashCode;
```
tstein: I'm surprised to see us explicitly dealing with hash codes in many places in this PR. How much of that is optimization, and how much is semantically important? If any of it is semantically important, why was it necessary?

java/core/src/main/java/com/microsoft/bond/Bonded.java
tstein: Simon: If you're reading this and remembering me objecting to TODOs in source, I don't particularly mind when the TODO is necessarily going to happen in the immediate future.

That definitely applies here. I think the criterion for merging this back in is that it's able to codegen, serialize, and deserialize (at least its own output from) the compat types again, and that will require us to implement this.

java/core/src/main/java/com/microsoft/bond/BondEnum.java
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package com.microsoft.bond;
+
+/**
+ * Denotes a Bond enumeration type.
+ * All generated Bond enum classes implement this interface.
+ */
+public interface BondEnum {
```
sigansky: OK

---
In reply to: [122853157](https://github.com/Microsoft/bond/pull/502#discussion_r122853157) [](ancestors = 122853157)

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -0,0 +1,501 @@
...
+ * schema inference, initialization, serialization and deserialization. A Bond type cane be a primitive,
+ * a struct, a specialization of a generic type (including containers, nullable, or generic Bond structs).
+ * @param <T> the class of the value
+ */
+public abstract class BondType<T> {
+
+    // The global type cache
+    static final TypeCache typeCache = new TypeCache();
+
+    // Registry of struct type resolvers; entries are added by static class initializers
```
sigansky: Yes, this is added in the next iteration.

---
In reply to: [123089140](https://github.com/Microsoft/bond/pull/502#discussion_r123089140) [](ancestors = 123089140)

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -0,0 +1,501 @@
...
+    private static final ConcurrentHashMap<
+            Class<? extends BondSerializable>,
+            StructBondTypeResolver<? extends BondSerializable>> structTypeResolverRegistry =
+            new ConcurrentHashMap<
+                    Class<? extends BondSerializable>,
+                    StructBondTypeResolver<? extends BondSerializable>>();
+
+    // Registry of enum types; entries are added by static class initializers
```
sigansky: See above.

---
In reply to: [123089182](https://github.com/Microsoft/bond/pull/502#discussion_r123089182) [](ancestors = 123089182)

java/core/src/main/java/com/microsoft/bond/Throw.java
```diff
@@ -0,0 +1,180 @@
...
+ * Some message details (e.g. list indices, map keys) may be added to
+ * exception message by certain specialized implementations of Bond type.
+ */
+final class Throw {
+
+    // invariant locale
+    private static final Locale LOCALE = Locale.ROOT;
+
+    // general entry point to raise a data exception on serialization or deserialization
+    static void raiseInvalidDataError(
```
sigansky: Not sure why it's not "Javaism". Is there a "Javaism" authority?

---
In reply to: [123379014](https://github.com/Microsoft/bond/pull/502#discussion_r123379014) [](ancestors = 123379014)

java/core/src/main/java/com/microsoft/bond/Throw.java
```diff
@@ -0,0 +1,180 @@
...
+            Exception cause,
+            String format,
+            Object... args) throws InvalidBondDataException {
+        String message = String.format(LOCALE, format, args);
```
sigansky: To be consistent. These strings may reach end user if there is an exception. It's a good practice to keep invariant (root) locale for exception text.

---
In reply to: [123379360](https://github.com/Microsoft/bond/pull/502#discussion_r123379360) [](ancestors = 123379360)

java/core/src/main/java/com/microsoft/bond/Throw.java
```diff
@@ -0,0 +1,180 @@
...
+        throw new InvalidBondDataException(message, cause);
+    }
+
+    // general entry point to raise a data exception on value comparison
+    static void raiseIllegalArgumentException(
```
sigansky: Removed this method, it was remnant of older functionality I decided to exclude for now.

---
In reply to: [123380136](https://github.com/Microsoft/bond/pull/502#discussion_r123380136) [](ancestors = 123380136)

java/core/src/main/java/com/microsoft/bond/Bonded.java
sigansky: See Chad's comment earlier. Added a TODO.

---
In reply to: [123404614](https://github.com/Microsoft/bond/pull/502#discussion_r123404614) [](ancestors = 123404614)

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -0,0 +1,514 @@
...
+    public abstract Class<T> getValueClass();
+
+    /**
+     * Gets the {@link Class} instance for the Java primitive values (non-objects) of this type
+     * or null if a Java primitive type does not exist. This method complements the
+     * {@link BondType#getValueClass()} method which returns the classes for object instances.
+     *
+     * @return the class object for a primitive type or null if a primitive type doesn't exists
+     */
+    public abstract Class<T> getPrimitiveValueClass();
```
sigansky: Added comments.

---
In reply to: [123091392](https://github.com/Microsoft/bond/pull/502#discussion_r123091392) [](ancestors = 123091392)

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -0,0 +1,514 @@
...
+     * returns an initialized value equivalent to invoking the default constructor. For nullable
+     * values this method returns null.
+     *
+     * @return the default initialized value of this type
+     */
+    protected abstract T newDefaultValue();
+
+    /**
+     * Serializes a value of this type into a protocol writer.
+     * This method is not recommended for Java primitive (non-object) types. To serialize primitive
```
sigansky: There is no "non-object" type; every type has to know how to deal with objects. Certain primitive types (e.g. Int32) also need to know how to deal with primitives. The point of this comment is to avoid unnecessary boxing/unboxing every time.

---
In reply to: [123092975](https://github.com/Microsoft/bond/pull/502#discussion_r123092975) [](ancestors = 123092975)

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -0,0 +1,514 @@
...
+     * @return the deserialized value
+     * @throws IOException if an I/O error occurred
+     */
+    protected abstract T deserializeField(
+            TaggedDeserializationContext context,
+            StructBondType.StructField<T> field) throws IOException;
+
+    /**
+     * Deserializes a struct field of this type with "nothing" as the default value from a tagged protocol reader,
+     * excluding field metadata which is assumed to be already deserialized earlier and whose value is available
```
sigansky: It never returns null as is clear from the code. Its purpose is to deserialize a value and then wrap it by "something". I don't see anything strange here.

---
In reply to: [123094559](https://github.com/Microsoft/bond/pull/502#discussion_r123094559) [](ancestors = 123094559)

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -0,0 +1,514 @@
...
+    }
+
+    // helper for subclasses to compute hash code
+    static int multiplyAndShift(int value, int n) {
+        return (value * n) ^ (value >>> (32 - n));
+    }
+
+    // implemented by subclasses to test type decriptors for equality
+    // (the argument is not null and can be safely cast to the actual subclass)
+    abstract boolean equalsInternal(BondType<?> obj);
```
sigansky: Sure, although it doesn't really help with anything. I added this.

---
In reply to: [123094981](https://github.com/Microsoft/bond/pull/502#discussion_r123094981) [](ancestors = 123094981)

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -0,0 +1,514 @@
...
+
+    @Override
+    public final String toString() {
+        return this.getFullName();
```
sigansky: It returns full IDL name.

---
In reply to: [123095383](https://github.com/Microsoft/bond/pull/502#discussion_r123095383) [](ancestors = 123095383)

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -0,0 +1,514 @@
...
+
+        StructDefOrdinalTuple(StructDef structDef, int ordinal) {
+            this.structDef = structDef;
+            this.ordinal = ordinal;
+        }
+    }
+
+    // a helper to create schema; initializes a new type def instance and
+    // maintains a hash map of all distinct struct defs discovered so far
+    abstract TypeDef createSchemaTypeDef(HashMap<StructBondType<?>, StructDefOrdinalTuple> structDefMap);
```
sigansky: BondType can only be used as a key in hash map. Also this is a private method, so not sure of the concern.

---
In reply to: [123095878](https://github.com/Microsoft/bond/pull/502#discussion_r123095878) [](ancestors = 123095878)

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -0,0 +1,514 @@
...
+
+    // validates non-nullable value
+    final void verifyNonNullableValueIsNotSetToNull(
```
sigansky: The name is as descriptive as it can get.

---
In reply to: [123098407](https://github.com/Microsoft/bond/pull/502#discussion_r123098407) [](ancestors = 123098407)

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -0,0 +1,514 @@
...
+            BondType<?> that = (BondType<?>) obj;
+            if (this.precomputedHashCode != that.precomputedHashCode) {
+                return false;
+            } else {
+                return this.equalsInternal(that);
+            }
+        }
+    }
+
+    // helper for subclasses to compute hash code
```
sigansky: Private methods.

---
In reply to: [123096275](https://github.com/Microsoft/bond/pull/502#discussion_r123096275) [](ancestors = 123096275)

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -0,0 +1,514 @@
...
+    }
+
+    /**
+     * Returns a cached type descriptor that is equal to the argument, or caches the argument otherwise.
+     *
+     * @param type the type descriptor
+     * @param <T>  the Bond value class
+     * @return a cached type descriptor
+     */
+    protected static <T> BondType<T> getCachedType(BondType<T> type) {
```
sigansky: I have redesigned this a bit, and simplified type cache.

---
In reply to: [123102269](https://github.com/Microsoft/bond/pull/502#discussion_r123102269) [](ancestors = 123102269)

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -0,0 +1,514 @@
...
+        return BondType.typeCache.get(type);
+    }
+
+    // Implements type cache
+    static final class TypeCache {
+
+        // maps a bond type descriptor to itself; used as the primary cache of bond type descriptors
```
sigansky: See next iteration, with more comments and redesign of the type cache.

---
In reply to: [123102712](https://github.com/Microsoft/bond/pull/502#discussion_r123102712) [](ancestors = 123102712)

java/core/src/main/java/com/microsoft/bond/BondTypes.java
```diff
@@ -0,0 +1,102 @@
...
+     */
+    public static final WStringBondType WSTRING = WStringBondType.INSTANCE;
+
+    /**
+     * A singleton type descriptor for the Bond "blob" data type.
+     */
+    public static final BlobBondType BLOB = BlobBondType.INSTANCE;
+
+    static {
+        // cache primitive types for consistency
```
sigansky: Added comment.

---
In reply to: [123103914](https://github.com/Microsoft/bond/pull/502#discussion_r123103914) [](ancestors = 123103914)

java/core/src/main/java/com/microsoft/bond/BondedBondType.java
```diff
@@ -0,0 +1,149 @@
...
+public final class BondedBondType<TStruct extends BondSerializable> extends BondType<Bonded<TStruct>> {
+
+    /**
+     * The name of the type as it appears in Bond schemas.
+     */
+    public static final String TYPE_NAME = "bonded";
+
+    private final StructBondType<TStruct> valueType;
+
+    // restrict instantiation to the current package
```
sigansky: Comment means it is intentionally made so.

---
In reply to: [123108731](https://github.com/Microsoft/bond/pull/502#discussion_r123108731) [](ancestors = 123108731)

java/core/src/main/java/com/microsoft/bond/BondedBondType.java
```diff
@@ -0,0 +1,149 @@
...
+    BondedBondType(StructBondType<TStruct> valueType) {
+        super(-valueType.hashCode());
```
sigansky: Obviously you can't have unique hash code, but as distributed as possible for better perf. I'll keep it as it for now and we may revisit hash code computations later as part of perf tuning.

---
In reply to: [123108950](https://github.com/Microsoft/bond/pull/502#discussion_r123108950) [](ancestors = 123108950)

java/core/src/main/java/com/microsoft/bond/BondedBondType.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    @Override
+    public final boolean isNullableType() {
+        return false;
+    }
+
+    @Override
+    public final boolean isGenericType() {
+        return true;
```
sigansky: IsGenericType means what literally "generic" means, and bonded<> is a generic type (although native to bond) as well as list<>, map<,> and Box<>. Please refer to the API documentation.

---
In reply to: [123110230](https://github.com/Microsoft/bond/pull/502#discussion_r123110230) [](ancestors = 123110230)

java/core/src/main/java/com/microsoft/bond/BondedBondType.java
```diff
@@ -0,0 +1,149 @@
...
+        // TODO: complete deserialization story for bonded
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    protected final void serializeField(
+            SerializationContext context,
+            Bonded<TStruct> value,
+            StructBondType.StructField<Bonded<TStruct>> field) throws IOException {
+        // struct fields are never omitted
```
sigansky: bonded applies only to structs, so the current statement is more general. Updated comment.

---
In reply to: [123113105](https://github.com/Microsoft/bond/pull/502#discussion_r123113105) [](ancestors = 123113105)

java/core/src/main/java/com/microsoft/bond/BondedBondType.java
```diff
@@ -0,0 +1,149 @@
...
+        context.writer.writeFieldBegin(BondDataType.BT_STRUCT, field.getId(), field);
+        try {
+            this.serializeValue(context, value);
+        } catch (InvalidBondDataException e) {
+            // throws
```
sigansky: It indicates termination of the code that compiler doesn't catch, so it makes it clear that the code below always throws.

---
In reply to: [123113269](https://github.com/Microsoft/bond/pull/502#discussion_r123113269) [](ancestors = 123113269)

java/core/src/main/java/com/microsoft/bond/PrimitiveBondType.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    public final BondType<?>[] getGenericTypeArguments() {
+        return null;
+    }
+
+    @Override
+    final boolean equalsInternal(BondType<?> obj) {
+        // the caller makes sure that the class of the argument is the same as the class of this object,
+        // and since primitive types are singletons all instance are considered equal
+        return true;
```
sigansky: Please go back to the public equals method that calls this private method and walk through what it does and/or read the comments. Two instances of the same primitive Bond Type are considered equal, as they represent the same Bond type.

---
In reply to: [123115913](https://github.com/Microsoft/bond/pull/502#discussion_r123115913) [](ancestors = 123115913)

java/core/src/main/java/com/microsoft/bond/BondSerializable.java
sigansky: Yes, Chad already pointed this out. It was a copy-paste.

---
In reply to: [123116280](https://github.com/Microsoft/bond/pull/502#discussion_r123116280) [](ancestors = 123116280)

java/core/src/main/java/com/microsoft/bond/Serializer.java
```diff
@@ -0,0 +1,31 @@
...
+import com.microsoft.bond.helpers.ArgumentHelper;
+import com.microsoft.bond.protocol.ProtocolWriter;
+
+import java.io.IOException;
+
+/**
+ * Resposible for serialization of objects of a particular struct.
+ * @param <TStruct> the class of the serialized struct value
+ */
+public final class Serializer<TStruct extends BondSerializable> {
```
sigansky: Yes, as we have discussed and agreed on the team meeting.

---
In reply to: [123371025](https://github.com/Microsoft/bond/pull/502#discussion_r123371025) [](ancestors = 123371025)

java/core/src/main/java/com/microsoft/bond/Deserializer.java
```diff
@@ -0,0 +1,41 @@
...
+public final class Deserializer<TStruct extends BondSerializable> {
+
+    private final StructBondType<TStruct> structType;
+
+    /**
+     * Initializes a new deserializer.
+     *
+     * @param structType the type descriptor of the deserialized struct
+     */
+    public Deserializer(StructBondType<TStruct> structType) {
```
sigansky: Yes. How else would you do it, given that you need to know what class the data is being deserialized into.

---
In reply to: [123371871](https://github.com/Microsoft/bond/pull/502#discussion_r123371871) [](ancestors = 123371871)

java/core/src/main/java/com/microsoft/bond/StructBondTypeResolver.java
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package com.microsoft.bond;
+
+/**
+ * Responsible for resolving generic Bond struct types.
```
sigansky: Updating comments.

---
In reply to: [123373480](https://github.com/Microsoft/bond/pull/502#discussion_r123373480) [](ancestors = 123373480)

java/core/src/main/java/com/microsoft/bond/StructBondTypeResolver.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Resolves generic type arguments and creates and initializes new instance of type descriptor.
+     * Generated sublasses add a public method to resolve and initialized the type descriptor.
+     *
+     * @param genericTypeArguments generic type arguments
+     * @return new unitialized type descriptor instance
+     */
+    protected final StructBondType<TStruct> resolveAndInitialize(BondType<?>... genericTypeArguments) {
+        StructBondType<TStruct> type = this.resolveUninitialized(genericTypeArguments);
+        ensureInitialized(type);
```
sigansky: I am pretty sure it's fine as is.

---
In reply to: [123376279](https://github.com/Microsoft/bond/pull/502#discussion_r123376279) [](ancestors = 123376279)

java/core/src/main/java/com/microsoft/bond/StructBondTypeResolver.java
```diff
@@ -0,0 +1,77 @@
...
+     * Resolves generic type arguments and caches the result in the type cache.
+     *
+     * @param resolver             type resolver
+     * @param genericTypeArguments generic type arguments
+     * @param <TStruct>            the class of the struct value
+     * @return a cached uninitialized type descriptor
+     */
+    protected static <TStruct extends BondSerializable> StructBondType<TStruct> resolveUninitializedWithCaching(
+            StructBondTypeResolver<TStruct> resolver, BondType<?>... genericTypeArguments) {
+        return StructBondType.resolveUninitializedWithCaching(resolver, genericTypeArguments);
```
sigansky: I am redesigning type caching, see next iteration.

---
In reply to: [123377588](https://github.com/Microsoft/bond/pull/502#discussion_r123377588) [](ancestors = 123377588)

java/core/src/main/java/com/microsoft/bond/StructBondTypeResolver.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /**
+     * Returns a cached type descriptor that is equal to the argument, or caches the argument otherwise.
+     *
+     * @param type the type descriptor
+     * @param <T>  the Bond value class
+     * @return a cached type descriptor
+     */
+    protected static <T> BondType<T> getCachedType(BondType<T> type) {
```
sigansky: I am redesigning type caching, see next iteration.

To answer: I don't want to make any of these public and they need to be accessed by generated implementation of BondType and BondTypeResolver. This protected method makes it available to the latter.

---
In reply to: [123377776](https://github.com/Microsoft/bond/pull/502#discussion_r123377776) [](ancestors = 123377776)

java/core/src/main/java/com/microsoft/bond/StructBondType.java
```diff
@@ -0,0 +1,1359 @@
...
+import com.microsoft.bond.protocol.*;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Partially implements the {@link BondType} contract for struct data types.
+ * Leaves some implementation details to subclasses specific to the a particular
```
sigansky: I changed "some" to "the rest". If you want to know what they are, look at the abstract methods.

---
In reply to: [123382905](https://github.com/Microsoft/bond/pull/502#discussion_r123382905) [](ancestors = 123382905)

java/core/src/main/java/com/microsoft/bond/protocol/StructMetadata.java
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package com.microsoft.bond.protocol;
+
+/**
+ * A metadata describing a struct, used by {@link ProtocolWriter} for certain protocol types.
+ */
+public interface StructMetadata {
```
sigansky: We will take another look at this when designing untagged protocols.

---
In reply to: [123385305](https://github.com/Microsoft/bond/pull/502#discussion_r123385305) [](ancestors = 123385305)

java/core/src/main/java/com/microsoft/bond/StructBondType.java
```diff
@@ -0,0 +1,1359 @@
...
+        this.verifyNonNullableValueIsNotSetToNull(value);
+        context.writer.writeStructBegin(this);
+        if (this.baseStructType != null) {
+            this.baseStructType.serializeValueHelperForBase(context, value);
+        }
+        this.serializeStructFields(context, value);
+        context.writer.writeStructEnd();
+    }
+
+    // recursive helper
```
sigansky: This is a private method. Also, you have figured out what it does, so I really think comments are redundant here.

---
In reply to: [123397110](https://github.com/Microsoft/bond/pull/502#discussion_r123397110) [](ancestors = 123397110)

java/core/src/main/java/com/microsoft/bond/StructBondType.java
```diff
@@ -0,0 +1,1359 @@
...
+    private void serializeValueHelperForBase(
```
sigansky: Sure, renamed.

---
In reply to: [123397272](https://github.com/Microsoft/bond/pull/502#discussion_r123397272) [](ancestors = 123397272)

java/core/src/main/java/com/microsoft/bond/StructBondType.java
```diff
@@ -0,0 +1,1359 @@
...
+        } catch (InvalidBondDataException e) {
+            // throws
+            Throw.raiseStructFieldSerializationError(true, field, e, null);
+        }
+        return value;
+    }
+
+    @Override
+    final boolean equalsInternal(BondType<?> obj) {
+        // the caller makes sure that the class of the argument is the same as the class of this object
```
sigansky: It's package-private and final, so should be safe. The intent is to have public equals method do all repetitive tasks in the base, and just let a subtype deal with equality of non-null objects whose type is verified.

---
In reply to: [123397450](https://github.com/Microsoft/bond/pull/502#discussion_r123397450) [](ancestors = 123397450)

java/core/src/main/java/com/microsoft/bond/StructBondType.java
```diff
@@ -0,0 +1,1359 @@
...
+        StructBondType that = (StructBondType) obj;
+        if (this.genericTypeSpecialization != null) {
+            return this.genericTypeSpecialization.equals(that.genericTypeSpecialization);
+        } else {
+            return that.genericTypeSpecialization == null;
+        }
+    }
+
+    // private top-level serialization entry point
```
sigansky: It is package private, which is effectively private to Bond core. It's not visible in public API.

---
In reply to: [123399250](https://github.com/Microsoft/bond/pull/502#discussion_r123399250) [](ancestors = 123399250)

java/core/src/main/java/com/microsoft/bond/StructBondType.java
```diff
@@ -0,0 +1,1359 @@
...
+            StructBondTypeResolver<TStruct> resolver, BondType<?>... genericTypeArguments) {
+        BondType<?>[] cachedGenericTypeArguments = new BondType[genericTypeArguments.length];
+        for (int i = 0; i < genericTypeArguments.length; ++i) {
+            cachedGenericTypeArguments[i] = typeCache.get(genericTypeArguments[i]);
+        }
+        StructBondType<TStruct> type = resolver.resolveUninitialized(cachedGenericTypeArguments);
+        return (StructBondType<TStruct>) BondType.typeCache.get(type);
+    }
+
+    private void initializeSchemaStructDef(
```
sigansky: There's a public method "createSchemaDef" in StructBondType. It's not called yet, and we'll take a closer look when designing untagged protocols.

---
In reply to: [123399536](https://github.com/Microsoft/bond/pull/502#discussion_r123399536) [](ancestors = 123399536)

java/core/src/main/java/com/microsoft/bond/BondType.java
sigansky: It's a useful optimization since we heavily rely on type caching. Precomputed hash code also helps object equality (if objects hash codes don't match then no need to do any deep object comparison).

---
In reply to: [123400869](https://github.com/Microsoft/bond/pull/502#discussion_r123400869) [](ancestors = 123400869)

java/core/src/main/java/com/microsoft/bond/BlobBondType.java
```diff
@@ -0,0 +1,124 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package com.microsoft.bond;
+
+import java.io.IOException;
+
+/**
+ * Implements the {@link BondType} contract for the Bond "blob" data type.
+ */
+public final class BlobBondType extends PrimitiveBondType<byte[]> {
```
sigansky: Yes, I realized that yesterday and changed it already.

---
In reply to: [123117294](https://github.com/Microsoft/bond/pull/502#discussion_r123117294) [](ancestors = 123117294)

java/core/src/main/java/com/microsoft/bond/BondedBondType.java
chadwalters: Does this mean "Java generic" or "Bond generic"? Because in Bond, we don't refer bonded<>, list<>, map<> as generic.

sigansky: It means Java generic, including both Bond native generics and user-defined generics.

---
In reply to: [123802589](https://github.com/Microsoft/bond/pull/502#discussion_r123802589) [](ancestors = 123802589)

java/core/src/main/java/com/microsoft/bond/BondType.java
tstein: Okay. To be clear, I would like this comment to say which static class initializers it's referring to. Same thing below.

java/core/src/main/java/com/microsoft/bond/BondType.java
tstein: Took another read, but I see where I was confused now. A `nullable<int32>` should result in a call to `Int32BondType.serializeValue(Integer)`; an `int32` should go through `Int32BondType.serializePrimitiveValue(int)`.

I'm fine with the structure, but I think this needs a clearer comment about the intent. How about something like this?
```
Serializes a value of this type into a protocol writer.

For non-nullable Bond types that map to Java value types, the corresponding
BondType implementation provides a serializeValue() that avoids boxing. For
nullable fields of those types, use this method.
```

java/core/src/main/java/com/microsoft/bond/BondType.java
tstein: "with nothing as the default value" read to me like a description of this code's behavior, not the IDL feature that triggers it. Let's come up with a more concise way to refer to that feature offline and start using it here.

The "not recommended" here is similar to the one above. I'd like something similar to the comment I suggested above here, too.

java/core/src/main/java/com/microsoft/bond/BondType.java
tstein: I understand, but it doesn't contain anything that would point a reader back to the `BondType` hierarchy. I would like it to do so.

java/core/src/main/java/com/microsoft/bond/Deserializer.java
tstein: This is fine for this PR. Let's revisit it later.

java/core/src/main/java/com/microsoft/bond/BondType.java
sigansky: nullable<int32> will go through NullableBondType.serializeValue(Integer) which will then (maybe) delegate to Int32BondType.serializeValue(Integer). There is no other place to delegate to since from NullableBondType is generic.

---
In reply to: [123839812](https://github.com/Microsoft/bond/pull/502#discussion_r123839812) [](ancestors = 123839812)

java/core/src/main/java/com/microsoft/bond/BondType.java
sigansky: Per our discussion, keeping this as is. It's the IDL type name that is relevant.

---
In reply to: [123843410](https://github.com/Microsoft/bond/pull/502#discussion_r123843410) [](ancestors = 123843410)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -40,9 +40,11 @@ package #{javaPackage};
 #{Java.generatedClassAnnotations}
 public class #{declName}#{params}#{maybe interface baseClass structBase} {
 
+    public static final com.microsoft.bond.StructBondType<#{declName}> BOND_TYPE = null;
```
sigansky: Yes, it needs to be public for static access.

---
In reply to: [124081480](https://github.com/Microsoft/bond/pull/502#discussion_r124081480) [](ancestors = 124081480)

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -12,38 +12,52 @@
...
-                    Class<? extends BondEnum>,
-                    EnumBondType<? extends BondEnum>>();
-
-    // precomputed hash code helps to speed up object retrieval from the type cache
-    private final int precomputedHashCode;
-
-    // restrict subclasses to the current package
-    BondType(int precomputedHashCode) {
-        this.precomputedHashCode = precomputedHashCode;
+    // The global type cache, which maps a Bond type descriptor to itself and is used to cache type descriptor
```
sigansky: I don't understand your comment. The purpose of this data structure is to retrieve a value given a key, which is what Map offers. Set API doesn't have that.

---
In reply to: [124083933](https://github.com/Microsoft/bond/pull/502#discussion_r124083933) [](ancestors = 124083933)

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -103,73 +121,83 @@ public final String getFullName() {
...
+     * @return an array containing the descriptors of the generic type arguments or null if not a generic type
      */
     public abstract BondType<?>[] getGenericTypeArguments();
 
     /**
      * Returns the default value of this type as a shared instance for immutable primitive types
      * or a new instance for all other types. For non-nullable structs and containers this method
-     * returns an initialized value equivalent to invoking the default constructor. For nullable
-     * values this method returns null.
+     * returns an initialized value equivalent to invoking the public constructor. For nullable
```
sigansky: Can you clarify what nullary means? I updated the comment to specify which constructor (it's not always the parameterless one).

---
In reply to: [124082288](https://github.com/Microsoft/bond/pull/502#discussion_r124082288) [](ancestors = 124082288)

tstein: Nullary means parameterless. If that's not accurate, thanks for adding the clarification to the comment.

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -232,44 +264,27 @@ protected abstract T deserializeField(
...
-        return (value * n) ^ (value >>> (32 - n));
+    /**
+     * Helper for container subclasses to compute hash code as a value multipled by
+     * a number and then having its bits shifted to the right by a certain number.
+     * @param value a value
+     * @param multiplier multiplier
+     * @param bitsShifted number of bits shifted
+     * @return result
+     */
+    static int multiplyAndShiftForHashCodeComputation(int value, int multiplier, int bitsShifted) {
```
sigansky: It's a function to help with hashcode computation. The caller may want to specify those for different hash behavior.

---
In reply to: [124084837](https://github.com/Microsoft/bond/pull/502#discussion_r124084837) [](ancestors = 124084837)

java/core/src/main/java/com/microsoft/bond/BondedBondType.java
```diff
@@ -22,7 +22,7 @@
 
     BondedBondType(StructBondType<TStruct> valueType) {
         this.valueType = valueType;
-        this.precomputedHashCode = -valueType.hashCode();
+        this.precomputedHashCode =HashCode.computeHashCodeForBondedContainer(valueType);
```
tstein: s/=Hash/= Hash/

Issue #414, opened 2017-04-12
Switch to bond.snk for all strong names in .NET


chwarr:
Currently, when you build Bond locally, it uses [bond.snk](https://github.com/Microsoft/bond/blob/581d88632e0ad8b2fc87f5674273f613e78af752/cs/build/internal/bond.snk) to sign the .NET assemblies. However, when an official release of Bond is produced, a Microsoft-internal key is used to sign the assemblies. This means that locally produced assemblies and Microsoft-produced assemblies have different strong names. The current guidance for Microsoft open source projects is to use one key that is accessible to everyone for strong name signing, and to Authenticode sign official Microsoft releases for integrity. (Official release of Bond are already Authenticode.)

This will be a breaking change, as it will change the strong name of the assemblies.

PR #505, opened 2017-06-20
[c#] Switch to bond.snk for all strong-name signing

chwarr:
Previously, when when an official release of Bond was produced, a
Microsoft-internal key instead of bond.snk was used to strong-name sign
the assemblies. This means that locally produced assemblies and
Microsoft-produced assemblies have different strong names, making it
difficult to consume non-official releases in applications that require
strong-name signed assemblies.

This change switches to using bond.snk for all strong-name signing.

Official Microsoft release will still be Authenticode signed for
integrity. (Official release of Bond are already Authenticode signed.)

This is a breaking change, as it will change the strong name of the
assemblies.

Resolves https://github.com/Microsoft/bond/issues/414


chwarr:
Yeah, NuGetTestReflection.cs should be removed when we next rev the version of that test. I had to remove it during my testing, as Bond.Reflection [now makes its internals visible to Bond.NuGetTest](https://github.com/Microsoft/bond/pull/505/files#diff-301206ac8cab0dde5ee0de4d81b1e5cdR21). Failure to remove it will break the build, so we'll catch this during the next release.

chwarr:
Build failures are being tracked in issue https://github.com/Microsoft/bond/issues/508

chadwalters:
Travis issues are independent. Ready to go into master.

Issue #508, opened 2017-06-23
Travis CI builds have stopped working: cabal and ghc versions appear not to be compatible


chwarr:
Travis CI builds have stopped working. gbc cannot be successfully built. From [a recent PR build that only changed some C# code](https://travis-ci.org/Microsoft/bond/jobs/245990932#L2614):

> on the commandline: warning:
>     -this-package-key is deprecated: Use -this-unit-id instead
> In-place registering bond-0.9.0.0...
> cabal: '/opt/ghc/bin/ghc-pkg' exited with an error:
> bond-0.9.0.0: Warning: haddock-interfaces:
> /home/travis/build/Microsoft/bond/build/compiler/doc/html/bond/bond.haddock
> doesn't exist or isn't a file
> bond-0.9.0.0: Warning: haddock-html:
> /home/travis/build/Microsoft/bond/build/compiler/doc/html/bond doesn't exist
> or isn't a directory
> bond-0.9.0.0: installed package info from too old version of Cabal (key field
> does not match id field)
> CMake Error at cabal_build.cmake:60 (message):

The Travis build is now using ghc 8.0.2 with cabal 1.22.9.0.

It may be that this combination [is not compatible](https://github.com/haskell/cabal/issues/3811).

lalo:
Non-issue after commit 52dbdb0ba882748b1d5857e7c60791c346a8e889

PR #509, opened 2017-06-23
Improve MSBuild codegen handling of gbc changes

chwarr:
Two commits here:

* teach the MSBuild targets to rebuild when gbc itself is changed
* improve MSBuild targets comments

Please rebase when merging.

### DIFF
--- a/cs/build/nuget/Common.targets
+++ b/cs/build/nuget/Common.targets
@@ -34,6 +34,8 @@
     </BondImportDirectory>
   </ItemDefinitionGroup>
 
+  <!-- Let Visual Studio know about the BondCodegen item so it can display
+       it in the UI as a potential action for files. -->
   <ItemGroup>
     <AvailableItemName Include="BondCodegen" />
   </ItemGroup>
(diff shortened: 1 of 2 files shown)


cs/build/nuget/Common.targets
```diff
@@ -34,22 +34,44 @@
...
-    *    must rebuild all .bond files when any of them change.  We force that by putting a non-transform output
-    *    (bondcodegen.done) in the list so IFF any input changes, then all inputs rebuild since MSBuild can't
-    *    know what should create the unmapped output item and so can't be selective.
-    *    NOTE - this still won't catch changes to files outside declared BondCodegen elements.
-  -->
+  <!-- Attempt to find gbc in a platform-agnostic way. -->
+  <PropertyGroup>
+    <_BondExe>gbc</_BondExe>
+    <_BondExe Condition="Exists('$(BOND_COMPILER_PATH)\gbc.exe')">$(BOND_COMPILER_PATH)\gbc.exe</_BondExe>
+    <_BondExe Condition="Exists('$(BOND_COMPILER_PATH)\gbc')">$(BOND_COMPILER_PATH)\gbc</_BondExe>
```
tstein: Just for the hell of it, what happens if both of these files exist? I haven't done this yet, but I think I could plausibly cause that situation if I shared a tree between a Windows-world build and and WSL-world one.

chwarr: It would use the second one. Properties are evaluated top-to-bottom, and the last one that evaluates wins.

chwarr:
CI failure is due to https://github.com/Microsoft/bond/issues/508, which @lalo is working on "fixing" with #512.

chwarr:
Merged in commit 0bce7292.

PR #513, opened 2017-06-28
Remove support for .NET 4.0

chadwalters:
Also, remove the reflection shim designed to cover over API differences in .NET 4.0 and 4.5.


chwarr:
The .NET Core build failed. Our AppVeyor config doesn't actually dump the errors to the log correctly, so you'll need to re-run that part of the build locally to get the errors. Sorry about the inconvenience. (This will get better when we can migrate to MSBuild 15 for all our build stuff instead of having this split build system.)

chwarr:
The nuproj files also need to be updated to remove the references to net40, but aside from that, LGTM.

chadwalters:
.nuspec files updated. Also removed cruft in Config.cmake and all usages of the NET45 constant.

chwarr:
Merge into master in commit 31211e10.

PR #514, opened 2017-06-28
[c#] Stamp version info into Bond assemblies

chwarr:
Support for embedding assembly and file version information has been
added to the MSBuild and .NET Core builds.

When building Bond, if the environment variable BOND_CORE_VERSION (or,
for Bond Comm, BOND_COMM_VERSION) is set to a valid NuGet package
version string, the generated assemblies will have versions details
embedded in them as well.

* For developer builds without any version set, the version 0.0.0.1 will
  be used.
* For pre-release builds (versions of the form x.y.z-someTag), the
  version x.y.z.0 will be used.
* For release builds (version of the form x.y.z), the version x.y.z.100
  will be used.


chwarr:
The Mono build is failing. I'm debugging this now. It _appears_ that `$(MSBuildThisFileDirectory)` behaves differently under Mono's xbuild...

chwarr:
Worked around the Mono issue in the most recent commit. Please squash when merging now.

Issue #417, opened 2017-04-13
C# SimpleJsonWriter incorrectly serializes non-nullable field set to null


chwarr:
Given the following .bond file

```
namespace bug;

struct NonNullableStringField
{
  0: string sf;
}
```

The following C# code should fail to serialize as `sf` is `null`.

```
namespace bug
{
    using System;
    using System.IO;

    using Bond;
    using Bond.Protocols;

    internal static class Program
    {
        private static void Main()
        {
            var obj = new NonNullableStringField
            {
                sf = null
            };

            var ser = new Serializer<SimpleJsonWriter>(typeof (NonNullableStringField));

            var stream = new StringWriter();
            var jw = new SimpleJsonWriter(stream);

            ser.Serialize(obj, jw);
            jw.Flush();

            Console.WriteLine(stream.ToString());
        }
    }
}
```

Instead, the following is output.

```
{"sf":null}
```

Other types than string, like sub-struct fields, are also affected.

`SimpleXmlWriter` is also affected. `CompactBinaryWriter`, `FastBinaryWriter`, and `SimpleBinaryWriter` all fail with a `NullReferenceException`, as expected.

PR #515, opened 2017-06-28
[c#] JSON+XML: Throw on null, non-nullable strings.

tstein:
resolves https://github.com/Microsoft/bond/issues/417

The noise in `SimpleJsonWriter.cs` is because the two string functions were not near each other. I moved one so they could be neighbors.

### DIFF
--- a/cs/src/core/protocols/SimpleXmlWriter.cs
+++ b/cs/src/core/protocols/SimpleXmlWriter.cs
@@ -212,12 +212,32 @@ namespace Bond.Protocols
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public void WriteString(string value)
         {
+            // Other protocols depend on expressions such as value.Count to
+            // throw an NRE if we've been asked to serialize a non-nullable
+            // string field that is set to null. Implementations of
+            // System.Xml.XmlWriter may successfully serialize it, so we need
+            // to check and throw explicitly before that.
+            if (value == null)
+            {
+                throw new NullReferenceException(
+                   "Attempted to serialize a null string. This may indicate a non-nullable string field that was set to null.");
+            }
             writer.WriteValue(value);
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public void WriteWString(string value)
         {
+            // Other protocols depend on expressions such as value.Count to
+            // throw an NRE if we've been asked to serialize a non-nullable
+            // string field that is set to null. Implementations of
+            // System.Xml.XmlWriter may successfully serialize it, so we need
+            // to check and throw explicitly before that.
+            if (value == null)
+            {
+                throw new NullReferenceException(
+                   "Attempted to serialize a null string. This may indicate a non-nullable string field that was set to null.");
+            }
             writer.WriteValue(value);
         }
         #endregion
--- a/cs/test/core/Core.csproj
+++ b/cs/test/core/Core.csproj
@@ -34,6 +34,7 @@
     <Compile Include="AttributesTests.cs" />
     <Compile Include="BondClass.cs" />
     <Compile Include="BondedTests.cs" />
+    <Compile Include="JsonSerializationTests.cs" />
     <Compile Include="CloningTests.cs" />
     <Compile Include="CustomBondedTests.cs" />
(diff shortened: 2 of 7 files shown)


chadwalters:
Does the XML protocol have the same issue? If so, can you also address that?

tstein:
I'm surprised to say that it does - I understand why the JSON library we're using handles nulls, but not why its XML equivalent does. Adding a second commit to this PR to fix XML.

cs/src/core/protocols/SimpleXmlWriter.cs
```diff
@@ -260,6 +260,16 @@ public void WriteBool(bool value)
 #endif
         public void WriteString(string value)
         {
+            // Other protocols depend on expressions such as value.Count to
+            // throw an NRE if we've been asked to serialize a non-nullable
+            // string field that is set to null. Newtonsoft.Json will
```
chadwalters: Copy-paste mistake?

cs/src/core/protocols/SimpleXmlWriter.cs
```diff
@@ -268,6 +278,16 @@ public void WriteString(string value)
 #endif
         public void WriteWString(string value)
         {
+            // Other protocols depend on expressions such as value.Count to
+            // throw an NRE if we've been asked to serialize a non-nullable
+            // string field that is set to null. Newtonsoft.Json will
```
chadwalters: Copy-paste mistake?

cs/test/core/JsonSerializationTests.cs
```diff
@@ -10,7 +10,7 @@
     class JsonSerializationTests
     {
         [Test]
-        public void Null_NonNullableStrings()
+        public void JsonSerialization_Null_NonNullableStrings()
```
chadwalters: Minor: Move to JSON commit instead? Or should we merge to master as a single commit?

cs/src/core/protocols/SimpleXmlWriter.cs
tstein: Yeah, my bad. Forgot that I left in a reference to the JSON specifics.

cs/test/core/JsonSerializationTests.cs
tstein: Moved to JSON commit.

cs/test/core/Core.csproj
```diff
@@ -81,6 +82,7 @@
     <Compile Include="$(IntermediateOutputPath)\NamespaceConflict_types.cs" Condition="False" />
     <Compile Include="$(IntermediateOutputPath)\NamespaceConflictBond_types.cs" Condition="False" />
     <Compile Include="$(IntermediateOutputPath)\Bond File With Spaces_types.cs" Condition="False" />
+    <Compile Include="XmlSerializationTests.cs" />
```
chwarr: Minor: can you put this with the other XML file and the Json file in alpha order with the rest of the Compile items?

cs/test/core/JsonSerializationTests.cs
```diff
@@ -0,0 +1,26 @@
...
+    using System.IO;
+    using Bond;
+    using Bond.Protocols;
+    using NUnit.Framework;
+
+    [TestFixture]
+    class JsonSerializationTests
+    {
+        [Test]
+        public void JsonSerialization_Null_NonNullableStrings()
```
chwarr: `JsonSerialization_NullNonNullableStrings_Throws`?

cs/test/core/XmlSerializationTests.cs
```diff
@@ -0,0 +1,26 @@
...
+    using System.Xml;
+    using Bond;
+    using Bond.Protocols;
+    using NUnit.Framework;
+
+    [TestFixture]
+    class XmlSerializationTests
+    {
+        [Test]
+        public void XmlSerialization_Null_NonNullableStrings()
```
chwarr: Minor: `XmlSerialization_NullNonNullableStrings_Throws`?

chwarr:
Travis builds are failing. This appears to be a known issue in Travis:
https://github.com/travis-ci/travis-ci/issues/7987

Issue #497, opened 2017-06-13
issues with spaces in paths in cmake


tstein:
On my Linux desktop, I cloned `bond` into a directory called `bond space`. A single-threaded make fails with:
```
gbc: user error (Can't find import file bond/comm/comm.bond)
cpp/CMakeFiles/bond.dir/build.make:115: recipe for target '../cpp/generated/bond/comm/comm_reflection.h' failed
```
Previously, I did this by copying a dirty tree, and I got different errors about not finding C++ source files in the build directory. They had been copied into a new directory alongside my root with an escaped backslash in it. (i.e., I had a single dir containing `bond space` and `bond\\ space`.)

PR #516, opened 2017-06-29
[compat] Fix _popen calls to handle spaces.

tstein:
With this change, Windows C++ builds work in paths with spaces (e.g., `C:\\bond space\`). *nix C++ and C# have other problems.

contributes to https://github.com/Microsoft/bond/issues/497

### DIFF
--- a/cpp/test/compat/comm/commcompat.cpp
+++ b/cpp/test/compat/comm/commcompat.cpp
@@ -12,6 +12,8 @@
 #include <bond/core/cmdargs.h>
 #include <bond/stream/stdio_output_stream.h>
 
+#include <boost/algorithm/string/replace.hpp>
+
 #include "commcmd_arg_types.h"
 #include "commcmd_arg_reflection.h"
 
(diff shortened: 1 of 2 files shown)


cpp/test/compat/comm/commcompat.cpp
```diff
@@ -22,7 +24,9 @@ using namespace unittest::compat;
 FILE* popen_wrapper(const char* command, const char* mode)
 {
 #ifdef _WIN32
-    return _popen(command, mode);
+    std::string escaped_command = command;
+    boost::replace_all(escaped_command, " ", "\" \"");
```
chwarr: Am I reading this correctly: replace a single pace with the string `" "` (quote space quote)? That doesn't seem quite right. Say we want to execute the program `cool beans.exe` and pass it the file `how now.txt`.

    popen_wrapper("cool beans.exe how now.txt", "r");

I expect:

    "cool beans.exe" "how now.txt"

But this code looks like it would produce:

    cool" "beans.exe" "how" "now.txt

I think the quoting needs to happen one level above, in CMake, where it knows what the individual parts of the arguments are. By the time we're here, I think we've lost the distinction between what is the command and what is an argument.

chadwalters:
Should we close this pull request since this approach doesn't appear to be the correct one?

chwarr:
Closing for now due to string escaping problems. We may re-open if we have a different approach.

PR #517, opened 2017-06-29
Jvm generics: codegen

sigansky:


compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -37,23 +228,157 @@ package #{javaPackage};
...
+                initializeBaseAndFieldsBaseParam = structBaseDescriptorInitStructExpr java structBase
+                initializeBaseAndFieldsFieldsParam = if null structFields
+                    then mempty
+                    else [lt|, #{sepBy ", " structFieldReference structFields}|]
+                    where
+                        structFieldReference Field {..} = [lt|this.#{fieldName}|]
+                buildNewInstanceMethodBody = if null declParams
+                    then [lt|return new __StructBondTypeImpl();|]
+                    else [lt|return new __StructBondTypeImpl(new com.microsoft.bond.GenericTypeSpecialization(genericTypeArguments));|]
+                structBondTypeImplConstructorDecl = if null declParams
```
tstein: We haven't formalized this, but I understand `*Decl` to be a declaration or a declaration+definition - `public int i;` or `public int i = 0;`. This is just a name.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -37,23 +228,157 @@ package #{javaPackage};
...
+                fieldDescriptorFieldInit Field {..} = [lt|this.#{fieldName} = new #{structFieldDescriptorTypeName java fieldType}(#{initMethodParams});|]
+                        where
+                            initMethodParams = [lt|this#{fieldTypeParam}, #{fieldOrdinal}, "#{fieldName}", #{modifierConstantName fieldModifier}#{fieldDefaultValueParam}|]
+                                where
+                                    fieldTypeParam = if isGenericStructFieldDescriptor fieldType
+                                        then [lt|, #{structFieldDescriptorInitTypeExpr java fieldType}|]
+                                        else mempty
+                                    fieldDefaultValueParam = if not (isPrimitiveBondType fieldType) || (fieldDefault == Nothing)
+                                        then mempty
+                                        else [lt|, #{fieldDefaultValueInitParamExpr java fieldType fieldDefault}|]
```
tstein: We're eight levels indented here - at the very least, field descriptors look complex enough to deserve their own top-level function.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -37,23 +228,157 @@ package #{javaPackage};
...
+    #{doubleLineSep 1 publicFieldDecl structFields}
+
+    #{publicConstructorDecl}
+
+    @Override
+    public com.microsoft.bond.StructBondType<? extends com.microsoft.bond.BondSerializable> getBondType() {
+        return #{getBondTypeReturnValue};
+    }
 }|]
             where
```
tstein: I'm finding this big `where` to be extremely difficult to read. It looks like most of the expressions branch on genericity; I'm thinking that we should have two separate blocks to deal with those two, and maybe even two templates.

compiler/src/Language/Bond/Codegen/Java/Util.hs
```diff
@@ -4,110 +4,63 @@
...
-    BT_UserDefined _ _       -> "BT_STRUCT"
-    -- FIXME: Marker for unsupported types that compiles.
-    _                        -> "BT_UNAVAILABLE"
+-- returns the fully qualified name for a declaration
+qualifiedDeclaredTypeName :: MappingContext -> Declaration -> String
+qualifiedDeclaredTypeName java d@Struct {..}  = intercalate "." $ getDeclNamespace java d ++ [declName]
+qualifiedDeclaredTypeName java d@Enum {..}  = intercalate "." $ getDeclNamespace java d ++ [declName]
+qualifiedDeclaredTypeName java d@Forward {..}  = intercalate "." $ getDeclNamespace java d ++ [declName]
+qualifiedDeclaredTypeName java d@Alias {..}  = intercalate "." $ getDeclNamespace java d ++ [declName]
+qualifiedDeclaredTypeName java d@Service {..}  = intercalate "." $ getDeclNamespace java d ++ [declName]
```
tstein: Something of a higher-level question - does gbc actually emit services if no rpc mode is enabled? Seems to me like it shouldn't, and Java should error out if a service makes it into Java-specific code before Java supports grpc.

java/core/src/test/java/com/microsoft/bond/testmodel/schema.bond
```diff
@@ -7,6 +7,8 @@ struct C;
 struct D;
 struct E<U>;
 
+struct NotHere;
```
tstein: FYI, when you rebase your changes on master, you'll pick up the gradle-native codegen. This file will want to live somewhere under `java/core/src/test/bond`.

compiler/Compiler.lkshw
```diff
@@ -0,0 +1,14 @@
+Version of workspace file format:
```
sigansky: It was added by the IDE I use. Will remove.

---
In reply to: [124936962](https://github.com/Microsoft/bond/pull/517#discussion_r124936962) [](ancestors = 124936962)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -37,23 +228,157 @@ package #{javaPackage};
...
+    }
+
+    #{bondTypeStaticVariableDecl}
+
+    public static void initializeBondType() {
+        __StructBondTypeImpl.__StructBondTypeBuilderImpl.register();
+    }
+
+    static {
+        initializeBondType();
```
sigansky: No, this function needs to be public. Static block alone aren't sufficient (consider complex/circular class dependencies in multi-threaded environments). This function is called by the code (alt. we could move this to the constructor with a small per-instance overhead). 

In fact, we could remove the static block.

---
In reply to: [124944189](https://github.com/Microsoft/bond/pull/517#discussion_r124944189) [](ancestors = 124944189)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -10,25 +10,216 @@ module Language.Bond.Codegen.Java.Class_java
...
+|]
+    where
+        paramList = sepBy ", " paramName declParams
+        methodArg TypeParam {..} = [lt|com.microsoft.bond.BondType<#{paramName}> #{paramName}|]
+        checkArgument TypeParam {..} = [lt|com.microsoft.bond.helpers.ArgumentHelper.ensureNotNull(#{paramName}, "#{paramName}");|]
+
+
+-- builds text for anonymous implementation of the GenericBondTypeBuilder abstract class and assignment to the BOND_TYPE variable
+bondTypeStaticVariableDeclAsGenericBondTypeBuilder :: String -> [TypeParam] -> Text
+bondTypeStaticVariableDeclAsGenericBondTypeBuilder declName declParams = [lt|public static final GenericBondTypeBuilder BOND_TYPE = new GenericBondTypeBuilder() {
```
sigansky: GenericBondTypeBuilder is public and this is a private implementation of it. 

---
In reply to: [125136658](https://github.com/Microsoft/bond/pull/517#discussion_r125136658) [](ancestors = 125136658)

sigansky: Not sure what you mean by "casting in constructors", please clarify.

---
In reply to: [125371147](https://github.com/Microsoft/bond/pull/517#discussion_r125371147) [](ancestors = 125371147,125136658)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -10,25 +10,216 @@ module Language.Bond.Codegen.Java.Class_java
...
 -- field -> public field
 data JavaFieldMapping = JavaPublicFields deriving Eq
 
+
+-- returns the name of the struct field descriptor type (a protected nested class within StructBondType)
+structFieldDescriptorTypeName :: MappingContext -> Type -> Text
+structFieldDescriptorTypeName java type_ = typeName type_
+    where
+        typeName (BT_Maybe BT_Int8) = [lt|com.microsoft.bond.StructBondType.SomethingInt8StructField|]
+        typeName BT_Int8 = [lt|com.microsoft.bond.StructBondType.Int8StructField|]
```
sigansky: This way, all types are together. As we discussed this morning, let's get the codegen changes in and then address formatting of Haskell code later.

---
In reply to: [124937143](https://github.com/Microsoft/bond/pull/517#discussion_r124937143) [](ancestors = 124937143)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -10,25 +10,216 @@ module Language.Bond.Codegen.Java.Class_java
...
+        typeName BT_Float = [lt|com.microsoft.bond.StructBondType.FloatStructField|]
+        typeName (BT_Maybe BT_Double) = [lt|com.microsoft.bond.StructBondType.SomethingDoubleStructField|]
+        typeName BT_Double = [lt|com.microsoft.bond.StructBondType.DoubleStructField|]
+        typeName (BT_Maybe BT_Bool) = [lt|com.microsoft.bond.StructBondType.SomethingBoolStructField|]
+        typeName BT_Bool = [lt|com.microsoft.bond.StructBondType.BoolStructField|]
+        typeName (BT_Maybe BT_String) = [lt|com.microsoft.bond.StructBondType.SomethingStringStructField|]
+        typeName BT_String = [lt|com.microsoft.bond.StructBondType.StringStructField|]
+        typeName (BT_Maybe BT_WString) = [lt|com.microsoft.bond.StructBondType.SomethingWStringStructField|]
+        typeName BT_WString = [lt|com.microsoft.bond.StructBondType.WStringStructField|]
+        typeName (BT_Maybe (BT_UserDefined e@Enum {..} _)) = [lt|com.microsoft.bond.StructBondType.SomethingWStringStructField<#{qualifiedDeclaredTypeName java e}>|]
```
sigansky: Thanks.

---
In reply to: [124937565](https://github.com/Microsoft/bond/pull/517#discussion_r124937565) [](ancestors = 124937565)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -10,25 +10,216 @@ module Language.Bond.Codegen.Java.Class_java
...
+        typeName(BT_UserDefined e@Enum {..} _) = [lt|com.microsoft.bond.StructBondType.EnumStructField<#{qualifiedDeclaredTypeName java e}>|]
+        typeName (BT_Maybe t) = [lt|com.microsoft.bond.StructBondType.SomethingObjectStructField<#{(getTypeName java) t}>|]
+        typeName t = [lt|com.microsoft.bond.StructBondType.ObjectStructField<#{(getTypeName java) t}>|]
+
+
+-- returns whether a struct field descriptor type (a protected nested class within StructBondType) is generic and hence needs an explicit type parameter
+isGenericStructFieldDescriptor :: Type -> Bool
+isGenericStructFieldDescriptor (BT_Maybe t) = not (isPrimitiveNonEnumBondType t)
+isGenericStructFieldDescriptor t = not (isPrimitiveNonEnumBondType t)
```
sigansky: Yes, isPrimitiveNonEnumBondType returns False for (Maybe _).

---
In reply to: [124937933](https://github.com/Microsoft/bond/pull/517#discussion_r124937933) [](ancestors = 124937933)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -10,25 +10,216 @@ module Language.Bond.Codegen.Java.Class_java
...
+
+
+-- returns an type descriptor expression for a given struct (identified by a name and generic parameters), used in intiialization of struct field descriptors
+structFieldDescriptorInitStructExpr :: MappingContext -> String -> [Type] -> Text
+structFieldDescriptorInitStructExpr java qualifiedTypeName params = javaStructFieldDescriptorInitStructExpr
+    where
+        javaStructFieldDescriptorInitStructExpr = [lt|#{typeCastExpr}getStructType(#{qualifiedTypeName}.class#{paramExprList params})|]
```
sigansky: Yes, I chose the style of resolving this for 'java' mapping context.

---
In reply to: [124938304](https://github.com/Microsoft/bond/pull/517#discussion_r124938304) [](ancestors = 124938304)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -10,25 +10,216 @@ module Language.Bond.Codegen.Java.Class_java
...
+        javaStructFieldDescriptorInitTypeExpr (BT_Vector t) = [lt|vectorOf(#{javaStructFieldDescriptorInitTypeExpr t})|]
+        javaStructFieldDescriptorInitTypeExpr (BT_List t) = [lt|listOf(#{javaStructFieldDescriptorInitTypeExpr t})|]
+        javaStructFieldDescriptorInitTypeExpr (BT_Set t) = [lt|setOf(#{javaStructFieldDescriptorInitTypeExpr t})|]
+        javaStructFieldDescriptorInitTypeExpr (BT_Map k v) = [lt|mapOf(#{javaStructFieldDescriptorInitTypeExpr k}, #{javaStructFieldDescriptorInitTypeExpr v})|]
+        javaStructFieldDescriptorInitTypeExpr (BT_TypeParam param) = [lt|#{paramName param}|]
+        javaStructFieldDescriptorInitTypeExpr (BT_UserDefined e@Enum {..} _) = [lt|#{qualifiedDeclaredTypeName java e}.BOND_TYPE|]
+        javaStructFieldDescriptorInitTypeExpr (BT_UserDefined s@Struct {..} params) = [lt|#{structFieldDescriptorInitStructExpr java (qualifiedDeclaredTypeName java s) params}|]
+        javaStructFieldDescriptorInitTypeExpr (BT_UserDefined s@Forward {..} params) = [lt|#{structFieldDescriptorInitStructExpr java (qualifiedDeclaredTypeName java s) params}|]
```
sigansky: Field can be Forward. That's how circular data structures must be represented in Haskell (which lacks pointers). 

---
In reply to: [125132058](https://github.com/Microsoft/bond/pull/517#discussion_r125132058) [](ancestors = 125132058)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -10,25 +10,216 @@ module Language.Bond.Codegen.Java.Class_java
...
+structBaseDescriptorInitStructExpr _ Nothing = [lt|null|]
+structBaseDescriptorInitStructExpr java (Just t) = structFieldDescriptorInitTypeExpr java t
+
+
+-- returns an expression for the default value of a field that has an explicit default, used in intiialization of struct field descriptors
+fieldDefaultValueInitParamExpr :: MappingContext -> Type -> Maybe Default -> Text
+fieldDefaultValueInitParamExpr java type_ default_ = javaFieldDefaultValueInitParamExpr type_ default_
+    where
+        javaFieldDefaultValueInitParamExpr _ (Just (DefaultBool val)) = if val
+            then [lt|true|]
```
sigansky: We shouldn't rely on Haskell's string constants to generate Java. Further, in Haskell they are capitalized and in Java are lowercase.

---
In reply to: [125134970](https://github.com/Microsoft/bond/pull/517#discussion_r125134970) [](ancestors = 125134970)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -10,25 +10,216 @@ module Language.Bond.Codegen.Java.Class_java
...
+            }
+|]
+    where
+        paramList = sepBy ", " paramName declParams
+        methodArg TypeParam {..} = [lt|com.microsoft.bond.BondType<#{paramName}> #{paramName}|]
+        checkArgument TypeParam {..} = [lt|com.microsoft.bond.helpers.ArgumentHelper.ensureNotNull(#{paramName}, "#{paramName}");|]
+
+
+-- builds text for anonymous implementation of the GenericBondTypeBuilder abstract class and assignment to the BOND_TYPE variable
+bondTypeStaticVariableDeclAsGenericBondTypeBuilder :: String -> [TypeParam] -> Text
```
sigansky: Let's address such issues later, per meeting this morning.

---
In reply to: [125135887](https://github.com/Microsoft/bond/pull/517#discussion_r125135887) [](ancestors = 125135887)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
sigansky: It's not just generic vs. non-generic class, it also has to do with generic base vs, non-generic base and the same for fields. Two templates will not help much and having many templates increases complexity.

---
In reply to: [125346136](https://github.com/Microsoft/bond/pull/517#discussion_r125346136) [](ancestors = 125346136)

sigansky: Is discussed, I'll try to split this by Java method.

---
In reply to: [125735980](https://github.com/Microsoft/bond/pull/517#discussion_r125735980) [](ancestors = 125735980,125346136)

compiler/Compiler.lkshw
tstein: Feel free to add a pattern to .gitignore for this.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: I don't understand. This code is
```
f = x
  where x = some stuff
```
What does binding a name to `x` add?

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: Line 192 in this diff:
`((__StructBondTypeImpl)BOND_TYPE).initializeStructFields(this);`

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: Honestly, I'm not sure which is better. I guess we can figure it out later.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -10,25 +10,216 @@ module Language.Bond.Codegen.Java.Class_java
...
+        javaFieldDefaultValueInitParamExpr BT_UInt8 (Just (DefaultInteger val)) = [lt|(byte)#{val}|]
+        javaFieldDefaultValueInitParamExpr BT_UInt16 (Just (DefaultInteger val)) = [lt|(short)#{val}|]
+        javaFieldDefaultValueInitParamExpr BT_UInt32 (Just (DefaultInteger val)) = [lt|#{val}|]
+        javaFieldDefaultValueInitParamExpr BT_UInt64 (Just (DefaultInteger val)) = [lt|#{val}L|]
+        javaFieldDefaultValueInitParamExpr BT_Float (Just (DefaultFloat val)) = [lt|#{val}F|]
+        javaFieldDefaultValueInitParamExpr BT_Double (Just (DefaultFloat val)) = [lt|#{val}D|]
+        javaFieldDefaultValueInitParamExpr BT_String (Just (DefaultString val)) = [lt|"#{val}"|]
+        javaFieldDefaultValueInitParamExpr BT_WString (Just (DefaultString val)) = [lt|"#{val}"|]
+        javaFieldDefaultValueInitParamExpr (BT_UserDefined e@Enum {..} _) (Just (DefaultEnum val)) = [lt|#{qualifiedDeclaredTypeName java e}.#{val}|]
+        javaFieldDefaultValueInitParamExpr t _ = error $ "invalid declaration type for fieldDefaultValueInitParamExpr: " ++ (show t)
```
tstein: With the code in this PR, building compat fails for me with
`gbc: invalid declaration type for fieldDefaultValueInitParamExpr: BT_Double`
Presumably, a `BT_Double None` is making it through to this function. You mentioned running it successfully, so maybe you've already fixed this.

java/core/src/test/java/com/microsoft/bond/testmodel/schema.bond
sigansky: Not sure what this is about, TBD.

---
In reply to: [125347070](https://github.com/Microsoft/bond/pull/517#discussion_r125347070) [](ancestors = 125347070)

compiler/src/Language/Bond/Codegen/Java/Util.hs
sigansky: Will check on this.

---
In reply to: [125346538](https://github.com/Microsoft/bond/pull/517#discussion_r125346538) [](ancestors = 125346538)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
sigansky: I can make a companion static variable that is used by constructor and is the same reference as BOND_TYPE. But BOND_TYPE is public while the implementation class should be private.

---
In reply to: [125776117](https://github.com/Microsoft/bond/pull/517#discussion_r125776117) [](ancestors = 125776117)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -10,25 +10,362 @@ module Language.Bond.Codegen.Java.Class_java
...
+structFieldDescriptorInitTypeExpr java (BT_UserDefined e@Enum {..} _) = [lt|#{qualifiedDeclaredTypeName java e}.BOND_TYPE|]
+structFieldDescriptorInitTypeExpr java t@(BT_UserDefined s@Struct {..} params) = [lt|#{structFieldDescriptorInitStructExpr java t (qualifiedDeclaredTypeName java s) params}|]
+structFieldDescriptorInitTypeExpr java t@(BT_UserDefined s@Forward {..} params) = [lt|#{structFieldDescriptorInitStructExpr java t (qualifiedDeclaredTypeName java s) params}|]
+structFieldDescriptorInitTypeExpr _ t = error $ "invalid declaration type for structFieldDescriptorInitTypeExpr: " ++ show t
+
+
+-- given struct base type, returns a type descriptor expression
+structBaseDescriptorInitStructExpr :: MappingContext -> Maybe Type -> Text
+structBaseDescriptorInitStructExpr _ Nothing = [lt|null|]
+structBaseDescriptorInitStructExpr java (Just t) = structFieldDescriptorInitTypeExpr java t
```
chwarr: Consider `structBaseDescriptorInitStructExpr java t = maybe [lt|null|] (structFieldDescriptorTypeExpr java) t`

See also: [the `maybe` function](https://hackage.haskell.org/package/base-4.9.1.0/docs/Prelude.html#v:maybe)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -10,25 +10,362 @@ module Language.Bond.Codegen.Java.Class_java
...
+        }|]
+    where
+        typeArgVarDeclList = newlineBeginSep 3 typeArgVarDecl indexedDeclParams
+            where
+                typeArgVarDecl (index, typeParam) = [lt|#{typeParamVarDecl typeParam} = this.getGenericSpecialization().getGenericTypeArgument(#{index});|]
+                indexedDeclParams = zip [0 :: Int ..] declParams
+        fieldDescriptorInitList = newlineBeginSep 3 fieldDescriptorInit structFields
+            where
+                fieldDescriptorInit Field {..} = [lt|this.#{fieldName} = new #{structFieldDescriptorTypeName java fieldType}(#{constructorParams});|]
+                    where
```
chwarr: The `where` clauses don't need to be nested each time. It's pretty common to write a `where` clause entry that refers to another where clause entry right below it. Here's a toy(ish) example:

    where
      fieldInits = basicInits ++ complexInits
      basicInits = map defaultInit basicFields
      complexInits = map explicitInit complexFields
      (complexFields,basicFields) = Data.List.partition hasExplicitDefault structFields

compiler/src/Language/Bond/Codegen/Java/Util.hs
```diff
@@ -4,110 +4,69 @@
...
+isPrimitiveBondType (BT_UserDefined Enum {..} _) = True
+isPrimitiveBondType t = isPrimitiveNonEnumBondType t
+
+-- returns a value indicating whether a type is a generic struct type with generic type parameters
+isGenericBondStructType :: Type -> Bool
+isGenericBondStructType (BT_UserDefined Struct {..} _) = not (null declParams)
+isGenericBondStructType _ = False
+
+-- takes a predicate and two texts and outputs one based on predicate
+ifThenElse :: Bool -> Text -> Text -> Text
```
tstein: I like the readability of `ifThenElse`, so I'd like to keep that; if we're going to have our own, I think I'd prefer to keep it typed as it is. Fresh eyes will more readily understand `Couldn't match expected type Text with actual type Yak in a call to ifThenElse` than `Not defined: toText Yak` and the rather prolix output that comes with that error. That okay with you?

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -10,25 +10,362 @@ module Language.Bond.Codegen.Java.Class_java
...
+        typeName (BT_Maybe BT_Bool) = [lt|com.microsoft.bond.StructBondType.SomethingBoolStructField|]
+        typeName BT_Bool = [lt|com.microsoft.bond.StructBondType.BoolStructField|]
+        typeName (BT_Maybe BT_String) = [lt|com.microsoft.bond.StructBondType.SomethingStringStructField|]
+        typeName BT_String = [lt|com.microsoft.bond.StructBondType.StringStructField|]
+        typeName (BT_Maybe BT_WString) = [lt|com.microsoft.bond.StructBondType.SomethingWStringStructField|]
+        typeName BT_WString = [lt|com.microsoft.bond.StructBondType.WStringStructField|]
+        typeName (BT_Maybe (BT_UserDefined e@Enum {..} _)) = [lt|com.microsoft.bond.StructBondType.SomethingEnumStructField<#{qualifiedDeclaredTypeName java e}>|]
+        typeName (BT_UserDefined e@Enum {..} _) = [lt|com.microsoft.bond.StructBondType.EnumStructField<#{qualifiedDeclaredTypeName java e}>|]
+        typeName (BT_Maybe t) = [lt|com.microsoft.bond.StructBondType.SomethingObjectStructField<#{(getTypeName java) t}>|]
+        typeName t = [lt|com.microsoft.bond.StructBondType.ObjectStructField<#{(getTypeName java) t}>|]
```
tstein: Agreed, but I don't want to bank on that. Changed to `t@(BT_UserDefined _ _)` and made the catch-all an error.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -10,25 +10,362 @@ module Language.Bond.Codegen.Java.Class_java
...
+
+
+-- given a type parameter, returns the declaration of a local variable containing the type descriptor
+typeParamVarDecl :: TypeParam -> Text
+typeParamVarDecl t@TypeParam {..} = [lt|com.microsoft.bond.BondType<#{paramName}> #{typeParamVarName t}|]
+
+
+-- given a list of type parameters, returns it as comma-separated text
+typeParamNameList :: [TypeParam] -> Text
+typeParamNameList declParams = [lt|#{sepBy ", " paramName declParams}|]
```
tstein: I actually find interpolation one-liners to be the easiest to read, though I am open to the possibility that this is because I've done too much Perl and sh. Will change it if you'd like.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: Eduardo and I flattened them a bit. We kept the `where`s that were being used for `Field {..}` expansion.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: I like it. Done.

compiler/src/Language/Bond/Codegen/Java/Util.hs
```diff
@@ -4,110 +4,69 @@
...
-    BT_Vector _              -> "BT_LIST"
-    BT_Set _                 -> "BT_SET"
-    BT_Map _ _               -> "BT_MAP"
-    BT_UserDefined Enum {} _ -> "BT_INT32"
-    BT_UserDefined _ _       -> "BT_STRUCT"
-    -- FIXME: Marker for unsupported types that compiles.
-    _                        -> "BT_UNAVAILABLE"
+-- returns the fully qualified name for a declaration
+qualifiedDeclaredTypeName :: MappingContext -> Declaration -> String
+qualifiedDeclaredTypeName java d@Struct {..}  = intercalate "." $ getDeclNamespace java d ++ [declName]
```
tstein: Done. Good call.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -10,25 +10,362 @@ module Language.Bond.Codegen.Java.Class_java
...
+fieldDefaultValueInitParamExpr java(BT_UserDefined e@Enum {..} _) (Just (DefaultEnum val)) = [lt|, #{qualifiedDeclaredTypeName java e}.#{val}|]
+fieldDefaultValueInitParamExpr _ t _ = error $ "invalid declaration type for fieldDefaultValueInitParamExpr: " ++ show t
+
+
+-- given a struct name and type parameters, and type, returns a type descriptor expression for the field, used in initialization of struct field descriptors
+structFieldDescriptorInitStructExpr :: MappingContext -> Type -> String -> [Type] -> Text
+structFieldDescriptorInitStructExpr java fieldType typeName params = [lt|#{typeCastExpr} getStructType(#{typeName}.class#{paramExprList params})|]
+    where
+        typeCastExpr = [lt|(com.microsoft.bond.StructBondType<#{(getTypeName java) fieldType}>)|]
+        paramExprList :: [Type] -> Text
```
tstein: Did option B.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: It seems we're expecting the default for those types. I'm going to put this back for now.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: This goes in an arg list that has stuff in it either way, so this needs to start with a `", "` if there are any params. We should probably refactor this to assemble all of those parameters and then `intercalate ", "` them all, but I'm going to put it back for now.

chwarr: Makes sense.

compiler/src/Language/Bond/Codegen/Java/Util.hs
chwarr: It's in Codegen/Java/Util.hs, so I'm fine leaving it as Text. If we move this some where more general, it should be polymorphic.

tstein:
Merged with 33747044

PR #520, opened 2017-07-04
[C#] Fix Reflection.IsBonded for custom IBonded implementations

chadwalters:
Reflection.IsBonded now correctly returns true for custom IBonded
implementations. In addition, corrected reversed logic in test utility
code that was likely related to the incorrect predicate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,12 +16,12 @@ different versioning scheme, following the Haskell community's
 * IDL core version: TBD
 * IDL comm version: TBD
 * C++ version: TBD
-* C# NuGet version: TBD  (major bump needed)
+* C# NuGet version: TBD  (bug fix bump needed)
 * C# Comm NuGet version: TBD
 
 ### C# ###
 
-* **Breaking change** Reflection.IsBonded now recognizes custom IBonded
+* Reflection.IsBonded now recognizes custom IBonded
 implementations.
 
 ## 6.0.0: 2017-06-29  ##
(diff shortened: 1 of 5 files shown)


cs/test/core/TypeAliasTests.cs
```diff
@@ -301,13 +302,15 @@ public void AliasBlob()
         public void AliasBonded()
         {
             var from = new BondedAlias {lazy = new Lazy<Foo>(UnitTest.Random.Init<Foo>())};
+            Assert.IsTrue(Reflection.IsBonded(typeof(Lazy<Foo>)));
```
tstein: It might be a little better if the two asserts you added got their own tests - `CustomIBondedImpl_Reflection_IsBonded`, or something like that. No big either way.

chadwalters: I added some tests in ReflectionTests -- see second commit.

CHANGELOG.md
```diff
@@ -11,11 +11,24 @@ tag versions. The Bond compiler (`gbc`) and
...
+* `gbc` & compiler library: TBD
+* IDL core version: TBD
+* IDL comm version: TBD
+* C++ version: TBD
+* C# NuGet version: TBD  (major bump needed)
+* C# Comm NuGet version: TBD
+
+### C# ###
+
+* **Breaking change** Reflection.IsBonded now recognizes custom IBonded
```
chwarr: I agree that some observable behavior has changed, which may cause some code somewhere to break. For similar fixes (like https://github.com/Microsoft/bond/issues/417) we've just considered these bug fixes. Am I missing some part of a larger picture?

chadwalters: For me this was on the border and I erred on the side of cautious. I will update to treat it as a bugfix.

PR #522, opened 2017-07-07
java: Add golden tests.

tstein:
Meta and alias don't work due to not being implemented in this version of Java codegen.

### DIFF
--- a/compiler/Options.hs
+++ b/compiler/Options.hs
@@ -58,6 +58,7 @@ data Options
         { files :: [FilePath]
         , import_dir :: [FilePath]
         , output_dir :: FilePath
+        , using :: [String]
         , namespace :: [String]
         , jobs :: Maybe Int
         , no_banner :: Bool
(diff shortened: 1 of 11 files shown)


compiler/Options.hs
```diff
@@ -58,6 +58,7 @@ data Options
         { files :: [FilePath]
         , import_dir :: [FilePath]
         , output_dir :: FilePath
+        , using :: [String]
```
tstein: gbc-tests will unconditionally try to `using options`, which will fail if it's a Java options. Rather than create code paths for "all target languages except Java", I just added it back in.

lalo: this will be misleading for the end user since this feature will not be supported right?

lalo:
Travis shows some warnings regarding the current codegen, not in the scope of this PR but keep it in mind.

compiler/Options.hs
tstein: Done.

PR #523, opened 2017-07-11
[C#] Add support for uint64 in JSON when possible

chadwalters:
Newtonsoft JSON.NET supports uint64 where BigInteger is supported. This pull request allows Bond's SimpleJsonProtocol to leverage that support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@ different versioning scheme, following the Haskell community's
 ### C# ###
 
 * Reflection.IsBonded now recognizes custom IBonded implementations.
+* Use Newtonsoft's JSON.NET BigInteger support -- when available -- to
+  handle the full range of uint64 values in the SimpleJson protocol (.NET
+  4.5 or greater, .NET Standard 1.6 or greater).
 
 ## 6.0.0: 2017-06-29  ##
 * `gbc` & compiler library: 0.10.0.0
--- a/cs/Compiler.vcxproj
+++ b/cs/Compiler.vcxproj
@@ -33,8 +33,6 @@ call cmake $(ProjectDir)\..\compiler -Wno-dev
 if %errorlevel% neq 0 goto :cmEnd
 call cmake --build $(OutDir) --target gbc
 if %errorlevel% neq 0 goto :cmEnd
-copy $(OutDir)\build\gbc\gbc.exe $(ProjectDir)\tools\
-if %errorlevel% neq 0 goto :cmEnd
 :cmEnd
 endlocal &amp; call :cmErrorLevel %errorlevel% &amp; goto :cmDone
 :cmErrorLevel
@@ -81,4 +79,7 @@ if %errorlevel% neq 0 goto :VCEnd
     <None Include="$(ProjectDir)..\compiler\src\Language\Bond\Codegen\Cs\Grpc_cs.hs" />
   </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
+  <Target Name="AfterBuild">
+    <Copy SourceFiles="$(OutDir)\build\gbc\gbc.exe" DestinationFolder="$(ProjectDir)\tools\" />
+  </Target>
 </Project>
--- a/cs/dnc/src/json/project.json
+++ b/cs/dnc/src/json/project.json
@@ -37,10 +37,10 @@
     "dependencies": {
         "attributes": "1.0.0-*",
         "core": "1.0.0-*",
+        "reflection": "1.0.0-*",
         "Newtonsoft.Json": {
             "version": "9.0.1"
-        },
-        "reflection": "1.0.0-*"
+        }
     },
     "frameworks": {
         "netstandard1.0": {
@@ -49,6 +49,11 @@
             }
         },
         "netstandard1.6": {
+            "buildOptions": {
+                "define": [
+                    "SUPPORTS_BIGINTEGER"
+                ]
+            },
             "dependencies": {
                 "NETStandard.Library": "1.6.0"
             }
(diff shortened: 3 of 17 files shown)


cs/Compiler.vcxproj
```diff
@@ -81,4 +79,8 @@ if %errorlevel% neq 0 goto :VCEnd
     <None Include="$(ProjectDir)..\compiler\src\Language\Bond\Codegen\Cs\Grpc_cs.hs" />
   </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
+  <Target Name="AfterBuild">
+    <Copy SourceFiles="$(OutDir)\build\gbc\gbc.exe" DestinationFolder="$(ProjectDir)\tools\" />
+    <Message Text="CHWARR WAS HERE" />
```
tstein: Uh, was he?

cs/src/json/JSON.csproj
```diff
@@ -19,16 +26,22 @@
     <Compile Include="protocols\SimpleJsonReader.cs" />
     <Compile Include="protocols\SimpleJsonWriter.cs" />
   </ItemGroup>
+  <ItemGroup Condition="'$(BuildFramework)' == 'net45-nonportable'">
```
tstein: Is it better to switch on `net45-nonportable` than something like `$(BuildNonportable) == true`?

cs/test/core/Core.csproj
```diff
@@ -109,7 +113,7 @@
       <HintPath>..\..\src\io\$(DependentOutputPath)\Bond.IO.dll</HintPath>
     </Reference>
     <Reference Include="Bond.JSON">
-      <HintPath>..\..\src\json\$(DependentOutputPath)\Bond.JSON.dll</HintPath>
+      <HintPath>..\..\src\json\$(OrigOutputPath)\Bond.JSON.dll</HintPath> <!-- Intentionally not DependentOutputPath -->
```
tstein: Why?

cs/test/core/SerializationTests.cs
```diff
@@ -49,7 +49,7 @@ public void IntegerLimits()
                 _uint8 = byte.MaxValue,
                 _uint16 = ushort.MaxValue,
                 _uint32 = uint.MaxValue,
-                // Note: not ulong.MaxValue because NewtonSoft JSON doesn't support it
+                // Note: not ulong.MaxValue because NewtonSoft JSON doesn't support it in portable manner
```
tstein: "portable profile", as you said in the next file?

cs/dnc/src/json/project.json
```diff
@@ -37,20 +37,28 @@
...
+            "buildOptions": {
+                "define": [
+                    "SUPPORTS_BIGINTEGER"
+                ]
+            },
             "dependencies": {
-                "NETStandard.Library": "1.6.0"
+                "NETStandard.Library": "1.6.0",
+                "Newtonsoft.Json": {
+                    "version": "10.0.1"
```
tstein: I'm a little confused. Version 10.0.1 doesn't appear anywhere else in this file - looks like the only thing referenced is v9.0.1, regardless of portable/non. What am I missing?

cs/Compiler.vcxproj
chadwalters: Oops. Deleted.

cs/src/json/JSON.csproj
chadwalters: Let me play with it. I tried it that way before and saw some oddities related to project references for dependencies but some other fixes may have addressed that now.

cs/dnc/src/json/project.json
chadwalters: Version 10.0.1 picked up support for BigInteger for .NET Standard 1.1 or greater.

cs/test/core/SerializationTests.cs
chadwalters: Done.

cs/test/core/Core.csproj
chadwalters: To pick up the version of Bpnd.JSON.dll that matches BuildFramework. Comment updated.

cs/src/json/JSON.csproj
chadwalters: Result is still the same. The behavior of the solution is less confusing with the current mechanism.

chadwalters:
Addressed Ted's comments and add entry in CHANGELOG.

Please squash commits when merging.

chadwalters:
gRPC build on Travis failed for what should be an unrelated reason. Restarting Travis build.

cs/dnc/src/json/project.json
chadwalters: On further review, addressed this differently. See fixup commit.

CHANGELOG.md
```diff
@@ -21,8 +21,9 @@ different versioning scheme, following the Haskell community's
 
 ### C# ###
 
-* Reflection.IsBonded now recognizes custom IBonded
-implementations.
+* Reflection.IsBonded now recognizes custom IBonded implementations.
+* Use Newtonsoft's BigInteger support to handle ful range of uint64 values in
```
lalo: full

cs/test/core/Util.cs
```diff
@@ -999,12 +999,16 @@ public static IBonded<T> MakeBondedSP<T>(T obj)
                 if (!AnyField<From>(Reflection.IsBonded))
                 {
                     streamTranscode(SerializeSP, TranscodeSPXml<From>, DeserializeXml<To>);
-                
-                    // NewtonSoft JSON doesn't support uint64
+
+#if SUPPORTS_BIGINTEGER
```
chwarr: This runs the transcoding twice when we do have BigInteger support. What do you think about structuring it this way?

    #if SUPPORTS_BIGINTEGER
        const bool hasBigInteger = true;
    #else
        const bool hasBigInteger = false;
    #endif

    // NewtonSoft JSON doesn't support uint64 in portable profile, so we skip that type
    if (hasBigInteger || (typeof (From) != typeof (MaxUInt64)))
    {
        streamTranscode(SerializeSP, TranscodeSPJson<From>, DeserializeJson<To>);
    }

And a similar condition later on for the if that guards to tests around line 1039.

chwarr: I misread where the `#if` and `#else` blocks were. This doesn't run the tests twice when there's BigInteger support. If you want to switch, go for it, but no need to.

chadwalters: I switched anyway -- it comes out a bit cleaner.

cs/src/json/JSON.csproj
```diff
@@ -19,16 +26,22 @@
...
+    <Reference Include="System.Numerics" />
```
chadwalters: Addressed.

CHANGELOG.md
chadwalters: Addressed

chwarr:
LGTM. Will wait for CI builds before merging.

PR #524, opened 2017-07-11
Update README for Win C++ Core build without gRPC

chadwalters:


README.md
```diff
@@ -211,22 +211,14 @@ pre-built libraries are only needed for unit tests, Python, gRPC, and Comm
...
-In order to generate a solution to build the C++ and Python versions with Visual
+In order to generate a solution to build the Bond Core C++ and Python with Visual
 Studio 2015 run the following commands from the root `bond` directory:
 
 ```bash
 mkdir build
 cd build
 set PreferredToolArchitecture=x64
-cmake -G "Visual Studio 14 2015 Win64" ..
+cmake -DBOND_ENABLE_COMM=FALSE -DBOND_ENABLE_GRPC=FALSE -G "Visual Studio 14 2015 Win64" ..
```
tstein: Consider swapping these `-D`s in both uses in this file - GRPC is the more interesting one.

chadwalters: Let's leave that to another round where we make Comm opt-in.

README.md
```diff
@@ -249,6 +241,21 @@ In order to build and execute the unit tests and examples run:
...
+To build Bond's gRPC++ integration from source, some of
+[gRPC's prerequisites](https://github.com/grpc/grpc/blob/master/INSTALL.md#building-using-cmake-with-boringssl)
+are also needed:
+
+```bash
+choco install activeperl golang ninja yasm
+```
+
+You will also need to enable gRPC in the cmake configuration step by running the following
+in the `build` directory from above and then following the other cmake commands above:
```
chwarr: Minor: consider "CMake commands..." or "`cmake` commands"

PR #527, opened 2017-07-11
[C++] Don't build Comm by default.

tstein:
Need to check that CI not only passes, but that each flavor includes or does not include Comm, as appropriate.


chadwalters:
The phrase "and Bond Comm" appears in the README text. While you are at it, add "[C++]" to the commit message.

Issue #498, opened 2017-06-13
Bond.IO.Unsafe.InputStream cannot handle an underlying stream that doesn't support Seek


chwarr:
A private bug report came to us:

> I’m facing a problem in which I’m trying to deserialize a bond enumerable object directly from a GZip stream.
> If I first decompress the stream into a memory stream, it all works great. But if I try to deserialize directly from the GZipStream I get an InvalidOperationException since GZipStream doesn’t support Seek method.
>
> This exception is thrown from \[[`Bond.IO.Unsafe.InputStream.EndOfStream()`](https://github.com/Microsoft/bond/blob/7b4b5f8821bee3a8577bfd3003379452f4b72a62/cs/src/io/unsafe/InputStream.cs#L102)\].
>
> NOTES:
> -	I believe Seek is not really required here and you can call ‘Read’ instead with the positive offset.
> -	That’s the only Seek I found in Bond sources.

gerlas:
This looks like it may also be an issue for me, as I look to move existing code to use Bond.
All (de)serialization I do - is done within a compressed stream.

Example usage...
```
// Writing data as compressed stream to file (with progress notification)
using (var fileStream = new FileStream(filename, FileMode.Create, FileAccess.Write, FileShare.None))
{
   using (var progressStream = new ReadProgressStream(fileStream)) 
   {
      progressStream.ProgressChanged += progressStream_ProgressChanged;
      using (var deflateStream = new DeflateStream(progressStream, CompressionMode.Decompress, true))
      {
         // WRITE data...  using deflateStream
      }
   }
}

// Reading data from compressed stream in file (with progress notification).
// Buffering required for de-compression to avoid memory bloat.
using (var fileStream = new FileStream(filename, FileMode.Open, FileAccess.Read, FileShare.Read))
{
   using (var progressStream = new ReadProgressStream(fileStream)) 
   {
      progressStream.ProgressChanged += progressStream_ProgressChanged;
      using (var bufferedStream = new BufferedStream(progressStream, bufferSize))
      {
         using (var deflateStream = new DeflateStream(bufferedStream, CompressionMode.Compress, true)
         {
            // READ data...  using deflateStream
         }
      }
   }
}
```

PR #529, opened 2017-07-14
[c#] Handle non-seekable streams in InputStream

chwarr:
InputStream would always try to call Stream.Seek when it needed to skip
over some bytes in the underlying stream, even if the stream didn't
support Seek.

Now, it checks whether the stream can seek. If not, it makes dummy Read
calls to advance the underlying stream.

Regression test added, and all the implementations of IInputStream that
ship with Bond are tested against these tests.

Existing InputStream tests folded into these IInputStreams tests, and
tested against both seekable and non-seekable streams.

Fixes https://github.com/Microsoft/bond/issues/498

### DIFF
--- a/cs/src/io/unsafe/InputStream.cs
+++ b/cs/src/io/unsafe/InputStream.cs
@@ -4,6 +4,7 @@
 namespace Bond.IO.Unsafe
 {
     using System;
+    using System.Diagnostics;
     using System.IO;
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


cs/src/io/unsafe/InputStream.cs
```diff
@@ -93,30 +96,45 @@ public override ArraySegment<byte> ReadBytes(int count)
...
+            // ton of memory ahead of time.
             if ((count > buffer.Length && (count - buffer.Length > ActiveAllocationChunk)))
             {
-                // Calculate number of temp buffers -1 in case difference is exactly
-                // multiple of chunk size.
-                numTempBuffers = (count - buffer.Length - 1) / ActiveAllocationChunk;
+                // Calculate number of temp buffers - 1 in case difference is
+                // exactly multiple of chunk size. If it's exactly a
+                // multiple, we need one fewer temp buffers, as we'll read
+                // the last chunk directly into the final buffer, below.
```
chadwalters: Comment is not really accurate (partly due to confused formatting on previous version). How about something like?
"Calculate number of temp buffers; we round down since last chunk is read directly into final buffer. Note: Difference is adjusted by -1 to round down correctly in case where difference is exactly multiple of allocation chunk size."

chwarr:
Addressed comments. Squash merge needed now.

PR #532, opened 2017-07-19
[Java Core] Bonded changes part 1 (interface and deep object cloning)

sigansky:


java/core/src/main/java/com/microsoft/bond/NullableBondType.java
```diff
@@ -82,6 +82,11 @@ protected final TValue newDefaultValue() {
     }
 
     @Override
+    protected final TValue cloneValue(TValue value) {
+        return value == null ? null : this.valueType.cloneValue(value);
```
tstein: If someone ends up cloning a struct with a non-nullable field which is set to null, how will that exception/stacktrace look?

java/core/src/main/java/com/microsoft/bond/Bonded.java
```diff
@@ -3,28 +3,117 @@
...
+    /**
+     * Creates a new Bonded instance from an input stream that is constrained to a given struct type.
+     *
+     * @param inputStream input stream containing the serialized representation of this Bonded value
+     * @param bondType    the type descriptor of the struct type
+     * @param <T>         the struct type
+     * @return new Bonded instance backed by the stream
+     * @throws IllegalArgumentException if an argument is null
+     */
+    public static <T extends BondSerializable> Bonded<T> fromStream(
```
chwarr: As a human reader, the `T` type parameters on these static methods shadow the class-level `T`. I know that the Java compiler doesn't seem them as shadowing, but consider using a different parameter name (`U`?) to make it easier for the humans to read.

java/core/src/main/java/com/microsoft/bond/BondedBondType.java
```diff
@@ -79,30 +81,35 @@ public final boolean isGenericType() {
 
     @Override
     protected final Bonded<TStruct> newDefaultValue() {
-        return new Bonded<TStruct>(this.valueType.newDefaultValue());
+        // since Bonded is immutable, the default value can be shared
```
sigansky: Bonded is immutable: the underlying stream/object is cloned but never modified directly. The default value wraps a default object value. 

---
In reply to: [128378278](https://github.com/Microsoft/bond/pull/532#discussion_r128378278) [](ancestors = 128378278)

java/core/src/main/java/com/microsoft/bond/ListBondType.java
```diff
@@ -98,13 +107,14 @@ protected final void serializeValue(SerializationContext context, List<TElement>
         this.verifyNonNullableValueIsNotSetToNull(value);
         int count = value.size();
         context.writer.writeContainerBegin(count, this.elementType.getBondDataType());
-        for (int i = 0; i < count; ++i) {
+        int i = 0;
+        for (TElement element : value) {
```
sigansky: I noticed an inefficient O(N^2) implementation here, that was copy-paste from the VectorBondType class. I'd like to keep it as is, since the change is small and important for perf.

---
In reply to: [128378679](https://github.com/Microsoft/bond/pull/532#discussion_r128378679) [](ancestors = 128378679)

sigansky: That is: I'd like to avoid an overhead of a separate PR.

---
In reply to: [128562299](https://github.com/Microsoft/bond/pull/532#discussion_r128562299) [](ancestors = 128562299,128378679)

java/core/src/main/java/com/microsoft/bond/NullableBondType.java
sigansky: Cloning doesn't check for this condition (serialization does). It may be cloned as is, and an error will manifest during serialization. Or it may fail with null reference, which we can't always detect without significant perf penalty.

---
In reply to: [128379441](https://github.com/Microsoft/bond/pull/532#discussion_r128379441) [](ancestors = 128379441)

java/core/src/main/java/com/microsoft/bond/Bonded.java
sigansky: I prefer to keep it as is, as it may be even more confusing if the parameter name was different. The "T" in static context means exactly the same thing as the "T" in instance/class context, that is the struct type within the Bonded.

---
In reply to: [128396886](https://github.com/Microsoft/bond/pull/532#discussion_r128396886) [](ancestors = 128396886)

java/core/src/main/java/com/microsoft/bond/BondedBondType.java
tstein: Okay, I'm fine with this with a clearer comment. How about:
`// Since Bonded<TStruct> provides no API for accessing the TStruct inside it, the default value can be shared.`

PR #534, opened 2017-07-21
[c++] Fix potential simultaneous codegen

chwarr:
In some CMakeLists.txt files, codegen was being run on the same .bond
files multiple times. Depending on build scheduling, this could result
in build failures (if the generation were to happen at the same time) or
spurious rebuilds (due to spurious rewrites of the generated files).

Now, .bond files that are shared across targets within the same
directory are generated once and compiled into a static library that is
then linked into all the consumers.

Since we're compiling the .bond files into a static library, we can take
advantage of the _apply.cpp precompilation as well.

### DIFF
--- a/cpp/test/comm/CMakeLists.txt
+++ b/cpp/test/comm/CMakeLists.txt
@@ -28,18 +28,36 @@ function (add_unit_test)
     cxx_target_compile_options (AppleClang ${name} PRIVATE -DBOOST_ASIO_HAS_STD_CHRONO)
 endfunction()
 
-
 # Build common code into its own library.
-add_library (comm_test_common EXCLUDE_FROM_ALL "main.cpp" "idl.cpp")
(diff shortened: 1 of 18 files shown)


cpp/test/comm/CMakeLists.txt
```diff
@@ -28,27 +28,45 @@ function (add_unit_test)
     cxx_target_compile_options (AppleClang ${name} PRIVATE -DBOOST_ASIO_HAS_STD_CHRONO)
 endfunction()
 
-
 # Build common code into its own library.
```
lalo: nit: misleading comment if we follow the other two Cmakelists.txt as the standard. "shared generated code"

chwarr: This one is more than just shared generated code. It also has shared `main()` and some other things.

Issue #535, opened 2017-07-21
Update macOS build instructions with gRPC specifics


chwarr:
The macOS build instructions do not mention/link to how to install the gRPC dependencies.
The "default" CMake command should also include `-DBOND_ENABLE_GRPC=FALSE` to make getting a successful build more likely.

chwarr:
Given that Bond-over-gRPC [is being deprecated and removed](https://github.com/microsoft/bond/issues/1131), I'm going to close this as "Won't Fix"

PR #537, opened 2017-07-21
Skip gRPC for initial build on macOS in README

chwarr:
Also, lay out the CMake commands following the template that `cmake
--help` documents: `cmake [options] <path-to-source>`

Partial fix for https://github.com/Microsoft/bond/issues/535

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ In the root `bond` directory run:
 ```bash
 mkdir build
 cd build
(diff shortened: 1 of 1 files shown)


lalo:
Partial fix as in we need to add the instructions on getting the pre-requisites on mac? If th

chwarr:
Yes, to close #535, we should also update the README with instructions/a link to getting the gRPC prerequisites for macOS as well.

PR #539, opened 2017-07-26
[Java] Updated codegen to handle type aliases and emit more specialized return type for getBondType

sigansky:


lalo:
This should produce a change on the golden tests and they need to be updated to reflect so.

sigansky:
Done.

---
In reply to: [318175883](https://github.com/Microsoft/bond/pull/539#issuecomment-318175883) [](ancestors = 318175883)

PR #540, opened 2017-07-31
[Java] Implement limited version of Bonded sufficient to pass compat tests, as well as made fixes to compat tests

sigansky:


java/core/src/main/java/com/microsoft/bond/TaggedProtocolStreamBonded.java
```diff
@@ -0,0 +1,72 @@
...
+import com.microsoft.bond.protocol.ProtocolWriter;
+import com.microsoft.bond.protocol.TaggedProtocolReader;
+
+import java.io.IOException;
+
+/**
+ * An implementation of {@link Bonded} backed by a tagged protocol reader, which effectively
+ * implements delayed deserialization for tagged protocols.
+ *
+ * @param <T> the struct type of this Bonded instance or {@link BondSerializable} if unspecified
```
tstein: So to construct an unspecified one, you would write `new TaggedProtocolStreamBonded<BondSerializable>(reader, null)`?

java/core/src/main/java/com/microsoft/bond/protocol/StreamNotCloneableException.java
```diff
@@ -0,0 +1,24 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package com.microsoft.bond.protocol;
+
+import java.io.IOException;
+
+/**
+ * Indicates that a stream cannot be cloned because the original stream is not cloneable.
+ */
+public class StreamNotCloneableException extends IOException {
```
tstein: Not going to try to deal with it in this PR, but this is absolutely not an `IOException`. We need to be more deliberate about the exception hierarchy we throw.

cpp/test/CMakeLists.txt
```diff
@@ -4,9 +4,7 @@ cxx_add_compile_options(MSVC /wd4310)
 # Core unit tests don't build under VS2013 without this.
 cxx_add_compile_options(MSVC /EHa)
 
-if (NOT BOND_SKIP_CORE_TESTS)
-    add_subfolder (compat/core "tests/compat/core")
-endif()
+add_subfolder (compat/core "tests/compat/core")
```
sigansky: BOND_SKIP_COMPAT_TESTS

---
In reply to: [130488475](https://github.com/Microsoft/bond/pull/540#discussion_r130488475) [](ancestors = 130488475)

java/compat/src/main/java/com/microsoft/bond/compat/CompatDriver.java
```diff
@@ -58,10 +58,9 @@ public static void main(final String[] args) throws IOException {
             System.exit(STATUS_FAILURE);
         }
 
-        final Compat compat = new Compat();
         final Serializer<Compat> serializer = new Serializer<>();
-        final Deserializer<Compat> deserializer = new Deserializer<>((StructBondType<Compat>) compat.getBondType());
-        deserializer.deserialize(reader);
+        final Deserializer<Compat> deserializer = new Deserializer<>((StructBondType<Compat>) Compat.BOND_TYPE);
```
sigansky: Please look at the original code. You initially created a new Compat instance and used its getBondType() to get the type descriptor, but the the result of deserialize() call was ignored. Thus, the Compat instance remained empty and the serialize() call produced empty result.

---
In reply to: [130488530](https://github.com/Microsoft/bond/pull/540#discussion_r130488530) [](ancestors = 130488530)

java/core/src/main/java/com/microsoft/bond/SpecificTypeObjectBonded.java
```diff
@@ -0,0 +1,91 @@
...
+    public <U extends BondSerializable> Bonded<U> convert(StructBondType<U> toBondType) {
+        ArgumentHelper.ensureNotNull(toBondType, "toBondType");
+        if (this.bondType.isSubtypeOf(toBondType)) {
+            // we can always convert to a base struct type by upcasting
+            @SuppressWarnings("unchecked")
+            U castObjectInstance = (U) this.objectInstance;
+            return new SpecificTypeObjectBonded<U>(castObjectInstance, toBondType);
+        } else {
+            // but we can't convert to any other type
+            return null;
```
chwarr: C# does return `null`: it uses an `as` cast [under the hood](https://github.com/Microsoft/bond/blob/6478f4b538e9171882bf047a885587640806863b/cs/src/core/Bonded.cs#L86-L89):

```csharp
        IBonded<U> IBonded.Convert<U>()
        {
            return this as IBonded<U>;
        }
```

java/core/src/main/java/com/microsoft/bond/Bonded.java
```diff
@@ -5,106 +5,124 @@
...
-
-        // TODO: return object-backed implementation
-        throw new UnsupportedOperationException();
-    }
+    public abstract StructBondType<? extends T> getBondType();
 
     /**
-     * Serializes content of the Bonded instance to a protocol writer.
+     * Serializes content of this Bonded instance to a protocol writer using the default Bond type descriptor.
+     * Throws an exception if there is no default type descriptor associated with this Bonded instance.
```
sigansky: When you create a new Bonded from protocol reader but don't specify any type. See Bonded.fromProtocolReader with 1 argument.

---
In reply to: [130509588](https://github.com/Microsoft/bond/pull/540#discussion_r130509588) [](ancestors = 130509588)

java/core/src/main/java/com/microsoft/bond/SpecificTypeObjectBonded.java
```diff
@@ -0,0 +1,91 @@
...
+        if (this.bondType.isSubtypeOf(asBondType)) {
+            // we can always serialize the base struct type portion by slicing
+            @SuppressWarnings("unchecked")
+            U castObjectInstance = (U) this.objectInstance;
+            asBondType.serialize(castObjectInstance, protocolWriter);
+        } else {
+            // but we need to transcode in all other cases
+            // TODO: complete this implementation after transcoding is implemented
+            throw new UnsupportedOperationException(
+                    "Serialization as a non-ancestor struct type is not currently implemented.");
```
sigansky: Please refer to .NET implementation for details. This case is supported in other Bond versions and thus needs to be supported in Java. Please note that this is only a limited Bonded implementation, as was necessary for tests and initial release and we will certainly reiterate on this in near future.

---
In reply to: [130510028](https://github.com/Microsoft/bond/pull/540#discussion_r130510028) [](ancestors = 130510028)

java/core/src/main/java/com/microsoft/bond/SpecificTypeObjectBonded.java
sigansky: I believe it does, since it tries to type cast using "as" operator which evaluates to null if cast is not possible. 

---
In reply to: [130510128](https://github.com/Microsoft/bond/pull/540#discussion_r130510128) [](ancestors = 130510128)

java/core/src/main/java/com/microsoft/bond/TaggedProtocolStreamBonded.java
sigansky: Only internally since these are private classes. The Bonded class has a public static method that does that under the hood and returns Bonded<?>.

---
In reply to: [130510520](https://github.com/Microsoft/bond/pull/540#discussion_r130510520) [](ancestors = 130510520)

java/core/src/main/java/com/microsoft/bond/protocol/StreamNotCloneableException.java
sigansky: I disagree. Stream cloning is an operation in input and hence an I/O operation. The error is that there is no support for the particular type of input. In my mind, it's not much different from UnsupportedEncodingException and other like that.

---
In reply to: [130514052](https://github.com/Microsoft/bond/pull/540#discussion_r130514052) [](ancestors = 130514052)

java/compat/src/main/java/com/microsoft/bond/compat/CompatDriver.java
tstein: Oh, I wasn't even looking at that - got thrown by the static vs. instance ways of getting bond types. No problem here.

java/core/src/main/java/com/microsoft/bond/Bonded.java
tstein: Sorry, not what I meant to ask. When would the exception be thrown/when would there be no default type descriptor?

cpp/test/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ cxx_add_compile_options(MSVC /wd4310)
 # Core unit tests don't build under VS2013 without this.
 cxx_add_compile_options(MSVC /EHa)
 
-if (NOT BOND_SKIP_CORE_TESTS)
+if (NOT BOND_SKIP_COMPAT_TESTS)
```
chadwalters: When you add a new CMake variable, you should add it to Config.cmake: https://github.com/Microsoft/bond/blob/master/cmake/Config.cmake#L189

chadwalters: Also consider updating AppVeyor and Travis configs to include the new switch where appropriate.

sigansky: Done

---
In reply to: [130746141](https://github.com/Microsoft/bond/pull/540#discussion_r130746141) [](ancestors = 130746141)

java/core/src/main/java/com/microsoft/bond/Bonded.java
sigansky: Correct. The non-generic serialize/deserialize methods rely on the default type and throw when it's not specified.

This reminds me to add a similar comment to deserialize as well. Thanks.

---
In reply to: [130734564](https://github.com/Microsoft/bond/pull/540#discussion_r130734564) [](ancestors = 130734564)

PR #542, opened 2017-08-01
java: untagged and simple binary

tstein:
Still working on tests, but there's already a lot to review. Strongly recommend doing it one commit at a time.


java/core/src/test/java/com/microsoft/bond/testmodel/A.java
```diff
@@ -6,7 +6,6 @@
 import com.microsoft.bond.*;
 import com.microsoft.bond.helpers.ArgumentHelper;
 
-import java.io.ByteArrayInputStream;
```
sigansky: > java.io.ByteArrayInputStream; [](start = 6, length = 30)

It's weird this was here. Thanks for cleanup.

java/core/src/test/java/com/microsoft/bond/testmodel/A.java
```diff
@@ -285,6 +284,11 @@ protected final void deserializeStructFields(TaggedDeserializationContext contex
         }
 
         @Override
+        protected final void deserializeStructFields(UntaggedDeserializationContext context, A<X, Y> value) throws IOException {
+        }
+
+
```
sigansky: Minor: single line

java/core/src/test/java/com/microsoft/bond/testmodel/A.java
```diff
@@ -285,6 +284,11 @@ protected final void deserializeStructFields(TaggedDeserializationContext contex
         }
 
         @Override
+        protected final void deserializeStructFields(UntaggedDeserializationContext context, A<X, Y> value) throws IOException {
+        }
```
sigansky: Can you please add deserialize calls?

sigansky: Also to B/C/D/E ?

---
In reply to: [130750811](https://github.com/Microsoft/bond/pull/542#discussion_r130750811) [](ancestors = 130750811)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -301,6 +301,18 @@ makeStructBondTypeMember_deserializeStructFields declName declParams structField
...
+-- given struct class name, generic type parameters, and struct fields, builds text for implementation of
+-- the StructBondTypeImpl.deserializeStructFields method for UntaggedProtocolReaders
+makeStructBondTypeMember_deserializeStructFields_untagged :: String -> [TypeParam] -> [Field] -> Text
+makeStructBondTypeMember_deserializeStructFields_untagged declName declParams structFields = [lt|
+        @Override
+        protected final void deserializeStructFields(#{methodParamDecl}) throws java.io.IOException {#{newlineBeginSep 3 serializeField structFields}
+        }|]
+            where
+                methodParamDecl = [lt|com.microsoft.bond.BondType.UntaggedDeserializationContext context, #{typeNameWithParams declName declParams} value|]
+                serializeField Field {..} = [lt|this.#{fieldName}.deserialize(context);|]
```
sigansky: >this.#{fieldName} [](start = 48, length = 17)

Need to assign the result to a field in the value:

value.<field> = this.<field>.deserialize(context)

sigansky:
>             StructBondType.StructField<T> field) throws IOException;

I assume you'll also add a version of this method for untagged?

---
Refers to: java/core/src/main/java/com/microsoft/bond/BondType.java:266 in f4a9ff2. [](commit_id = f4a9ff2c6f1c1f271b6a57c1c27ccb524890db62, deletion_comment = False)

java/core/src/main/java/com/microsoft/bond/BondedBondType.java
```diff
@@ -106,6 +106,14 @@ protected final void serializeValue(SerializationContext context, Bonded<TStruct
     }
 
     @Override
+    protected final Bonded<TStruct> deserializeValue(UntaggedDeserializationContext context) throws IOException {
+        TStruct value = this.valueType.deserializeValue(context);
+
+        // TODO: complete deserialization story for bonded
+        throw new UnsupportedOperationException();
```
sigansky: I assume you'll complete this after I push my change and you merge?

java/core/src/main/java/com/microsoft/bond/NullableBondType.java
```diff
@@ -120,6 +120,20 @@ protected final TValue deserializeValue(TaggedDeserializationContext context) th
     }
 
     @Override
+    protected final TValue deserializeValue(UntaggedDeserializationContext context) throws IOException {
+        final int count = context.reader.readContainerBegin();
+        if (count == 0) {
+            return null;
+        } else if (count == 1) {
```
sigansky: This makes me wonder if the code above for tagged protocols should also check for empty lists in payload. Such list would not be produced by serialization (since empty lists are omitted) but if we do get this in payload, throwing an exception complaining about "multiple elements" for an empty list sounds wrong.

What are your thoughts on this?

java/core/src/main/java/com/microsoft/bond/NullableBondType.java
```diff
@@ -120,6 +120,20 @@ protected final TValue deserializeValue(TaggedDeserializationContext context) th
...
+            return this.valueType.deserializeValue(context);
+        } else {
+            // throws
+            Throw.raiseNullableListValueHasMultipleElementsDeserializationError(this.getFullName());
+            return null;
+        }
```
sigansky: Please follow the pattern used in the tagged version (specifically a single return statement). Please note that you are not calling readContainerEnd here, which is a bug.

java/core/src/main/java/com/microsoft/bond/NullableBondType.java
```diff
@@ -120,6 +120,20 @@ protected final TValue deserializeValue(TaggedDeserializationContext context) th
     }
 
     @Override
+    protected final TValue deserializeValue(UntaggedDeserializationContext context) throws IOException {
+        final int count = context.reader.readContainerBegin();
+        if (count == 0) {
+            return null;
+        } else if (count == 1) {
+            return this.valueType.deserializeValue(context);
```
sigansky: >return [](start = 12, length = 6)

readContainerEnd

java/core/src/main/java/com/microsoft/bond/ListBondType.java
```diff
@@ -102,6 +103,11 @@ public final boolean isGenericType() {
         return new LinkedList<TElement>();
     }
 
+    private static <TElement> List<TElement> newDefaultValue(int initialCapacity) {
+        // custom initial capacity to match element count when deserializing
+        return new ArrayList<TElement>(initialCapacity);
+    }
```
sigansky: Why are you using ArrayList for list? We've decided to use ArrayList for vectors and LinkedList for lists.

java/core/src/main/java/com/microsoft/bond/ListBondType.java
```diff
@@ -148,6 +154,21 @@ protected final void serializeValue(SerializationContext context, List<TElement>
...
+        List<TElement> value = newDefaultValue(count);
+        for (int i = 0; i < count; ++i) {
+            try {
+                TElement element = this.elementType.deserializeValue(context);
+                value.add(element);
+            } catch (InvalidBondDataException e) {
+                Throw.raiseListContainerElementSerializationError(true, false, this.getFullName(), i, e, null);
+            }
+        }
+        return value;
```
sigansky: >return [](start = 8, length = 6)

readContainerEnd

java/core/src/main/java/com/microsoft/bond/ListBondType.java
```diff
@@ -148,6 +154,21 @@ protected final void serializeValue(SerializationContext context, List<TElement>
     }
 
     @Override
+    protected final List<TElement> deserializeValue(UntaggedDeserializationContext context) throws IOException {
+        final int count = context.reader.readContainerBegin();
+        List<TElement> value = newDefaultValue(count);
```
sigansky: LinkedList is unable to reserve capacity, so need to start empty.

java/core/src/main/java/com/microsoft/bond/VectorBondType.java
```diff
@@ -154,6 +154,21 @@ protected final void serializeValue(SerializationContext context, List<TElement>
...
+        List<TElement> value = newDefaultValue(count);
+        for (int i = 0; i < count; ++i) {
+            try {
+                TElement element = this.elementType.deserializeValue(context);
+                value.add(element);
+            } catch (InvalidBondDataException e) {
+                Throw.raiseListContainerElementSerializationError(true, false, this.getFullName(), i, e, null);
+            }
+        }
+        return value;
```
sigansky: >return [](start = 8, length = 6)

readContainerEnd

java/core/src/main/java/com/microsoft/bond/SetBondType.java
```diff
@@ -102,6 +102,12 @@ public final boolean isGenericType() {
         return new HashSet<TElement>();
     }
 
+    private static <TElement> Set<TElement> newDefaultValue(int initialCapacity) {
+        // custom initial capacity to match element count when deserializing
+        return new HashSet<TElement>(initialCapacity);
+    }
+
+
```
sigansky: Single space please.

java/core/src/main/java/com/microsoft/bond/SetBondType.java
```diff
@@ -148,6 +154,21 @@ protected final void serializeValue(SerializationContext context, Set<TElement>
...
+        Set<TElement> value = newDefaultValue(count);
+        for (int i = 0; i < count; ++i) {
+            try {
+                TElement element = this.elementType.deserializeValue(context);
+                value.add(element);
+            } catch (InvalidBondDataException e) {
+                Throw.raiseListContainerElementSerializationError(true, true, this.getFullName(), i, e, null);
+            }
+        }
+        return value;
```
sigansky: >return [](start = 8, length = 6)

readContainerEnd

java/core/src/main/java/com/microsoft/bond/SetBondType.java
```diff
@@ -148,6 +154,21 @@ protected final void serializeValue(SerializationContext context, Set<TElement>
     }
 
     @Override
+    protected final Set<TElement> deserializeValue(UntaggedDeserializationContext context) throws IOException {
+        final int count = context.reader.readContainerBegin();
+        Set<TElement> value = newDefaultValue(count);
```
sigansky: This doesn't map to capacity directly, as the latter is the size of hash table. Ideally, you should *compute* the initial capacity based on the expected size (which is the count) and load factor (default is 0.75).

Please see HashMap documentation (which is used to implement HashSet):

An instance of HashMap has two parameters that affect its performance: initial capacity and load factor. The capacity is the number of buckets in the hash table, and the initial capacity is simply the capacity at the time the hash table is created. The load factor is a measure of how full the hash table is allowed to get before its capacity is automatically increased. When the number of entries in the hash table exceeds the product of the load factor and the current capacity, the hash table is rehashed (that is, internal data structures are rebuilt) so that the hash table has approximately twice the number of buckets.

As a general rule, the default load factor (.75) offers a good tradeoff between time and space costs. Higher values decrease the space overhead but increase the lookup cost (reflected in most of the operations of the HashMap class, including get and put). The expected number of entries in the map and its load factor should be taken into account when setting its initial capacity, so as to minimize the number of rehash operations. If the initial capacity is greater than the maximum number of entries divided by the load factor, no rehash operations will ever occur.

sigansky: So, ceiling(count / loadFactor) could be a good initial capacity which can prevent rehashes while the set is being initialized.

---
In reply to: [130754748](https://github.com/Microsoft/bond/pull/542#discussion_r130754748) [](ancestors = 130754748)

java/core/src/main/java/com/microsoft/bond/MapBondType.java
```diff
@@ -112,6 +112,12 @@ public final boolean isGenericType() {
         return new HashMap<TKey, TValue>();
     }
 
+    private static <TKey, TValue> Map<TKey, TValue> newDefaultValue(int initialCapacity) {
+        // custom initial capacity to match element count when deserializing
+        return new HashMap<TKey, TValue>(initialCapacity);
+    }
+
+
     @Override
```
sigansky: Single space please.

java/core/src/main/java/com/microsoft/bond/MapBondType.java
```diff
@@ -181,6 +187,28 @@ protected final void serializeValue(SerializationContext context, Map<TKey, TVal
...
+            }
+            TValue mapEntryValue = null;
+            try {
+                mapEntryValue = this.valueType.deserializeValue(context);
+            } catch (InvalidBondDataException e) {
+                Throw.raiseMapContainerElementSerializationError(true, this.getFullName(), i, mapEntryKey, e, null);
+            }
+            value.put(mapEntryKey, mapEntryValue);
+        }
+        return value;
```
sigansky: >return [](start = 8, length = 6)

readContainerEnd

java/core/src/main/java/com/microsoft/bond/MapBondType.java
```diff
@@ -181,6 +187,28 @@ protected final void serializeValue(SerializationContext context, Map<TKey, TVal
     }
 
     @Override
+    protected final Map<TKey, TValue> deserializeValue(UntaggedDeserializationContext context) throws IOException {
+        final int count = context.reader.readContainerBegin();
+        Map<TKey, TValue> value = newDefaultValue(count);
```
sigansky: Please see the same comment I have for Set. I suggest (int)(count / 0.75 + 1) as initial capacity.

java/core/src/main/java/com/microsoft/bond/StructBondType.java
```diff
@@ -1828,10 +1956,8 @@ public final void serialize(
             FloatBondType.serializePrimitiveSomethingField(context, value, this);
         }
 
-        public final SomethingFloat deserialize(
-                TaggedDeserializationContext context, boolean wasAlreadyDeserialized) throws IOException {
-            this.verifyFieldWasNotYetDeserialized(wasAlreadyDeserialized);
-            return FloatBondType.deserializePrimitiveSomethingField(context, this);
```
sigansky: Looks like a method was accidentally removed.

java/core/src/main/java/com/microsoft/bond/protocol/UntaggedProtocolReader.java
```diff
@@ -0,0 +1,192 @@
...
+
+package com.microsoft.bond.protocol;
+
+import java.io.IOException;
+
+/**
+ * Reads from serialized payload encoded using an untagged protocol.
+ */
+public interface UntaggedProtocolReader {
+    /**
```
sigansky: Please add a line separator.

java/core/src/main/java/com/microsoft/bond/protocol/UntaggedProtocolReader.java
```diff
@@ -0,0 +1,192 @@
...
+     * Read an int8.
+     *
+     * @return the value
+     */
+    byte readInt8() throws IOException;
+
+    /**
+     * Skip an int8.
+     */
+    void skipInt8() throws IOException;
```
sigansky: >skipInt8 [](start = 9, length = 8)

Who calls skip* methods and why?

sigansky: Is this for Bonded?

---
In reply to: [130756104](https://github.com/Microsoft/bond/pull/542#discussion_r130756104) [](ancestors = 130756104)

java/core/src/test/java/com/microsoft/bond/testmodel/A.java
tstein: Not sure what you mean. I don't want to maintain these files by hand now that we have golden files set up.

java/core/src/main/java/com/microsoft/bond/BondedBondType.java
tstein: Yup.

java/core/src/main/java/com/microsoft/bond/NullableBondType.java
tstein: It probably should. I don't think omitting fields is ever mandatory, so we need to be able to handle nullables that come as explicit lists of length 0.

chwarr: For example, some of the C++ protocol implementation can be configured to never omit fields.

java/core/src/main/java/com/microsoft/bond/ListBondType.java
tstein: Confusion. Fixed.

java/core/src/main/java/com/microsoft/bond/NullableBondType.java
tstein: Done for all of these.

java/core/src/main/java/com/microsoft/bond/SetBondType.java
tstein: Attempting to tune this is well outside the scope of this PR. Changed back to default constructors.

java/core/src/main/java/com/microsoft/bond/StructBondType.java
tstein: Thanks.

java/core/src/main/java/com/microsoft/bond/protocol/UntaggedProtocolReader.java
tstein: Yes.

tstein:
The comment "I assume you'll also add a version of this method for untagged?" in reference to `deserializeField` methods got added here, but I couldn't find any way to reply to it.

I don't think `deserializeField(Untagged)` has a reason to exist. AFAICT, `deserializeField(Tagged)` deals with the tag, then hands off to `deserializeValue(Tagged)`.

java/core/src/test/java/com/microsoft/bond/testmodel/A.java
sigansky: Can you please delete them, if you don't want to maintain them?

---
In reply to: [131010024](https://github.com/Microsoft/bond/pull/542#discussion_r131010024) [](ancestors = 131010024)

tstein: Done.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -310,6 +310,18 @@ makeStructBondTypeMember_deserializeStructFields declName declParams structField
                 verifyField Field {..} = [lt|this.#{fieldName}.verifyDeserialized(__has_#{fieldName});|]
 
 
+-- given struct class name, generic type parameters, and struct fields, builds text for implementation of
+-- the StructBondTypeImpl.deserializeStructFields method for UntaggedProtocolReaders
+makeStructBondTypeMember_deserializeStructFields_untagged :: String -> [TypeParam] -> [Field] -> Text
+makeStructBondTypeMember_deserializeStructFields_untagged declName declParams structFields = [lt|
+        @Override
+        protected final void deserializeStructFields(#{methodParamDecl}) throws java.io.IOException {#{newlineBeginSep 3 serializeField structFields}
```
sigansky: >serializeField [](start = 121, length = 14)

deserializeField

cpp/inc/bond/protocol/simple_binary.h
```diff
@@ -36,8 +36,8 @@
                      string, wstring        | count | characters |
                                             '-------'------------'
 
-                           count            variable encoded uint32 count of 1-byte (for
-                                            string) or 2-byte (for wstring) Unicode code
+                           count            uint32 count of 1-byte (for string)
+                                            or 2-byte (for wstring) Unicode code
                                             units (variable encoded in v2)
```
sigansky: Should this be updated in other places? For example, SimpleBinary.cs in the .NET version has the same comment. 

Also, what is the process to update the C++ API reference pages?

java/core/src/main/java/com/microsoft/bond/BondedBondType.java
```diff
@@ -112,6 +112,14 @@ protected final void serializeValue(SerializationContext context, Bonded<TStruct
...
```
sigansky: This is still active, please let me know when it's ready.

java/core/src/main/java/com/microsoft/bond/MapBondType.java
```diff
@@ -181,6 +181,29 @@ protected final void serializeValue(SerializationContext context, Map<TKey, TVal
     }
 
     @Override
+    protected final Map<TKey, TValue> deserializeValue(UntaggedDeserializationContext context) throws IOException {
+        final int count = context.reader.readContainerBegin();
+        Map<TKey, TValue> value = newDefaultValue();
```
sigansky: >value [](start = 26, length = 5)

final? :-)

sigansky:
>         Map<TKey, TValue> value = newDefaultValue();

final

---
Refers to: java/core/src/main/java/com/microsoft/bond/MapBondType.java:163 in 4b5974a. [](commit_id = 4b5974a745cb95c849a7f8173ff7afb9477aeefc, deletion_comment = False)

java/core/src/main/java/com/microsoft/bond/MapBondType.java
sigansky: >value [](start = 26, length = 5)

final

sigansky: >value [](start = 26, length = 5)

final could be useful here

java/core/src/main/java/com/microsoft/bond/ListBondType.java
```diff
@@ -148,6 +148,22 @@ protected final void serializeValue(SerializationContext context, List<TElement>
     }
 
     @Override
+    protected final List<TElement> deserializeValue(UntaggedDeserializationContext context) throws IOException {
+        final int count = context.reader.readContainerBegin();
+        List<TElement> value = newDefaultValue();
```
sigansky: >List<TElement> value [](start = 8, length = 20)

final could be useful here

java/core/src/main/java/com/microsoft/bond/SetBondType.java
```diff
@@ -148,6 +148,22 @@ protected final void serializeValue(SerializationContext context, Set<TElement>
     }
 
     @Override
+    protected final Set<TElement> deserializeValue(UntaggedDeserializationContext context) throws IOException {
+        final int count = context.reader.readContainerBegin();
+        Set<TElement> value = newDefaultValue();
```
sigansky: >value [](start = 22, length = 5)

final could be useful here

java/core/src/main/java/com/microsoft/bond/VectorBondType.java
```diff
@@ -154,6 +154,22 @@ protected final void serializeValue(SerializationContext context, List<TElement>
     }
 
     @Override
+    protected final List<TElement> deserializeValue(UntaggedDeserializationContext context) throws IOException {
+        final int count = context.reader.readContainerBegin();
+        List<TElement> value = newDefaultValue(count);
```
sigansky: >value [](start = 23, length = 5)

final could be useful here

java/core/src/main/java/com/microsoft/bond/StructBondType.java
```diff
@@ -344,6 +358,16 @@ protected final TStruct deserializeValue(TaggedDeserializationContext context) t
         return value;
     }
 
+    @Override
+    protected final TStruct deserializeValue(UntaggedDeserializationContext context) throws IOException {
+        TStruct value = this.newDefaultValue();
+        if (this.baseStructType != null) {
+            this.baseStructType.deserializeValueAsBase(context, value);
+        }
+        this.deserializeStructFields(context, value);
```
sigansky: Observe that due to absence of readStructBegin and readStructEnd, the logic to deserialize a struct and a base is the same. This method can just call deserializeValueAsBase (code is identical).

java/core/src/main/java/com/microsoft/bond/StructBondType.java
```diff
@@ -13,12 +13,16 @@
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
+import static com.microsoft.bond.Something.wrap;
```
sigansky: I am not a fan of this language feature, as it can easily make the code hard to read. Further, half of methods call Something.wrap explicitly and others rely on the imported method. Can you please make it consistent?

java/core/src/main/java/com/microsoft/bond/StructBondType.java
```diff
@@ -2053,6 +2197,10 @@ public final void serialize(
             this.verifyFieldWasNotYetDeserialized(wasAlreadyDeserialized);
             return this.fieldType.deserializeSomethingField(context, this);
         }
+
+        public final SomethingObject<String> deserialize(UntaggedDeserializationContext context) throws IOException {
+            return SomethingObject.wrap(this.fieldType.deserializeValue(context));
```
sigansky: >SomethingObject.wrap [](start = 19, length = 20)

SomethingObject inherits wrap methods from Something class. I suggest using the actual declaring class when you call a static method. 

Is there a reason you didn't rely on the import of the "wrap" methods and need to add the class name?

java/core/src/main/java/com/microsoft/bond/protocol/SimpleBinaryWriter.java
```diff
@@ -0,0 +1,238 @@
...
+        this.writer.writeInt16((short) MAGIC.value);
+        this.writer.writeInt16(this.protocolVersion);
+    }
+
+    @Override
+    public void writeFieldOmitted(
+            final BondDataType type, final int id, final FieldMetadata metadata) throws IOException {
+        assert !metadata.isDefaultNothing();
+
+        switch (type.value) {
```
sigansky: >switch (type.value) { [](start = 8, length = 21)

Can you please add a comment like that in the .NET version:

"// Simple doesn't support omitting fields so instead we write the default value"

java/core/src/main/java/com/microsoft/bond/protocol/SimpleBinaryWriter.java
```diff
@@ -0,0 +1,238 @@
...
+                writeInt16((Short) metadata.getDefaultValue());
+                break;
+            case BondDataType.Values.BT_INT32:
+                writeInt32((Integer) metadata.getDefaultValue());
+                break;
+            case BondDataType.Values.BT_INT64:
+                writeInt64((Long) metadata.getDefaultValue());
+                break;
+            case BondDataType.Values.BT_UINT8:
+                writeInt8((Byte) metadata.getDefaultValue());
```
sigansky: >writeInt8 [](start = 16, length = 9)

writeUInt8

java/core/src/main/java/com/microsoft/bond/protocol/SimpleBinaryWriter.java
```diff
@@ -0,0 +1,238 @@
...
+                break;
+            case BondDataType.Values.BT_UINT16:
+                writeInt16((Short) metadata.getDefaultValue());
```
sigansky: >writeInt16 [](start = 16, length = 10)

writeUInt16

java/core/src/main/java/com/microsoft/bond/protocol/SimpleBinaryWriter.java
```diff
@@ -0,0 +1,238 @@
...
+                break;
+            case BondDataType.Values.BT_UINT32:
+                writeInt32((Integer) metadata.getDefaultValue());
```
sigansky: >writeInt32 [](start = 16, length = 10)

writeUInt32

java/core/src/main/java/com/microsoft/bond/protocol/SimpleBinaryWriter.java
```diff
@@ -0,0 +1,238 @@
...
+                break;
+            case BondDataType.Values.BT_UINT64:
+                writeInt64((Long) metadata.getDefaultValue());
```
sigansky: >writeInt64 [](start = 16, length = 10)

writeUInt64

tstein:
Simple now passes compat for v1 and v2. I'm not adding anything else to this PR - please review it as is.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: Done, thanks.

cpp/inc/bond/protocol/simple_binary.h
tstein: Yeah. Taking a separate task for this.

java/core/src/main/java/com/microsoft/bond/BondedBondType.java
tstein: Now implemented.

java/core/src/main/java/com/microsoft/bond/MapBondType.java
tstein: Done, and some more.

java/core/src/main/java/com/microsoft/bond/StructBondType.java
tstein: I don't remember doing this intentionally. IDE auto-cleanup funkiness, maybe?

Changed all instances to `Something.wrap()`.

java/core/src/main/java/com/microsoft/bond/protocol/SimpleBinaryWriter.java
tstein: Took as separate task.

java/core/src/main/java/com/microsoft/bond/protocol/SimpleBinaryWriter.java
tstein: Done.

cpp/inc/bond/protocol/simple_binary.h
chwarr: The C++ API reference pages are [automatically rebuilt](https://github.com/Microsoft/bond/blob/5391636d5b4a8c59981527822841649a292d6be6/appveyor.yml#L232-L259) when a push to master is made. They're generated from doc comments in [the C++ files using Doxygen](https://github.com/Microsoft/bond/blob/5391636d5b4a8c59981527822841649a292d6be6/doc/CMakeLists.txt#L129-L153).

java/compat/src/main/java/com/microsoft/bond/compat/CompatDriver.java
```diff
@@ -33,34 +32,69 @@ public static void main(final String[] args) throws IOException {
...
+            case "compact2":
+                compat = deserializer.deserialize(taggedReader);
+                break;
+            case "simple":
+            case "simple2":
+                compat = deserializer.deserialize(untaggedReader);
+                break;
+            default:
+                System.err.println("Unsupported input protocol: " + fromProtocol);
+                System.exit(STATUS_FAILURE); return; // javac doesn't realize System.exit() never returns.
```
sigansky: This check is already made earlier. I suggest an alternative approach: define boolean isTaggedFromProtocol and set it in the first switch, then this just becomes a small if/else. Further, the boolean may be replaced by check for taggedReader != null.

java/core/src/main/java/com/microsoft/bond/StructBondType.java
```diff
@@ -344,8 +357,18 @@ protected final TStruct deserializeValue(TaggedDeserializationContext context) t
         return value;
     }
 
+    @Override
+    protected final TStruct deserializeValue(UntaggedDeserializationContext context) throws IOException {
+        TStruct value = this.newDefaultValue();
+        if (this.baseStructType != null) {
+            this.baseStructType.deserializeValueAsBase(context, value);
+        }
+        this.deserializeStructFields(context, value);
```
sigansky: I already made a comment that this block is identical to deserializeValueAsBase and you can just call that method here. It's different for tagged since need to call begin/end methods.

java/core/src/main/java/com/microsoft/bond/protocol/ProtocolWriter.java
```diff
@@ -13,6 +13,12 @@
 public interface ProtocolWriter {
 
     /**
+     * @return true if bonded fields in this protocol are marshaled; false if
+     * they are serialized in the same protocol
+     */
+    boolean usesMarshaledBonded();
```
sigansky: I don't understand why this method is needed here. The .NET version doesn't have it.

java/core/src/main/java/com/microsoft/bond/protocol/SimpleBinaryWriter.java
sigansky: What os the task? Adding a comment?

---
In reply to: [132343301](https://github.com/Microsoft/bond/pull/542#discussion_r132343301) [](ancestors = 132343301)

java/core/src/main/java/com/microsoft/bond/protocol/SimpleBinaryWriter.java
```diff
@@ -0,0 +1,250 @@
...
+    @Override
+    public void writeVersion() throws IOException {
+        this.writer.writeInt16((short) MAGIC.value);
+        this.writer.writeInt16(this.protocolVersion);
+    }
+
+    @Override
+    public void writeFieldOmitted(
+            final BondDataType type, final int id, final FieldMetadata metadata) throws IOException {
+        assert !metadata.isDefaultNothing();
```
sigansky: > assert !metadata.isDefaultNothing(); [](start = 7, length = 37)

That's what I was saying this morning: an assert in library code. Are you going to remove it?

java/core/src/main/java/com/microsoft/bond/BondedBondType.java
```diff
@@ -112,6 +128,28 @@ protected final void serializeValue(SerializationContext context, Bonded<TStruct
     }
 
     @Override
+    protected final Bonded<TStruct> deserializeValue(UntaggedDeserializationContext context) throws IOException {
+        // Bonded fields are always marshaled and prefixed with a fixed-width length.
+        final int bondedCount = context.reader.readUInt32();
```
sigansky: >bondedCount [](start = 18, length = 11)

NIT: "length" is more appropriate than "count".

java/core/src/main/java/com/microsoft/bond/protocol/ProtocolWriter.java
chwarr: Is has [similar code in a different place](https://github.com/Microsoft/bond/blob/2bdf48c618d1fb61de5488af0ed5d6988fcacb7e/cs/src/core/expressions/ProtocolWriter.cs#L162-L173) due its architecture of generating expressions.

java/core/src/main/java/com/microsoft/bond/StructBondType.java
tstein: Talked offline - `deserializeValueAsBase(Untagged)` is now `deserializeValue(Untagged)` and fits nicely.

java/core/src/main/java/com/microsoft/bond/protocol/SimpleBinaryWriter.java
tstein: Talked offline - changed to an if/throw.

java/core/src/main/java/com/microsoft/bond/BondedBondType.java
tstein: Agreed; we use count a lot. I'll change this one.

Issue #538, opened 2017-07-23
Compile error with GCC 


ramon-garcia:
While building the master version with GCC, the following error appeared:
`In file included from /devel/bond/cpp/inc/bond/core/parser.h:8:0,
                 from /devel/bond/cpp/inc/bond/core/apply.h:9,
                 from /devel/bond/cpp/inc/bond/core/bond.h:7,
                 from /devel/bond/build/cpp/bond/core/bond_apply.h:16,
                 from /devel/bond/build/cpp/bond/core/bond_apply.cpp:13:
/devel/bond/cpp/inc/bond/core/value.h: In member function ‘void bond::DeserializeElement(X&, const I&, const T&)::Deserialize::operator()(typename bond::element_type<T>::type&)’:
/devel/bond/cpp/inc/bond/core/value.h:644:57: error: non-template type ‘Deserialize’ used as a template
             this->element.template Deserialize<Protocols>(e);
                                                         ^
`

This looks like a limitation of GCC, though I tested it with GCC 4.9, 6.3.0 and 7.1

Attached a workaround.

[fix_gcc_core_value.diff.txt](https://github.com/Microsoft/bond/files/1168381/fix_gcc_core_value.diff.txt)

ramon-garcia:
I confirm that this is a bug in GCC.

A testcase fails with GCC and succeeds with Clang (checked by testing in https://gcc.godbolt.org/ )

chwarr:
Thanks for the heads up.
@ara-ayvazyan, you're C++ standards knowledge if better than mine. Do you have time to come up with a cross-compiler, cross-version fix, even if there's a bug in GCC here?

(GCC *may* be following the standard, and Clang and MSVC are missing this one. Or, it may very well be a bug in GCC...)

ara-ayvazyan:
Sure, I'll take a look.

ara-ayvazyan:
Here is a shorter repro:
``` C++
template <typename T>
void foo()
{
    struct X
    {
        void bar()
        {
            this->var.template X<T>();
        }

        T var;
    };
}
```
I'm not sure which compiler is correct (will need to contact some C++ language lawyer 😃), but this can be resolved by adding `X::` just before the `template` keyword:
``` C++
this->var.X::template X<T>();
```
Renaming the `struct X` is probably safer.

ramon-garcia:
@chwarr Yes, I posted the bug in GCC bug tracker.

[...]

PR #543, opened 2017-08-01
[c++] Fix G++ 5.4 errors

chwarr:
* Deserialize was used as the name for both a struct and a member
  function. This causes some versions of G++ to emit errors, so rename
  the helper struct. Fixes https://github.com/Microsoft/bond/issues/538
* The Protocols template argument could not be deduced in one place
  where bond::SelectProtocolAndApply was called. It is now explicitly
  specified.


chwarr:
@ara-ayvazyan, @ramon-garcia: this should fix the G++ errors. I've tested this against G++ 5.4, and the CI builds will get Clang 3.5.0 and MSVC.

PR #545, opened 2017-08-04
java core: Codegen .equals() and .hashCode().

tstein:
Needs significant testing, but if it looks vaguely correct, might be a good idea to merge it early to unblock work on examples.


compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -373,6 +373,41 @@ public #{declName}(com.microsoft.bond.StructBondType<#{declName}<#{paramList}>>
...
+        final #{declName} other = (#{declName}) o;
+        #{newlineSep 2 compareField fields}
+        return true;
+    }
+|]
+    where
+        compareField Field {..} = [lt|if (!(#{equals_member fieldType fieldName})) return false;|]
+        equals_member t f
+            | isJavaPrimitive t = [lt|this.#{f} == other.#{f}|]
+            | otherwise         = [lt|(this.#{f} == null && other.#{f} == null) || (this.#{f} != null && this.#{f}.equals(other.#{f}))|]
```
chwarr: What does the implementation of `.equals()` do for bonded types?

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -373,6 +373,41 @@ public #{declName}(com.microsoft.bond.StructBondType<#{declName}<#{paramList}>>
...
+        if (!(o instanceof #{declName})) return false;
+        final #{declName} other = (#{declName}) o;
+        #{newlineSep 2 compareField fields}
+        return true;
+    }
+|]
+    where
+        compareField Field {..} = [lt|if (!(#{equals_member fieldType fieldName})) return false;|]
+        equals_member t f
+            | isJavaPrimitive t = [lt|this.#{f} == other.#{f}|]
```
chwarr: I would expect == to be the wrong check for float/double. If both at NaN, won't this return false still?

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -373,6 +373,41 @@ public #{declName}(com.microsoft.bond.StructBondType<#{declName}<#{paramList}>>
...
+        int result = 17;
+        #{newlineSep 2 hashField fields}
+        return result;
+    }
+|]
+    where
+        hashField Field {..} = [lt|result = (31 * result) + #{hash_member fieldType fieldName};|]
+        hash_member BT_Bool f   = [lt|(#{f} ? 0 : 1)|]
+        hash_member t f
+            | isJavaPrimitive t = [lt|(int) #{f}|]
```
chwarr: Casting a float/double to an int for hashing purposes doesn't seem like it will give good distribution, particularly if, say, the float is representing numbers in \[0, 1).

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -373,6 +373,41 @@ public #{declName}(com.microsoft.bond.StructBondType<#{declName}<#{paramList}>>
...
+object_hashCode :: [Field] -> Text
+object_hashCode fields = [lt|@Override
+    public int hashCode() {
+        int result = 17;
+        #{newlineSep 2 hashField fields}
+        return result;
+    }
+|]
+    where
+        hashField Field {..} = [lt|result = (31 * result) + #{hash_member fieldType fieldName};|]
```
chwarr: An algorithm like this will probably mix things better:

```
result += #{hash_member fieldType fieldName};
// Constant used to redistribute the bits. Should have about an equal number of 1s and 0s, and needs to be odd so the low bit isn't always zero.
result *= 0xeadbeef;
result ^= result >> 16
```

compiler/src/Language/Bond/Codegen/Java/Util.hs
```diff
@@ -9,6 +9,7 @@ module Language.Bond.Codegen.Java.Util
     , modifierConstantName
     , isPrimitiveNonEnumBondType
     , isPrimitiveBondType
+    , isJavaPrimitive
```
sigansky: >isJavaPrimitive [](start = 6, length = 15)

suggest name: isJavaPrimitiveType

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -373,6 +373,41 @@ public #{declName}(com.microsoft.bond.StructBondType<#{declName}<#{paramList}>>
...
+        if (o == null) return false;
+        if (!(o instanceof #{declName})) return false;
+        final #{declName} other = (#{declName}) o;
+        #{newlineSep 2 compareField fields}
+        return true;
+    }
+|]
+    where
+        compareField Field {..} = [lt|if (!(#{equals_member fieldType fieldName})) return false;|]
+        equals_member t f
```
sigansky: >equals_member [](start = 8, length = 13)

Style: please do not mix camel case and underscores

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
sigansky: NaN == NaN is always false. Further, 0D == -0D is always true. There is FloatingPointHelper class that compares bitwise, which is what we use when serializing.

Java has different comparison implementations for primitive double and object Double types; the latter is comparing bits. We should keep this consistent and compare primitives using bits. Note that collections already do the right comparison since elements are objects.

---
In reply to: [131460893](https://github.com/Microsoft/bond/pull/545#discussion_r131460893) [](ancestors = 131460893)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
sigansky: It's not in this PR; I believe it should be.

Ted, can you override equals/hashCode in Bonded? I suppose for starters we consider them equal if the underlying types are equal and if they are deserialized to equal objects.

---
In reply to: [131460787](https://github.com/Microsoft/bond/pull/545#discussion_r131460787) [](ancestors = 131460787)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -373,6 +373,41 @@ public #{declName}(com.microsoft.bond.StructBondType<#{declName}<#{paramList}>>
...
 
 
+-- builds text for Object.equals(Object) override
+object_equals :: String -> [Field] -> Text
+object_equals declName fields = [lt|@Override
+    public boolean equals(Object o) {
+        if (o == null) return false;
+        if (!(o instanceof #{declName})) return false;
+        final #{declName} other = (#{declName}) o;
+        #{newlineSep 2 compareField fields}
```
sigansky: >#{newlineSep 2 compareField fields} [](start = 8, length = 35)

If there are no fields, this will produce a blank line. Can you move this to the previous or next line?

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -373,6 +373,41 @@ public #{declName}(com.microsoft.bond.StructBondType<#{declName}<#{paramList}>>
...
+        return true;
```
sigansky: >return true; [](start = 8, length = 12)

Need to also call the base class.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -373,6 +373,41 @@ public #{declName}(com.microsoft.bond.StructBondType<#{declName}<#{paramList}>>
...
+    }
+|]
+    where
+        compareField Field {..} = [lt|if (!(#{equals_member fieldType fieldName})) return false;|]
```
sigansky: >!(#{equals_member fieldType fieldName}) [](start = 42, length = 39)

You can simplify this and avoid extra negation, which makes the code harder to read and is unnecessary.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -373,6 +373,41 @@ public #{declName}(com.microsoft.bond.StructBondType<#{declName}<#{paramList}>>
...
+            | otherwise         = [lt|(this.#{f} == null && other.#{f} == null) || (this.#{f} != null && this.#{f}.equals(other.#{f}))|]
+
+
+-- builds text for Object.hashCode() override
+object_hashCode :: [Field] -> Text
+object_hashCode fields = [lt|@Override
+    public int hashCode() {
+        int result = 17;
+        #{newlineSep 2 hashField fields}
+        return result;
```
sigansky: >return result; [](start = 8, length = 14)

Need to also call the base class.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
sigansky: >lt|(int) #{f}| [](start = 35, length = 14)

This ignores higher bytes in long and double.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -373,6 +373,41 @@ public #{declName}(com.microsoft.bond.StructBondType<#{declName}<#{paramList}>>
...
+    }
+|]
+    where
+        hashField Field {..} = [lt|result = (31 * result) + #{hash_member fieldType fieldName};|]
+        hash_member BT_Bool f   = [lt|(#{f} ? 0 : 1)|]
+        hash_member t f
+            | isJavaPrimitive t = [lt|(int) #{f}|]
+            | otherwise         = [lt|#{f}.hashCode()|]
```
sigansky: >{f}.hashCode() [](start = 39, length = 14)

Missing null check.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -373,6 +373,41 @@ public #{declName}(com.microsoft.bond.StructBondType<#{declName}<#{paramList}>>
...
+
+
```
sigansky: Overall design suggestion: please consider being consistent with how other generated methods are implemented and add valueEquals and valueHashCode methods to BondType. Then every line looks the same and codegen is greatly simplified. This also allows to address different implementation for double/float in equals (and long/double/bool in hashCode) and future improvements in Bond library, such as more efficient collection comparisons (the defaults ones are too general).

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: It sucks that Java is inconsistent, but I don't think it's our place to try to fix that.

If I made a struct with a field in it that codegen'd to a `double`, and I tried to compare two of my structs in Java, I would expect them to compare like `double`s in Java.

sigansky: I don't have strong opinion either way, as I am not convinced that we even need deep comparison and hash code on generated types. I wasn't there where this was decided; let's reiterate on this today.

If C++/C# don't have this feature, let's not add this to Java. It appears that C# doesn't have it.

If C++/C# have this feature, let's keep it consistent.

---
In reply to: [131471537](https://github.com/Microsoft/bond/pull/545#discussion_r131471537) [](ancestors = 131471537)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: chwarr: Currently, `Object.equals()`, which will never be true.

I'm not convinced that defining `.equals()` is a good idea. Can a `Bonded<T>` and a `Bonded<U>` that's been sliced into a `Bonded<T>` be equal? Are `Bonded<Void>`s always equal or never equal? Are we sure we want to force deserialization every time we do an `.equals()`.

sigansky: Agree. Let's discuss this.

---
In reply to: [131472247](https://github.com/Microsoft/bond/pull/545#discussion_r131472247) [](ancestors = 131472247)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
sigansky: Good point.

---
In reply to: [131461258](https://github.com/Microsoft/bond/pull/545#discussion_r131461258) [](ancestors = 131461258)

sigansky: Java primitive types have static hashCode method for this.

---
In reply to: [131479794](https://github.com/Microsoft/bond/pull/545#discussion_r131479794) [](ancestors = 131479794,131461258)

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
chwarr: Likely OK to have bonded fall back to instance equality.

tstein: Done. We already have this behavior; I did something slightly wonky to make that hard to change and added a test to assert it. Take a look in the next iteration and let me know if you're buying what I'm selling.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: Yup. Unfortunately, the static [Float.hashCode(float)](https://docs.oracle.com/javase/8/docs/api/java/lang/Float.html#hashCode-float-) wasn't added until Java 8. Went with something based on `Float.floatToIntBits()`.

compiler/src/Language/Bond/Codegen/Java/Util.hs
tstein: I think that's redundant - it's defined on a `Type`, and it's used in places where it's clearly acting on a `Type`. I'd like to keep it like this.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: Artifact of when it was a top-level function. Thanks.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: It will, but I don't like the tradeoff. Putting two things one line like that makes this code harder to read; it saves 10 bytes (`\n\s{8}\n`) of codegen and zero bytes of bytecode in the uncommon case of a struct with no fields.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: By "the code", I think you're referring to the generated code? This seems like a trade-off between compiler code complexity and generated code complexity, and I'm more worried about the former.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: Fixed.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: Done. Thanks.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: Done. We're still starting with 17.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
chadwalters: Was this addressed?

compiler/src/Language/Bond/Codegen/Java/Util.hs
```diff
@@ -62,6 +63,21 @@ isPrimitiveBondType :: Type -> Bool
...
+isJavaPrimitive BT_Int32 = True
+isJavaPrimitive BT_Int64 = True
+isJavaPrimitive BT_UInt8 = True
+isJavaPrimitive BT_UInt16 = True
+isJavaPrimitive BT_UInt32 = True
+isJavaPrimitive BT_UInt64 = True
+isJavaPrimitive BT_Float = True
+isJavaPrimitive BT_Double = True
+isJavaPrimitive BT_Bool = True
+isJavaPrimitive _ = False
```
sigansky: This returns false for (BT_UserDefined Alias), so the generated code will be wrong when a Java primitive type is aliased. Please refer to isPrimitiveNonEnumBondType for solution and I suggest you add a test case with an aliased primitive type.

java/core/src/main/java/com/microsoft/bond/Bonded.java
```diff
@@ -135,4 +134,14 @@
...
+
+    /**
+     * Equality on Bonded and Bonded subtypes is always reference equality
+     * (i.e., the same as Object.equals()). The semantics of .equals() in
+     * codegen'd types depend on this.
+     */
+    @Override
+    public final boolean equals(Object other) {
+        return this == other;
+    }
```
sigansky: Please seal hashCode as well. Also, you can just call super, whose equals/hashCode do exactly what we need.

java/core/src/test/java/com/microsoft/bond/BondedTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+        // Exhaustively test equality.
+        for (int i = 0; i < bondeds.size(); i++) {
+            final Bonded one = bondeds.get(i);
+            for (int j = 0; j < bondeds.size(); j++) {
+                final Bonded another = bondeds.get(j);
+                if (i == j) {
+                    // Same instance - should be .equals().
+                    Assert.assertTrue(
+                        String.format("%s.equals(%s) was false; should have been true", one, another),
```
chwarr: Does `toString` on Bonded do anything useful? Would it be more informative to use the indices?

java/core/src/test/java/com/microsoft/bond/codegen/EqualsHashCodeTests.java
```diff
@@ -0,0 +1,185 @@
...
+        another.str = one.str;
+
+        another.wstr = one.wstr + "eee";
+        Assert.assertNotEquals(one, another);
+        another.wstr = one.wstr;
+    }
+
+    @Test
+    public void containerFields() {
+        final Containers one = new Containers(), another = new Containers();
```
chwarr: Consider adding tests for containers with struct values to ensure that structural equality is recursively applied to the elements of containers.

compiler/tests/generated/complex_types_concatenated.java
```diff
@@ -90,6 +90,21 @@ public Foo() {
         ((StructBondTypeImpl)BOND_TYPE).initializeStructFields(this);
     };
 
+    @Override
+    public boolean equals(Object o) {
+        if (!(o instanceof Foo)) return false;
+        
+        final Foo other = (Foo) o;
```
chwarr: Do we need to worry about a warning along the lines of "unused local `other`"?

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -394,6 +394,56 @@ public #{declName}(com.microsoft.bond.StructBondType<#{declName}<#{paramList}>>
...
+    }
+|]
+    where
+        hash codeExpr = [lt|result += #{codeExpr};
+        result *= 0xeadbeef;
+        result ^= result >> 16;|]
+
+        hashInputs = (hashBase structBase) ++ hashFields
+
+        hashBase (Just t)  = [[lt|super.hashCode()|]]
```
chwarr: Consider `hashBase = maybe [] (const [[lt|super.hashCode()|]]) structBase`

If you still want to use pattern matching--which I'm OK with--we can drop simplify the `Just` case to `Just _`.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -394,6 +394,56 @@ public #{declName}(com.microsoft.bond.StructBondType<#{declName}<#{paramList}>>
...
+        hashBase Nothing = []
+
+        hashFields = map (\Field {..} -> hashCode fieldType fieldName) fields
+        hashCode BT_Bool e    = [lt|(#{e} ? 0 : 1)|]
+        hashCode BT_Int64 e   = [lt|(#{e} ^ (#{e} >> 32)) & 0xFFFFFFFF|]
+        hashCode BT_UInt64 e  = [lt|(#{e} ^ (#{e} >> 32)) & 0xFFFFFFFF|]
+        hashCode BT_Float e   = [lt|Float.floatToRawIntBits(#{e})|]
+        hashCode BT_Double e  = [lt|(Double.doubleToRawLongBits(#{e}) ^ (Double.doubleToRawLongBits(#{e}) >> 32)) & 0xFFFFFFFF|]
```
sigansky: >Double.doubleToRawLongBits(#{e}) ^ (Double.doubleToRawLongBits(#{e}) >> 32) [](start = 37, length = 75)

This won't work well if comparing 0 with -0. Equality holds but hash codes are different.

java/core/src/main/java/com/microsoft/bond/Bonded.java
tstein: Done.

java/core/src/test/java/com/microsoft/bond/BondedTest.java
tstein: I wanted failures to be clearer than "it should have been equal to itself" - this seems like a situation where it's easy to confuse whether the test is broken or the testee is. `Object.toString()` includes the hash code, which is based on the address, so you can tell at a glance whether `one` and `another` were the same object. It looks likes this:
```
java.lang.AssertionError: com.microsoft.bond.WildcardTypeObjectBonded@6f75e721.equals(com.microsoft.bond.WildcardTypeObjectBonded@6f75e721) was false; should have been true
```

java/core/src/test/java/com/microsoft/bond/codegen/EqualsHashCodeTests.java
tstein: Added `list<Primitives>` to `Containers` and broke equivalence by changing a member of one of its elements.

compiler/tests/generated/complex_types_concatenated.java
tstein: I don't get one. `@generated` quiets the IDE on this; I don't think the default invocation of `javac` minds either way.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: The pattern match is easier to read to my eyes. `_`'d the `t`.

java/core/src/test/java/com/microsoft/bond/BondedTest.java
chwarr: For sure better than just "com.microsoft.bond.WildcardTypeObjectBonded"

PR #548, opened 2017-08-07
[Java] Add simple serialization example

chadwalters:
Add an initial example for Java.

Note that, when run, this example will assert until we code generate equals() methods on Bond objects.


examples/java/.gitignore
```diff
@@ -0,0 +1,13 @@
+
+build/
+
+.gradle/
+gradle/
+gradlew
+gradlew.bat
+
+**/.idea/**/workspace.xml
```
tstein: This is a slightly odd mix of styles. `build/` and `**/.idea/...` will both have the effect of matching those files in any dir in `examples/java`. I'm fine with it either way, but have a slight preference for having them match.

examples/java/core/serialization/src/main/bond/serialization.bond
```diff
@@ -0,0 +1,8 @@
+namespace examples.serialization
```
tstein: This would be inappropriate for Javaland because it's not a domain we control. I would say that if you're writing a bondfile specifically for Java consumption, your namespace should match Java conventions (e.g., `com.microsoft.bond.examples.serialization`) and if not, you should use a mapping (see `java/core/build.gradle`). I'd rather see this example do one of those two things. Any thoughts?

examples/java/core/serialization/src/main/java/com/microsoft/bond/examples/Serialization.java
```diff
@@ -0,0 +1,43 @@
...
+import examples.serialization.Struct;
+
+public class Serialization {
+
+    public static void main(final String[] args) throws IOException {
+
+        final Struct obj = new Struct();
+        obj.n = 0x1000;
+        obj.str = "test";
+        obj.items = new ArrayList<>(2);
```
tstein: `obj.items` should be an empty `ArrayList` by the time you get here, and will auto-resize as necessary. You don't need to do this, and it incorrectly implies that collection members aren't initialized by default.

examples/java/.gitignore
chwarr: @chadwalters and I experimented with a number of different styles to get this to work on Windows. This was the one we found that worked to ignore any workspace.xml under an .idea directory potentially in a subdirectory on Windows. The best we can figure is that the implementation of `fnmatch` that Windows Git uses differs from what's documented in the Git manual. Or, we could have misinterpreted the manual. :-) 

If you have a more standard one that works on Windows, I'd be interested in seeing it.

tstein: Gotcha. Fine with me as is.

examples/java/core/serialization/src/main/java/com/microsoft/bond/examples/Serialization.java
```diff
@@ -0,0 +1,43 @@
...
+        final ByteArrayOutputStream output = new ByteArrayOutputStream();
+        final CompactBinaryWriter writer = new CompactBinaryWriter(output, (short) 1);
+
+        final Serializer<Struct> serializer = new Serializer<>();
+        serializer.serialize(obj, writer);
+
+        final ByteArrayInputStream input = new ByteArrayInputStream(output.toByteArray());
+
+        final CompactBinaryReader reader = new CompactBinaryReader(input, (short) 1);
+        final Deserializer<Struct> deserializer = new Deserializer<>((StructBondType<Struct>) obj.getBondType());
```
chwarr: Does the following work? It's potentially a bit shorter and removes the need to import and cast to `StructBondType`:

    new Deserializer<>(Struct.BOND_TYPE)

examples/java/core/serialization/src/main/java/com/microsoft/bond/examples/Serialization.java
```diff
@@ -0,0 +1,43 @@
...
+
+        final Struct obj = new Struct();
+        obj.n = 0x1000;
+        obj.str = "test";
+        obj.items = new ArrayList<>(2);
+        obj.items.add(3.14D);
+        obj.items.add(0D);
+
+        final ByteArrayOutputStream output = new ByteArrayOutputStream();
+        final CompactBinaryWriter writer = new CompactBinaryWriter(output, (short) 1);
```
chwarr: Do we have constants for the versions?

examples/java/core/serialization/src/main/java/com/microsoft/bond/examples/Serialization.java
```diff
@@ -0,0 +1,43 @@
...
+        final Serializer<Struct> serializer = new Serializer<>();
+        serializer.serialize(obj, writer);
+
+        final ByteArrayInputStream input = new ByteArrayInputStream(output.toByteArray());
+
+        final CompactBinaryReader reader = new CompactBinaryReader(input, (short) 1);
+        final Deserializer<Struct> deserializer = new Deserializer<>((StructBondType<Struct>) obj.getBondType());
+        final Struct obj2 = deserializer.deserialize(reader);
+
+        assert obj.equals(obj2) : "Roundtrip failed";
```
chwarr: What about, for now:

    assert obj.n == obj2.n;
    assert obj.str.equals(obj2.str);
    ...

I know it's a bit tedious, but it means we can actually run the examples too.

examples/java/core/serialization/src/main/bond/serialization.bond
chadwalters: I agree with your point. I lean towards using the namespace mapping rather than putting the Java-oriented namespace in the .bond file, since the former facilitates sharing the files across languages a bit more naturally.

examples/java/core/serialization/src/main/java/com/microsoft/bond/examples/Serialization.java
chadwalters: Not yet. Those APIs need to be updated but I want to get the example in now.

examples/java/core/serialization/src/main/java/com/microsoft/bond/examples/Serialization.java
chadwalters: Updated.

examples/java/core/serialization/src/main/java/com/microsoft/bond/examples/Serialization.java
chadwalters: We haven't hooked up the examples to run automatically yet in any case. I'd rather have this in the final form and see us add .equals() support pronto.

examples/java/core/serialization/src/main/java/com/microsoft/bond/examples/Serialization.java
chadwalters: Addressed.

examples/java/core/serialization/src/main/java/com/microsoft/bond/examples/Serialization.java
chwarr: Makes sense.

chadwalters:
CR comments addressed. I removed the IDEA files for now since they weren't proving to be stable. We should look at the use of the IDEA plugin for Gradle but I'd rather not play with that now.

I have added the Application plugin in this version which allows the example to be run via 'gradle run'.

examples/java/core/serialization/build.gradle
```diff
@@ -25,8 +27,15 @@ dependencies {
     compile files('../../../../java/core/build/libs/bond-5.0.0.jar')
 }
 
+compileBond {
+
```
tstein: ~~ The newlines in this look weird to me.

PR #549, opened 2017-08-08
[Java] Codegen bug fixes

sigansky:
This fix contains a few small codegen bug fixes:

1. Fixed the problem with large value assigned to integral numeric types, both for enum and default field values. Enum has still challenges since the parser allows arbitrary-length integers which are then truncated to Haskell's Int type.
2. Type in the generated name of "Something" type for Boolean fields, causing Java compilation error.
3. A bug where Meta struct field was incorrectly initialized with STRING type descriptor which should have been omitted.


compiler/src/Language/Bond/Codegen/Java/Util.hs
```diff
@@ -66,3 +69,9 @@ isGenericBondStructType _ = False
 ifThenElse :: Bool -> Text -> Text -> Text
 ifThenElse True thenCondition _ = thenCondition
 ifThenElse False _ elseCondition = elseCondition
+
+-- takes a bit count and a number and returns its two's complement
+twosComplement :: Integer -> Integer -> Integer
```
tstein: `twosComplement` explains the what, but not the why, which is more important. I'd rather you called this `asSigned` or something like that.

compiler/src/Language/Bond/Codegen/Java/Util.hs
```diff
@@ -66,3 +69,9 @@ isGenericBondStructType _ = False
...
+twosComplement bitCount value = if value < (2 ^ (bitCount - 1))
```
tstein: `bitCount` sounds like there's some arbitrary-length masking going on. `width`, maybe?

compiler/src/Language/Bond/Codegen/TypeMapping.hs
```diff
@@ -127,6 +128,10 @@ getTypeName c t = fix' $ runReader (typeName t) c
 getInstanceTypeName :: MappingContext -> Type -> Builder
 getInstanceTypeName c t = runReader (instanceTypeName t) c
 
+-- todo
```
chwarr: Left over "todo" comment?

compiler/src/Language/Bond/Codegen/TypeMapping.hs
```diff
@@ -329,6 +334,17 @@ aliasTypeName a args = do
     fragment (Fragment s) = pureText s
     fragment (Placeholder i) = typeName $ args !! i
 
+-- | Builder for the type alias element name in context of 'TypeNameBuilder' monad.
+aliasElementTypeName :: Declaration -> [Type] -> TypeNameBuilder
+aliasElementTypeName a args = do
```
chwarr: The only difference I see between `aliasElementTypeName` and `aliasTypeName` is that the former used `elementTypeName` and the latter uses `typeName`. Consider abstracting this out, particularly given how long this function is.

PR #551, opened 2017-08-09
Add more Java examples

chadwalters:
Port blob, enumerations, and import examples from C#.

Note Java version of import example is named "imports" here due to reserved word in Java.

### DIFF
--- a/examples/java/core/blob/src/main/bond/schema.bond
+++ b/examples/java/core/blob/src/main/bond/schema.bond
@@ -1,0 +1,8 @@
+﻿namespace Examples
+
+struct Example
+{
+    0: vector<blob> ListOfBlobs;
+    1: nullable<blob> NullableBlob;
+    2: blob UninitializeBlob = nothing;
+}
--- a/examples/java/core/blob/src/main/java/com/microsoft/bond/examples/Blob.java
+++ b/examples/java/core/blob/src/main/java/com/microsoft/bond/examples/Blob.java
@@ -12,15 +12,16 @@ import com.microsoft.bond.Something;
 import com.microsoft.bond.protocol.CompactBinaryReader;
 import com.microsoft.bond.protocol.CompactBinaryWriter;
 
+// See build.gradle for namespace mapping
 import com.microsoft.bond.examples.blob.Example;
 
 public class Blob {
 
     public static void main(final String[] args) throws IOException {
 
-        byte[] data = new byte[256];
+        final byte[] data = new byte[256];
         for (int i = 0; i < data.length; i++) {
-            data[i] = (byte)i;
+            data[i] = (byte) i;
         }
 
         final Example obj = new Example();
--- a/examples/java/core/blob/src/main/java/com/microsoft/bond/examples/Blob.java
+++ b/examples/java/core/blob/src/main/java/com/microsoft/bond/examples/Blob.java
@@ -7,7 +7,6 @@ import java.util.Arrays;
 
 import com.microsoft.bond.Deserializer;
 import com.microsoft.bond.Serializer;
-import com.microsoft.bond.StructBondType;
 import com.microsoft.bond.Something;
 
 import com.microsoft.bond.protocol.CompactBinaryReader;
@@ -39,7 +38,7 @@ public class Blob {
         final ByteArrayInputStream input = new ByteArrayInputStream(output.toByteArray());
 
         final CompactBinaryReader reader = new CompactBinaryReader(input, (short) 1);
-        final Deserializer<Example> deserializer = new Deserializer<>((StructBondType<Example>) obj.getBondType());
+        final Deserializer<Example> deserializer = new Deserializer<>(Example.BOND_TYPE);
         final Example obj2 = deserializer.deserialize(reader);
 
         assert obj.equals(obj2) : "Roundtrip failed";
--- a/examples/java/core/blob/src/main/java/com/microsoft/bond/examples/Blob.java
+++ b/examples/java/core/blob/src/main/java/com/microsoft/bond/examples/Blob.java
@@ -1,0 +1,48 @@
+package com.microsoft.bond.examples;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Arrays;
+
+import com.microsoft.bond.Deserializer;
+import com.microsoft.bond.Serializer;
+import com.microsoft.bond.StructBondType;
+import com.microsoft.bond.Something;
+
+import com.microsoft.bond.protocol.CompactBinaryReader;
+import com.microsoft.bond.protocol.CompactBinaryWriter;
+
+import com.microsoft.bond.examples.blob.Example;
+
+public class Blob {
+
+    public static void main(final String[] args) throws IOException {
+
+        byte[] data = new byte[256];
+        for (int i = 0; i < data.length; i++) {
+            data[i] = (byte)i;
+        }
+
+        final Example obj = new Example();
+        obj.ListOfBlobs.add(Arrays.copyOfRange(data, 0, 10));
+        obj.ListOfBlobs.add(Arrays.copyOfRange(data, 10, 20));
+        obj.NullableBlob = Arrays.copyOfRange(data, 20, 30);
+        obj.UninitializeBlob = Something.wrap(Arrays.copyOfRange(data, 30, 100));
+
+        final ByteArrayOutputStream output = new ByteArrayOutputStream();
+        final CompactBinaryWriter writer = new CompactBinaryWriter(output, (short) 1);
+
+        final Serializer<Example> serializer = new Serializer<>();
+        serializer.serialize(obj, writer);
+
+        final ByteArrayInputStream input = new ByteArrayInputStream(output.toByteArray());
+
+        final CompactBinaryReader reader = new CompactBinaryReader(input, (short) 1);
+        final Deserializer<Example> deserializer = new Deserializer<>((StructBondType<Example>) obj.getBondType());
+        final Example obj2 = deserializer.deserialize(reader);
+
+        assert obj.equals(obj2) : "Roundtrip failed";
+    }
+
+}
--- a/examples/java/core/enumerations/src/main/bond/enumerations.bond
+++ b/examples/java/core/enumerations/src/main/bond/enumerations.bond
@@ -1,0 +1,21 @@
+namespace Examples;
+
+enum Color
+{
+    Orange = 1,
+    Yellow = 2,
+}
+
+enum Fruit
+{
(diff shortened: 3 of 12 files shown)


examples/java/core/blob/src/main/java/com/microsoft/bond/examples/Blob.java
```diff
@@ -0,0 +1,48 @@
...
+        final ByteArrayOutputStream output = new ByteArrayOutputStream();
+        final CompactBinaryWriter writer = new CompactBinaryWriter(output, (short) 1);
+
+        final Serializer<Example> serializer = new Serializer<>();
+        serializer.serialize(obj, writer);
+
+        final ByteArrayInputStream input = new ByteArrayInputStream(output.toByteArray());
+
+        final CompactBinaryReader reader = new CompactBinaryReader(input, (short) 1);
+        final Deserializer<Example> deserializer = new Deserializer<>((StructBondType<Example>) obj.getBondType());
```
sigansky: >obj.getBondType() [](start = 96, length = 17)

Should this be Example.BOND_TYPE? Using "obj" doesn't seem right to me since you typically don't have an instance prior to deserializing.

examples/java/core/enumerations/src/main/bond/enumerations.bond
```diff
@@ -0,0 +1,21 @@
...
+    Apple = 1,
+    Orange = 2,
+}
+
+enum Limits
+{
+    Int32Min = -2147483648,
+    Int32Max = 2147483647,
+    UInt32Min = 0,
+    UInt32Max = 0xFFFFFFFF,
```
sigansky: This example seems to imply that enum values can be used to represent unsigned 32-bit integers, which is not the case since an enum is backed by a signed 32-bit integer.

examples/java/core/blob/src/main/java/com/microsoft/bond/examples/Blob.java
```diff
@@ -0,0 +1,48 @@
...
+        final Example obj2 = deserializer.deserialize(reader);
+
+        assert obj.equals(obj2) : "Roundtrip failed";
```
sigansky: Should this wait until we implement equals, which is a separate PR right now?

examples/java/core/blob/src/main/java/com/microsoft/bond/examples/Blob.java
chadwalters: Yes, I forgot to backport these changes from the first example. Will fix.

examples/java/core/enumerations/src/main/bond/enumerations.bond
chadwalters: It's a direct port of the C# example, using the same Bond IDL. What do you propose?

examples/java/core/blob/src/main/java/com/microsoft/bond/examples/Blob.java
chadwalters: I'd rather just get these in now as is. They compile and we should add the correct .equals support pronto.

examples/java/core/enumerations/src/main/bond/enumerations.bond
sigansky: I don't agree with the example (IMO it's misleading) but I agree with the change since this is just a port. No issue.

---
In reply to: [132254217](https://github.com/Microsoft/bond/pull/551#discussion_r132254217) [](ancestors = 132254217)

examples/java/core/blob/src/main/java/com/microsoft/bond/examples/Blob.java
```diff
@@ -0,0 +1,47 @@
...
+import com.microsoft.bond.protocol.CompactBinaryReader;
+import com.microsoft.bond.protocol.CompactBinaryWriter;
+
+import com.microsoft.bond.examples.blob.Example;
+
+public class Blob {
+
+    public static void main(final String[] args) throws IOException {
+
+        byte[] data = new byte[256];
```
chwarr: Can `data` be final?

examples/java/core/imports/src/main/java/com/microsoft/bond/examples/Imports.java
```diff
@@ -0,0 +1,41 @@
...
+import com.microsoft.bond.protocol.CompactBinaryReader;
+import com.microsoft.bond.protocol.CompactBinaryWriter;
+
+import com.microsoft.bond.examples.imports.Message;
+import com.microsoft.bond.examples.imports.common.Priority;
+
+public class Imports {
+
+    public static void main(final String[] args) throws IOException {
+
```
chwarr: Minor: is this blank line after `main` before content intentional? Not sure what our style is for Java, having only done minimal work in it.

examples/java/core/imports/src/main/java/com/microsoft/bond/examples/Imports.java
```diff
@@ -0,0 +1,41 @@
...
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+
+import com.microsoft.bond.Deserializer;
+import com.microsoft.bond.Serializer;
+
+import com.microsoft.bond.protocol.CompactBinaryReader;
+import com.microsoft.bond.protocol.CompactBinaryWriter;
+
+import com.microsoft.bond.examples.imports.Message;
```
chwarr: Consider adding a comment pointing people at the .gradle file to show how the namespace remapping works.

examples/java/core/imports/build.gradle
```diff
@@ -0,0 +1,44 @@
...
+
+dependencies {
+    compile files('../../../../java/core/build/libs/bond-5.0.0.jar')
+}
+
+compileBond {
+
+    bondfiles 'src/main/bond/schema.bond'
+
+    options '-n', 'Examples=com.microsoft.bond.examples.imports',
```
chwarr: For readability, consider using the long form of `-n` (`--namespace=`) and `-i` (`--import-dir=`).

examples/java/core/enumerations/src/main/bond/enumerations.bond
tstein: Perhaps we shouldn't port it exactly as-is. Rather than copy the uint stuff, which isn't really meaningful in Java world, it would be more useful to define an enum with a single value >= 2^31, and instead of doing the hex comparison, have the Java show that the value is cast to signed? Something like:
```
enum HowMany
{
  None,
  One,
  Many = 3,
  Lots = 0x80000000
}
```
In the Java:
```
// Enum values greater than INT_MAX will cast to their signed equivalents.
assert HowMany.Lots == Integer.MIN_VALUE;
```

examples/java/core/imports/build.gradle
tstein: Any thoughts on prioritizing adding these as first-class params to the plugin? It would be a single-sitting task to make this work:
```
compileBond {
  bondfiles 'whatever'
  mapNamespace 'Examples=com.microsoft.bond.examples.imports',
               'Common=com.microsoft.bond.examples.imports.common'
  importDir 'src/main/bond/common'
}
```

examples/java/core/imports/src/main/java/com/microsoft/bond/examples/Imports.java
tstein: I don't think there's much value in consistency on this particular point, but most of our code has the blank. IDEA has no opinion and its auto-format will leave it alone. Did a spot check of some top trending and starred Java repos and the blank beats !(the blank) 4-2.

examples/java/core/imports/src/main/java/com/microsoft/bond/examples/Imports.java
```diff
@@ -0,0 +1,41 @@
...
+
+        final ByteArrayInputStream input = new ByteArrayInputStream(output.toByteArray());
+
+        final CompactBinaryReader reader = new CompactBinaryReader(input, (short) 1);
+        final Deserializer<Message> deserializer = new Deserializer<>(Message.BOND_TYPE);
+        final Message dst = deserializer.deserialize(reader);
+
+        assert src.equals(dst) : "Roundtrip failed";
+    }
+
```
tstein: ~~ While we're talking about whitespace, I don't think anyone goes to bat for this one.

examples/java/core/enumerations/src/main/bond/enumerations.bond
chadwalters: I'd really like to keep the examples the same across the different languages so that people can see the implications of some of their choices in various languages -- especially because many of our current customers who would be picking up Java have C# interop needs. We can add more documentation around this.

examples/java/core/blob/src/main/bond/schema.bond
```diff
@@ -0,0 +1,8 @@
+﻿namespace Examples
+
+struct Example
+{
+    0: vector<blob> ListOfBlobs;
+    1: nullable<blob> NullableBlob;
+    2: blob UninitializeBlob = nothing;
```
chadwalters: This is consistent with the C#.

examples/java/core/imports/build.gradle
chadwalters: For now, I think it is best just to leave them as passthrough to the underlying gbc.

examples/java/core/imports/src/main/java/com/microsoft/bond/examples/Imports.java
chadwalters: I prefer it for symmetry with the one at the start.

examples/java/core/blob/src/main/java/com/microsoft/bond/examples/Blob.java
chadwalters: Done.

examples/java/core/enumerations/src/main/bond/enumerations.bond
chadwalters: Let's consider revamping this example in both C# and Java at a future date.

chadwalters: I did add the comment on the Java side for now.

chadwalters:
CR comments addressed. Please squash into a single commit when merging.

examples/java/core/blob/src/main/bond/schema.bond
tstein: Point taken about keeping the IDL in the examples the same, but this really looks like a typo to me. How about fixing it in both Java and C# in this commit?

chwarr:
Merging now (by hand to deal with the fixups).

chwarr:
Merged into the jvm branch in commit 6a571447928b80cb4d0b2523959f3ca319bbdf16.

PR #552, opened 2017-08-09
[Java] Added parse method to enums to convert from strings

sigansky:


java/core/src/test/java/com/microsoft/bond/testmodel/MyEnum.java
```diff
@@ -78,6 +78,30 @@ public static MyEnum get(int value) {
...
+        } else if (str.equals("Blue")) {
+            return Blue;
+        } else if (str.equals("Green")) {
+            return Green;
+        } else if (str.equals("Brown")) {
+            return Brown;
+        } else if (str.equals("Black")) {
+            return Black;
+        } else {
+            throw new IllegalArgumentException("Invalid 'MyEnum' enum value: '" + str + "'.");
```
chwarr: Consider "Unknown" instead of "Invalid". It may be that this schema is old and doesn't know about a new value.

java/core/src/test/java/com/microsoft/bond/testmodel/schema.bond
```diff
@@ -59,3 +59,15 @@ struct E<U> : D
...
+enum MyEnum
+{
+    White = 1,
+    Yellow = 2,
+    Orange = 3,
+    Purple = 4,
+    Blue = 5,
+    Green = 6,
+    Brown = 7,
+    Black = 8
```
chwarr: Add a test case like `AllTheColors = 1` to check the behavior when there's an aliasing label?

java/core/src/test/java/com/microsoft/bond/testmodel/MyEnum.java
```diff
@@ -78,6 +78,30 @@ public static MyEnum get(int value) {
         }
     }
 
+    public static MyEnum parse(String str) {
```
chwarr: Would it make sense for there to be two methods: `tryParse` that returns `null` when `str` doesn't match anything and `parse` that throws (implemented by calling `tryParse` and throwing if the return is `null`)?

sigansky: I don't know the requirements. Parse makes sense, but I am not sure about tryParse (which is .NET's naming convention). Java native enums only have method that throws exception (Enum.valueOf). 

I think I should rename "parse" to "valueOf" to keep it close to Java's true enums (and also since we're not really parsing), but I prefer to not add a "try" method at this time.

---
In reply to: [132336715](https://github.com/Microsoft/bond/pull/552#discussion_r132336715) [](ancestors = 132336715)

java/core/src/test/java/com/microsoft/bond/testmodel/schema.bond
sigansky: I tested that case manually. Note that this schema and related classes are to be deleted when Ted checks in his change.

---
In reply to: [132336293](https://github.com/Microsoft/bond/pull/552#discussion_r132336293) [](ancestors = 132336293)

java/core/src/test/java/com/microsoft/bond/testmodel/MyEnum.java
chwarr: I'll defer to you and @tstein on the Java way to do things.

java/core/src/test/java/com/microsoft/bond/testmodel/schema.bond
chwarr: Can you add it here anyway. My mental Haskell evaluator is struggling to generate the Java code for this construct.

tstein:
This change should also update golden files.

java/core/src/test/java/com/microsoft/bond/testmodel/MyEnum.java
tstein: Small point, but I don't think that applies here - the behavior you're talking about is a promise about pulling enums off the wire, if you let bond decide how to {,de}serialize them. If you've gone out of your way to serialize them as strings and then tried to get enums back through this function, you have done something weird and out of the bounds of the promise.

java/core/src/test/java/com/microsoft/bond/testmodel/schema.bond
tstein: If any of the golden file inputs cover this case, you can also see it when this PR is updated with golden file changes.

java/core/src/test/java/com/microsoft/bond/testmodel/MyEnum.java
tstein: If the behavior matches the true `Enum`'s `valueOf`, the name should definitely be the same.

Turning a function that could comfortably be 
```
String -> Nullable EnumT
```
into
```
String -> Nullable EnumT, throws IllegalArgException
// we pinky-swear never to return null
```
is a bad Java convention I am broadly willing to violate. For the sake of expediency, I'm okay with this, but I want to revisit it later. Might make sense to do so at the same time as we revisit the top-level APIs.

java/core/src/test/java/com/microsoft/bond/testmodel/schema.bond
sigansky: Why?

---
In reply to: [132341756](https://github.com/Microsoft/bond/pull/552#discussion_r132341756) [](ancestors = 132341756)

java/core/src/test/java/com/microsoft/bond/testmodel/MyEnum.java
sigansky: I prefer we keep it close to true Java enums until we get some constructive feedback from customers on this.

---
In reply to: [132561015](https://github.com/Microsoft/bond/pull/552#discussion_r132561015) [](ancestors = 132561015)

sigansky:
Yes, working on this.

---
In reply to: [321634791](https://github.com/Microsoft/bond/pull/552#issuecomment-321634791) [](ancestors = 321634791)

java/core/src/test/java/com/microsoft/bond/testmodel/schema.bond
chwarr: With the golden file updates, we're good.

PR #553, opened 2017-08-11
[java] [c#] Add inheritance example

chwarr:
NB: This is a port of the [C++ example](https://github.com/Microsoft/bond/tree/5391636d5b4a8c59981527822841649a292d6be6/examples/cpp/core/inheritance), but doesn't have the deserialize-the-base-parts-into-a-derived.


examples/java/core/inheritance/src/main/java/com/microsoft/bond/examples/Inheritance.java
```diff
@@ -0,0 +1,48 @@
...
+    public static void main(final String[] args) throws IOException {
+
+        final Derived derived = new Derived();
+        derived.str = "derived";
+        ((Base)derived).str = "derived base";
+        final CompactBinaryReader derived_payload = new CompactBinaryReader(
+            new ByteArrayInputStream(Serialize(derived)),
+            (short) 1);
+
+        {
```
sigansky: What is the reason for this block?

examples/cs/core/inheritance/program.cs
```diff
@@ -0,0 +1,38 @@
...
+    static class Program
+    {
+        static void Main()
+        {
+            var derivedObj = new Derived {str = "derived"};
+            ((Base) derivedObj).str = "derived base";
+            var derivedPayload = new CompactBinaryReader<InputBuffer>(
+                new InputBuffer(Serialize(derivedObj)));
+
+            {
```
sigansky: What is the reason for this block?

examples/cs/core/inheritance/program.cs
```diff
@@ -0,0 +1,38 @@
...
+                // Deserialize Base from payload containing Derived
+                Base obj = Deserialize<Base>.From(derivedPayload);
+                Debug.Assert(Comparer.Equal<Base>(obj, derivedObj));
+            }
+        }
+
+        static ArraySegment<byte> Serialize<T>(T obj)
+        {
+            var output = new OutputBuffer(128);
```
sigansky: Why not just leave it to the default? 

As this is an example code it may create the impression that 128 is the recommended initial capacity (if that is in indeed the case then it should appear in OutputBuffer code).

examples/java/core/inheritance/src/main/java/com/microsoft/bond/examples/Inheritance.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    private static <T extends BondSerializable> byte[] Serialize(T obj) throws IOException {
+        final ByteArrayOutputStream output = new ByteArrayOutputStream();
+        final CompactBinaryWriter writer = new CompactBinaryWriter(output, (short) 1);
+
+        final Serializer<T> serializer = new Serializer<>();
+        serializer.serialize(obj, writer);
+        return output.toByteArray();
+    }
```
sigansky: This method is general enough and could be useful in many places, so am considering adding a static utility for it in the Serializer class. 

What are your thoughts on this?

examples/java/core/inheritance/src/main/java/com/microsoft/bond/examples/Inheritance.java
```diff
@@ -0,0 +1,48 @@
...
+        {
+            // Deserialize Base from payload containing Derived
+            final Deserializer<Base> deserializer = new Deserializer<>(Base.BOND_TYPE);
+            final Base obj = deserializer.deserialize(derived_payload);
+            assert obj.str.equals(((Base)derived).str) : "Base parts should be equal";
+            // assert obj.equals((Base)derived): "Base parts should be equal";
+        }
+    }
+
+    private static <T extends BondSerializable> byte[] Serialize(T obj) throws IOException {
```
sigansky: >Serialize [](start = 55, length = 9)

Lowercase.

examples/java/core/inheritance/src/main/java/com/microsoft/bond/examples/Inheritance.java
```diff
@@ -0,0 +1,48 @@
...
+import com.microsoft.bond.examples.inheritance.Derived;
+
+public class Inheritance {
+
+    public static void main(final String[] args) throws IOException {
+
+        final Derived derived = new Derived();
+        derived.str = "derived";
+        ((Base)derived).str = "derived base";
+        final CompactBinaryReader derived_payload = new CompactBinaryReader(
```
sigansky: >derived_payload [](start = 34, length = 15)

camelCase

examples/java/core/inheritance/src/main/java/com/microsoft/bond/examples/Inheritance.java
```diff
@@ -0,0 +1,48 @@
...
+            new ByteArrayInputStream(Serialize(derived)),
+            (short) 1);
+
+        {
+            // Deserialize Base from payload containing Derived
+            final Deserializer<Base> deserializer = new Deserializer<>(Base.BOND_TYPE);
+            final Base obj = deserializer.deserialize(derived_payload);
+            assert obj.str.equals(((Base)derived).str) : "Base parts should be equal";
+            // assert obj.equals((Base)derived): "Base parts should be equal";
```
sigansky: Please remove the field comparison line and uncomment the object comparison line. We should have equals implemented within 1-2 days.

sigansky: >(Base)derived [](start = 33, length = 13)

You don't need to cast this since the method takes an Object.

examples/java/core/inheritance/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    compile files('../../../../java/core/build/libs/bond-5.0.0.jar')
+}
+
+compileBond {
+
```
tstein: ~~ I would rather these blanks not become a thing. Why not:
```
compileBond {
  bondfiles ...
  options ...
}
```

examples/java/core/inheritance/src/main/java/com/microsoft/bond/examples/Inheritance.java
tstein: How would it know what kind of writer to make?

chwarr: Yes, something like this makes sense. Perhaps as a member of the CB writer?

    byte[] payload = CompactBinaryWriter.serialize(obj);

examples/cs/core/inheritance/program.cs
chwarr: Forgot to clean these up when porting the C++ example. It has two example sections, while C# and Java only have one.

chwarr:
Rebased to pickup the golden test fixes.

PR #555, opened 2017-08-11
[Java] Fixed collection serialization bugs

sigansky:


tstein:
Needs manual rebase on origin/jvm before we can merge.

java/core/src/main/java/com/microsoft/bond/ListBondType.java
```diff
@@ -154,7 +154,7 @@ protected final void serializeField(
             StructBondType.StructField<List<TElement>> field) throws IOException {
         this.verifySerializedNonNullableFieldIsNotSetToNull(value, field);
         int count = value.size();
-        if (count == 0 && field.isOptional()) {
+        if (!field.isDefaultNothing() && count == 0 && field.isOptional()) {
             context.writer.writeFieldOmitted(BondDataType.BT_LIST, field.getId(), field);
         } else {
```
chwarr: Let's say the field is `= nothing` and hasn't been set. Where does the writing of the field get skipped completely? In the generated `BOND_TYPE` serialize method? If it doesn't get skipped at a higher level, this condition looks like it will go down the `writeFieldBegin` path, which doesn't seem right: there's no value to write...

java/core/src/main/java/com/microsoft/bond/NullableBondType.java
```diff
@@ -109,9 +108,10 @@ protected final TValue deserializeValue(TaggedDeserializationContext context) th
                     this.valueType.getBondDataType(),
                     this.getFullName());
         }
+        TValue value = null;
         if (context.readContainerResult.count == 1) {
             value = this.valueType.deserializeValue(context);
-        } else {
+        } else if (context.readContainerResult.count > 1){
```
chwarr: Are there any unit tests that need to be updated/added to cover this behavior?

java/core/src/main/java/com/microsoft/bond/ListBondType.java
sigansky: A field with "= nothing" should be serialized using serializeSomethingField which does that check. The StructField instance for each field knows which method to call for serialization. The same pattern holds for deserialization.

java/core/src/main/java/com/microsoft/bond/NullableBondType.java
sigansky: There should be. Do you think it's blocking this change from pushed?

java/core/src/main/java/com/microsoft/bond/ListBondType.java
chwarr: Thanks for the pointer. I see it now.

java/core/src/main/java/com/microsoft/bond/NullableBondType.java
chwarr: Gotcha.

PR #560, opened 2017-08-14
java core: Add doc for simple/omitted/=nothing interaction.

tstein:
This addresses a request for a comment from #542 that I misunderstood at the time as targeting C#, not Java.

### DIFF
--- a/java/core/src/main/java/com/microsoft/bond/protocol/SimpleBinaryWriter.java
+++ b/java/core/src/main/java/com/microsoft/bond/protocol/SimpleBinaryWriter.java
@@ -49,6 +49,13 @@ public final class SimpleBinaryWriter implements ProtocolWriter {
     @Override
     public void writeFieldOmitted(
             final BondDataType type, final int id, final FieldMetadata metadata) throws IOException {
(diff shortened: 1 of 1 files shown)


java/core/src/main/java/com/microsoft/bond/protocol/SimpleBinaryWriter.java
```diff
@@ -49,6 +49,13 @@ public void writeVersion() throws IOException {
     @Override
     public void writeFieldOmitted(
             final BondDataType type, final int id, final FieldMetadata metadata) throws IOException {
+        // Simple doesn't support omitting fields, so we need to explicitly write the default values
+        // for optional fields.
+        //
+        // Fields that are = nothing have no default, so it is an unrecoverable error to attempt to
+        // serialize such a field if it is actually set to nothing (e.g., a SomethingObject which is
+        // null).
+
```
chwarr: Minor: is the blank line on 58 intentional?

tstein: Yeah. My intent was that you would read it as an implementer's note on this function, not on any particular lines within it.

PR #563, opened 2017-08-15
[c#] Fix cs.sln inheritance project

chwarr:
When the schema_view example was merged into cs.sln in commit
1406febda4, a syntax error was introduced. This commit fixes that and
ensures that inheritance and schema_view have the correct
project-to-project dependencies as well.


chwarr:
Rebased atop commit e3abbe48 which has a similar syntax error in cs.sln.

PR #564, opened 2017-08-16
Add record_streaming example

chwarr:
Still to do:

* [x] Figure out if there's a better pair of input/output streams to use. `ByteArrayOutputStream` maintains a contiguous buffer, which isn't needed for record streaming.
* [x] Write C# example.

NB: This PR is based atop of another PR, #563, which fixes the cs.sln problem. That's why you see commit d51bd26 here. Please ignore that. It will be rebased away after #563 is merged.


examples/java/core/record_streaming/src/main/java/com/microsoft/bond/examples/RecordStreaming.java
```diff
@@ -0,0 +1,77 @@
...
+        // Read them all back.
+        final ByteArrayInputStream input = new ByteArrayInputStream(output.toByteArray());
+        deserializeStream(input);
+    }
+
+    private static ByteArrayOutputStream serializeStream() throws IOException {
+        // An output stream can be used to accumulate multiple serialized
+        // structs. This can be done by re-using the same writer over the same
+        // stream.
+        final ByteArrayOutputStream output = new ByteArrayOutputStream();
```
chwarr: `ByteArrayOutputStream` maintains a contiguous buffer. Record streaming doesn't need that. Is there a better pair of streams to use in this case?

For example, it could use a stream that just allocated buffers on demand, and then on the reading side keeps track of where and in which buffer the current position is.

examples/java/core/record_streaming/src/main/java/com/microsoft/bond/examples/RecordStreaming.java
```diff
@@ -0,0 +1,77 @@
...
+
+        // Each call to deserialize will advance the underlying stream's
+        // position. When there are no more bytes available, all the
+        // structs have been deserialized.
+        //
+        // You may find that you need to pack serialized Bond structs into a
+        // stream in a different way depending on your needs. The reader just
+        // needs its underlying stream to be positioned at the beginning of a
+        // serialized struct.
+        while (input.available() > 0) {
```
chwarr: Is this the correct idiom here?

examples/java/core/record_streaming/src/main/java/com/microsoft/bond/examples/RecordStreaming.java
```diff
@@ -0,0 +1,77 @@
...
+
+        // Each call to deserialize will advance the underlying stream's
+        // position. When there are no more bytes available, all the structs
+        // have been deserialized.
+        //
+        // You may find that you need to pack serialized Bond structs into a
+        // stream in a different way depending on your needs. The reader
+        // just needs its underlying stream to be positioned at the
+        // beginning of a serialized struct.
+        while (input.available() > 0) {
```
chwarr: Is this the correct idiom to check if there's more data?

examples/java/core/record_streaming/src/main/java/com/microsoft/bond/examples/RecordStreaming.java
```diff
@@ -0,0 +1,77 @@
...
+        // Read them all back.
+        final ByteArrayInputStream input = new ByteArrayInputStream(output.toByteArray());
+        deserializeStream(input);
+    }
+
+    private static ByteArrayOutputStream serializeStream() throws IOException {
+        // An output stream can be used to accumulate multiple serialized
+        // structs. This can be done by re-using the same writer over the
+        // same stream.
+        final ByteArrayOutputStream output = new ByteArrayOutputStream();
```
chwarr: `ByteArrayOutputStream` maintains a contiguous buffer. Record streaming doesn't need that. Is there a better pair of streams to use in this case?

For example, it could use a stream that just allocated buffers on demand, and then on the reading side keeps track of where and in which buffer the current position is.

examples/java/core/record_streaming/src/main/java/com/microsoft/bond/examples/RecordStreaming.java
chwarr: @tstein says "Yes".

PR #566, opened 2017-08-16
WIP: [java] Add nothing_default example

chwarr:
Still left to do:

* [x] investigate why serialization is working when we expect it to throw
* [x] add C# example


examples/java/core/nothing_default/src/main/java/com/microsoft/bond/examples/NothingDefault.java
```diff
@@ -0,0 +1,87 @@
...
+        final Struct_v1 v1 = new Struct_v1();
+
+        boolean sawException = false;
+
+        // Struct_v1 has a required field foo which by default is set to 'nothing'.
+        // If we try to serialize object v1 w/o initializing the field to some value
+        // Bond will throw an exception.
+        try {
+            System.out.println("Serializing v1...");
+            serialize(v1);
```
chwarr: I expected serialization of Struct_v1 in its current state to throw, but it doesn't. Looking at the code, it seems like we're never checking whether the field is required before we skip serializing it because it is null.

The generated code calls `StructBondType.SomethingInt32Field()`:

```
@Override
protected final void serializeStructFields(com.microsoft.bond.BondType.SerializationContext context, Struct_v1 value) throws java.io.IOException {
    this.foo.serialize(context, value.foo);
    this.baz.serialize(context, value.baz);
}
```

From [`StructBondType.SomethingInt32Field()`](https://github.com/Microsoft/bond/blob/jvm/java/core/src/main/java/com/microsoft/bond/StructBondType.java#L1639-L1642):

```
public final void serialize(
    SerializationContext context, SomethingInteger value) throws IOException {
    Int32BondType.serializePrimitiveSomethingField(context, value, this);
}
```

From [`Int32BondType.serializePrimitiveSomethingField()`](https://github.com/Microsoft/bond/blob/jvm/java/core/src/main/java/com/microsoft/bond/Int32BondType.java#L167-L174):

```
protected static void serializePrimitiveSomethingField(
        SerializationContext context,
        SomethingInteger value,
        StructBondType.StructField<Integer> field) throws IOException {
    if (value != null) {
        serializePrimitiveField(context, value.value, field);
    }
}
```

sigansky: This has been fixed. Fix is in core review.

---
In reply to: [133536169](https://github.com/Microsoft/bond/pull/566#discussion_r133536169) [](ancestors = 133536169)

chwarr:
@tstein: rebased and ready for final review

examples/java/core/nothing_default/src/main/java/com/microsoft/bond/examples/NothingDefault.java
chwarr: Rebased atop the fix. Thanks.

examples/java/core/nothing_default/src/main/java/com/microsoft/bond/examples/NothingDefault.java
```diff
@@ -0,0 +1,87 @@
...
+        // Struct_v1. It is initialized to 'nothing' by default. By checking
+        // if the field is 'nothing' (== null) after de-serialization we can
+        // detect if it was present in the payload or not.
+        assert v2.baz == null : "baz was expected to have been omitted because it was set to nothing";
+        assert v2.bar == null : "v1 didn't have a bar field, so it should be nothing in v2";
+    }
+
+    private static <T extends BondSerializable> byte[] serialize(T obj) throws IOException {
+        final ByteArrayOutputStream output = new ByteArrayOutputStream();
+        final CompactBinaryWriter writer = new CompactBinaryWriter(output, (short) 1);
```
tstein: These `(short)` casts are no longer necessary.

chwarr: Fixed. Thanks.

PR #567, opened 2017-08-17
Fix Java StructDef full name to come from IDL

chadwalters:
This aligns with C++/C# implementations.

Also adds Java filter to compiler project in Visual Studio solution.

### DIFF
--- a/compiler/src/Language/Bond/Codegen/Java/Class_java.hs
+++ b/compiler/src/Language/Bond/Codegen/Java/Class_java.hs
@@ -470,7 +470,7 @@ package #{javaPackage};
         javaPackage = sepBy "." toText $ getNamespace java
 
         -- struct -> Java class
-        typeDefinition Struct {..} = [lt|
+        typeDefinition s@Struct {..} = [lt|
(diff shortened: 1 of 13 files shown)


compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -485,6 +485,16 @@ public class #{typeNameWithParams declName declParams}#{maybe interface baseClas
         #{makeStructBondTypeMember_initialize java declParams structFields structBase}
 
         @Override
+        public final String getName() {
+            return "#{declName}";
+        }
+
+        @Override
+        public final String getQualifiedName() {
+            return "#{getDeclTypeName idl s}";
```
tstein: This seems off. We have a `qualifiedDeclaredTypeName` in `Java/Util.hs`, but even if you don't want to use it, there's already a `MappingContext` passed to `class_java` called `java`.

chadwalters:
Updated.

chwarr:
Ping: @tstein, please look at the updated changes.

PR #568, opened 2017-08-17
[C#] Fix interaction with CBv2 and bonded fields 

chadwalters:
Serialization of bonded fields was interacting with the state in the CompactBinaryWriter because it triggered first-pass again. This fix prevents the first-pass writer from being returened in the middle of the
second pass.

### DIFF
--- a/cs/src/core/protocols/CompactBinary.cs
+++ b/cs/src/core/protocols/CompactBinary.cs
@@ -176,7 +176,11 @@ namespace Bond.Protocols
         {
             if (version == 2)
             {
-                return firstPassWriter.Value;
+                // Only return first pass if not in middle of second pass
+                if (lengths.Count == 0)
+                {
+                    return firstPassWriter.Value;
+                }
             }
 
             return null;
--- a/cs/test/core/BondedTests.cs
+++ b/cs/test/core/BondedTests.cs
@@ -438,5 +438,15 @@
             Assert.IsTrue(Comparer.Equal(poly1, obj.poly[1].Deserialize()));
             Assert.IsTrue(Comparer.Equal(poly2, obj.poly[2].Deserialize<Derived>()));
         }
+
+        [Test]
+        public void SerializeBondedCB2Multiple()
+        {
+            var obj = new StructWithBonded();
+
+            var stream = new MemoryStream();
+            stream.SetLength(0);
+            Util.SerializeCB2Multiple(obj, stream, 2);
+        }
     }
 }
--- a/cs/test/core/Util.cs
+++ b/cs/test/core/Util.cs
@@ -393,6 +393,8 @@ namespace UnitTest
             for (uint i = 0; i < times; i++)
             {
                 Serialize.To(writer, obj);
+
+                // Verify that CBv2 writer has an active FirstPassWriter before the next top-level serialization
                 Assert.NotNull(writer.GetFirstPassWriter());
             }
             output.Flush();
--- a/cs/test/core/Util.cs
+++ b/cs/test/core/Util.cs
@@ -385,6 +385,19 @@ namespace UnitTest
             output.Flush();
         }
 
+        public static void SerializeCB2Multiple<T>(T obj, Stream stream, uint times)
+        {
+            var output = new OutputStream(stream, 11);
+            var writer = new CompactBinaryWriter<OutputStream>(output, 2);
+
+            for (uint i = 0; i < times; i++)
+            {
+                Serialize.To(writer, obj);
+                Assert.NotNull(writer.GetFirstPassWriter());
+            }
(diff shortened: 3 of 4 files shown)


chwarr:
You had mentioned wanting to clean some things up before the rest of us reviewed. Are you ready for review yet?

chadwalters:
Yes.

cs/test/core/Util.cs
```diff
@@ -385,6 +385,18 @@ public static void SerializeCB2<T>(T obj, Stream stream)
             output.Flush();
         }
 
+        public static void SerializeCB2Multiple<T>(T obj, Stream stream, uint times)
+        {
+            var output = new OutputStream(stream, 11);
```
tstein: What's this 11?

cs/test/core/BondedTests.cs
```diff
@@ -438,5 +438,17 @@ public void SerializeBonded()
...
+        [Test]
+        public void SerializeBondedCB2Multiple()
+        {
+            var obj = new StructWithBonded();
+
+            var stream = new MemoryStream();
+            stream.SetLength(0);
+            Util.SerializeCB2Multiple(obj, stream, 2);
+
+            Assert.AreEqual(stream.Position, stream.Position);
```
tstein: This would appear to always succeed.

cs/test/core/BondedTests.cs
```diff
@@ -438,5 +438,17 @@ public void SerializeBonded()
...
         }
+
+        [Test]
+        public void SerializeBondedCB2Multiple()
+        {
+            var obj = new StructWithBonded();
+
+            var stream = new MemoryStream();
+            stream.SetLength(0);
+            Util.SerializeCB2Multiple(obj, stream, 2);
```
tstein: Consider also asserting that the CB writer's state is empty after all that, or - maybe more importantly - that its `.getFirstPassWriter()` returns non-null.

cs/test/core/Util.cs
chadwalters: Not entirely sure. Used throughout this file.

cs/test/core/BondedTests.cs
chadwalters: Artifact previous revs. Removed.

cs/test/core/BondedTests.cs
chadwalters: Added now inside util function (no access to writer here.

cs/test/core/BondedTests.cs
```diff
@@ -438,5 +438,15 @@ public void SerializeBonded()
             Assert.IsTrue(Comparer.Equal(poly1, obj.poly[1].Deserialize()));
             Assert.IsTrue(Comparer.Equal(poly2, obj.poly[2].Deserialize<Derived>()));
         }
+
+        [Test]
```
chwarr: Minor: in the first commit, consider adding `[Ignore("Intentionally failing CBv2 regression test before bug fix")]` to this test, as it's known broken. That way, the build and tests still "pass".

cs/src/core/protocols/CompactBinary.cs
```diff
@@ -176,7 +176,11 @@ public IProtocolWriter GetFirstPassWriter()
         {
             if (version == 2)
             {
-                return firstPassWriter.Value;
+                // Only return first pass if not in middle of second pass
+                if (lengths.Count == 0)
```
chwarr: I checked to confirm. `Count` on a `LinkedList` is [constant time](https://msdn.microsoft.com/en-us/library/t56y0h99(v=vs.110).aspx#Anchor_1).

cs/test/core/Util.cs
```diff
@@ -385,6 +385,19 @@ public static void SerializeCB2<T>(T obj, Stream stream)
...
+            var writer = new CompactBinaryWriter<OutputStream>(output, 2);
+
+            for (uint i = 0; i < times; i++)
+            {
+                Serialize.To(writer, obj);
+                Assert.NotNull(writer.GetFirstPassWriter());
```
chwarr: Consider adding a comment about this assertion. Something like:

```csharp
// CBv2 writer should have an active FirstPassWriter before the next top-level serialization.
```

cs/test/core/BondedTests.cs
chadwalters: That seems odd to me. Let's discuss general policy offline.

chwarr:
Looks like AppVeyor timed out installing stack.exe. Re-running.

chwarr:
Are you OK if we squash this when merging instead of fixing up the first commit?

chadwalters:
Yes. Go ahead and squash into a single commit.

PR #569, opened 2017-08-17
java core: Change all constructor version args to int.

tstein:


java/core/src/main/java/com/microsoft/bond/protocol/CompactBinaryReader.java
```diff
@@ -28,7 +28,7 @@ public CompactBinaryReader(InputStream inputStream, short version) {
         }
 
         this.reader = new BinaryStreamReader(inputStream);
-        this.version = version;
+        this.version = (short) version;
```
chwarr: Should there be a range check before the cast?

tstein: We'll throw on line 27 if version isn't 1 or 2.

chwarr: Ahh... The joys of diff folding... Thanks.

tstein:
Removed casts in tests and examples.

PR #570, opened 2017-08-18
java core: Add Bonded.cast().

tstein:
Also refactored Java core's test bondfiles a little.

### DIFF
--- a/java/core/src/main/java/com/microsoft/bond/Bonded.java
+++ b/java/core/src/main/java/com/microsoft/bond/Bonded.java
@@ -129,11 +129,40 @@ public abstract class Bonded<T extends BondSerializable> {
      * conversion was successful or null otherwise. Please note that a successful conversion of Bonded instance does
      * not promise successful deserialization which may still fail when {@link #deserialize(StructBondType)} is called.
      *
+     * If you need to avoid slicing, see {@link #cast(StructBondType)}.
+     *
      * @param toBondType type descriptor for the Bond struct type to convert to
      * @param <U>        the Bond struct type to convert to
      * @return a new Bonded instance constrained to the given type or null if this conversion is not possible
      */
     public abstract <U extends BondSerializable> Bonded<U> convert(StructBondType<U> toBondType);
+
+    /**
+     * Returns this instance of Bonded, cast to a Bonded constrained to another type. The other type must be assignable
+     * from the current Bond type, and this method will throw if it isn't.
+     *
+     * Note that this preserves the underlying Bond type, meaning that a Bonded&lt;T&gt; that has been cast to a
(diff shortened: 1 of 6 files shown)


java/core/src/main/java/com/microsoft/bond/Bonded.java
```diff
@@ -135,6 +135,19 @@
      */
     public abstract <U extends BondSerializable> Bonded<U> convert(StructBondType<U> toBondType);
 
+    public <U extends BondSerializable> Bonded<U> cast(StructBondType<U> castBondType) {
```
chwarr: Can we document this method? The documentation should contrast `cast` with `convert`, and the documentation for `convert` should probably talk about `cast` as well.

java/core/src/main/java/com/microsoft/bond/Bonded.java
```diff
@@ -135,6 +135,19 @@
...
+        if (!(getBondType().isSubtypeOf(castBondType))) {
+            final String err = String.format("Bonded<%s> cannot be cast to Bonded<%s>",
+                getBondType().getQualifiedName(), castBondType.getQualifiedName()
+            );
+            throw new ClassCastException(err);
```
chwarr: `cast` throws, but `convert` returns null? Is the plan to change `convert` to throw too? (It has this behavior in the two object-backed implementations at the moment.) Should that happen now?

java/core/src/test/java/com/microsoft/bond/BondedTest.java
```diff
@@ -70,4 +73,31 @@ public void bondedEquality() throws IOException {
             }
         }
     }
+
+    @Test
+    public void covariance() throws IOException {
```
chwarr: Consider adding a test case for the throwing behavior as well. E.g., a downcast and an unrelated type.

java/core/src/main/java/com/microsoft/bond/Bonded.java
tstein: Had a Javadoc, gitted it away by accident. Restored.

java/core/src/main/java/com/microsoft/bond/Bonded.java
tstein: I don't know. I'm sure `cast()` should throw, because it's supposed to be the same as a real cast. `convert()` returning null seems off to me, but I'm not sure I understand the use case well enough to make proclamations.

If `cast()` fails, you 100% definitely have a broken program. Is the same true for `convert()`?

java/core/src/test/java/com/microsoft/bond/BondedTest.java
tstein: Yeah, good idea.

PR #571, opened 2017-08-18
[Java] Marshal and unmarshal implementation

sigansky:


java/core/src/main/java/com/microsoft/bond/Marshal.java
```diff
@@ -0,0 +1,52 @@
...
+     * @param obj       object to marshal
+     * @param writer    protocol writer
+     * @param <TStruct> the type of the object
+     * @throws IOException if an I/O error occurred
+     */
+    public static <TStruct extends BondSerializable> void marshal(TStruct obj, ProtocolWriter writer)
+            throws IOException {
+        ArgumentHelper.ensureNotNull(obj, "obj");
+        ArgumentHelper.ensureNotNull(writer, "writer");
+        Serializer<TStruct> serializer = new Serializer<TStruct>();
```
chwarr: Minor: use diamond on right-hand side to minimize repitition?
```java
new Serializer<>()
```

chwarr: Minor: looks like `serializer` can be `final`.

java/core/src/main/java/com/microsoft/bond/Marshal.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Marshals a struct object using the given protocol writer.
+     *
+     * @param obj       object to marshal
+     * @param writer    protocol writer
+     * @param <TStruct> the type of the object
+     * @throws IOException if an I/O error occurred
+     */
+    public static <TStruct extends BondSerializable> void marshal(TStruct obj, ProtocolWriter writer)
```
chwarr: I like the consistency between the argument order in `marshal` and `serializer`: `obj` then `writer`.

java/core/src/main/java/com/microsoft/bond/Unmarshal.java
```diff
@@ -0,0 +1,63 @@
...
+     *
+     * @param inputStream input stream
+     * @param bondType    bond type of unmarshalled struct
+     * @param <TStruct>   the type of unmarshalled struct
+     * @return a bonded instance
+     * @throws IOException if an I/O error occurred
+     */
+    public static <TStruct extends BondSerializable> Bonded<TStruct> unmarshal(
+            InputStream inputStream, StructBondType<TStruct> bondType) throws IOException {
+        ArgumentHelper.ensureNotNull(inputStream, "inputStream");
```
chwarr: Is it OK for `bondType` be null on entry?

java/core/src/main/java/com/microsoft/bond/UntaggedProtocolStreamBonded.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Override
+    public void serialize(ProtocolWriter protocolWriter) throws IOException {
+        ArgumentHelper.ensureNotNull(protocolWriter, "protocolWriter");
+        if (this.bondType == null) {
+            throw new InvalidBondDataException("Cannot serialize an unknown struct type within a Bonded instance.");
+        }
+        // deserialize into object using object-backed temporary instance, then serialize it
+        Bonded.fromObject(this.deserialize(), this.bondType).serialize(protocolWriter);
```
chwarr: A temporary transcode. Works for me for now.
What's this code base's style on TODO comments? I expect we'll want to switch this over to transcoding when that is implemented.

java/core/src/main/java/com/microsoft/bond/Marshal.java
sigansky: Diamond operator is supported by Java 7+. We are using 6.

---
In reply to: [134009984](https://github.com/Microsoft/bond/pull/571#discussion_r134009984) [](ancestors = 134009984)

java/core/src/main/java/com/microsoft/bond/Marshal.java
sigansky: I assume this is not an actionable comment. Please let me know otherwise.

---
In reply to: [134010191](https://github.com/Microsoft/bond/pull/571#discussion_r134010191) [](ancestors = 134010191)

java/core/src/main/java/com/microsoft/bond/Unmarshal.java
sigansky: No it's not. The check for null is made in fromProtocolReader, but I can add it here as well.

---
In reply to: [134010489](https://github.com/Microsoft/bond/pull/571#discussion_r134010489) [](ancestors = 134010489)

java/core/src/main/java/com/microsoft/bond/Marshal.java
chwarr: Gotcha. The examples are in Java 8, so I'm used to it there.

java/core/src/main/java/com/microsoft/bond/Unmarshal.java
chwarr: Up to you. You know this style of this code base better than I.

java/core/src/main/java/com/microsoft/bond/Unmarshal.java
```diff
@@ -0,0 +1,63 @@
...
+
+        short protocol = readInt16(inputStream);
+        short version = readInt16(inputStream);
```
tstein: Don't know that we've properly specified this anywhere, but I would think these are both treated as uints. Doesn't matter so much for `protocol`, but I don't think anyone intends for `version` to go negative. Convert to signed int here?

java/core/src/main/java/com/microsoft/bond/Unmarshal.java
```diff
@@ -0,0 +1,63 @@
...
+
+            case ProtocolType.Values.SIMPLE_PROTOCOL:
+                return Bonded.fromProtocolReader(new SimpleBinaryReader(inputStream, version), bondType);
+
+            default:
+                throw new InvalidBondDataException("Unknown protocol type: " + protocol);
+        }
+    }
+
+    private static short readInt16(InputStream inputStream) throws IOException {
```
tstein: Not a huge fan of repeating logic that belongs to `BinaryStreamReader`, especially in what I expect to be a pretty cold path. Any particular reason you don't want to use that?

java/core/src/main/java/com/microsoft/bond/protocol/SimpleBinaryReader.java
```diff
@@ -204,5 +204,11 @@ public void skipWString() throws IOException {
     public InputStream cloneStream() throws IOException {
         return Cloning.cloneStream(this.reader.inputStream);
     }
+
+    @Override
+    public UntaggedProtocolReader cloneProtocolReader() throws IOException {
```
sigansky: Clone is inherited from Object and can't be overridden with this throws.

---
In reply to: [134026347](https://github.com/Microsoft/bond/pull/571#discussion_r134026347) [](ancestors = 134026347)

java/core/src/main/java/com/microsoft/bond/Unmarshal.java
sigansky: BinaryStreamReader is package-private. Making it public isn't desired and consolidating all into a single package is outside the scope.

---
In reply to: [134077034](https://github.com/Microsoft/bond/pull/571#discussion_r134077034) [](ancestors = 134077034)

java/core/src/main/java/com/microsoft/bond/protocol/SimpleBinaryReader.java
tstein: Damn. Good as is.

java/core/src/main/java/com/microsoft/bond/Unmarshal.java
chwarr: Particularly in light of #569 where the version parameters have all been switched to ints.

tstein:
Merged.

PR #572, opened 2017-08-18
[Java] Fixed serialization bugs of bonded and non-optional missing fields

sigansky:


java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -245,6 +245,9 @@ protected final void serializeSomethingField(
             StructBondType.StructField<T> field) throws IOException {
         if (value != null) {
             serializeField(context, value.getValue(), field);
+        } else if (!field.isOptional()) {
+            // throws
```
chwarr: Minor: I find the `Throw.raise...` line to be clear enough on its own, so I'd remove the comment.

java/core/src/main/java/com/microsoft/bond/Bonded.java
```diff
@@ -92,6 +92,17 @@
     public abstract void serialize(ProtocolWriter protocolWriter) throws IOException;
 
     /**
+     * Serializes content of this Bonded instance to using the given context and the default Bond type descriptor.
+     * Throws an exception if there is no default type descriptor associated with this Bonded instance.
+     * This method is called when this Bonded is serialized as a field of another object, as opposed to top-level
+     * serializaton. For the top-level serialization, use the public serialize method.
```
chwarr: Minor: link the "serialize" in the JavaDoc?

java/core/src/main/java/com/microsoft/bond/BondType.java
```diff
@@ -245,6 +245,9 @@ protected final void serializeSomethingField(
             StructBondType.StructField<T> field) throws IOException {
         if (value != null) {
             serializeField(context, value.getValue(), field);
+        } else if (!field.isOptional()) {
```
chwarr: For my understanding, is this the method that will enforce this behavior for strings and collections?

java/core/src/main/java/com/microsoft/bond/Bonded.java
```diff
@@ -92,6 +92,17 @@
...
+     *
+     * @param context the serialization context
+     * @throws IOException if an I/O error occurred during serialization
+     */
+    abstract void serialize(BondType.SerializationContext context) throws IOException;
```
chwarr: Should I expect additional test cases in this PR, or are they coming in a different one?

sigansky: There will be other PR's with test cases covering this as well.

---
In reply to: [134047122](https://github.com/Microsoft/bond/pull/572#discussion_r134047122) [](ancestors = 134047122)

java/core/src/main/java/com/microsoft/bond/BondType.java
sigansky: Yes, it's used to serialize every field that defaults to nothing and is not a Java primitive type.

---
In reply to: [134046697](https://github.com/Microsoft/bond/pull/572#discussion_r134046697) [](ancestors = 134046697)

java/core/src/main/java/com/microsoft/bond/Bonded.java
tstein: The difference in behavior between `serialize(ProtocolWriter)` and `serialize(SerializationContext)` isn't clear from the difference in param types. We already have a way of talking about this - how about making this method `serializeValue(SerializationContext)`?

sigansky: It should be clear from parameters (or documentation) that it's the "this" object that is serialized, not some "value". I can't call it serializeValue since there is no "value" passed. It does exactly what the other "serialize" does, and I'll keep it as is.

---
In reply to: [134343912](https://github.com/Microsoft/bond/pull/572#discussion_r134343912) [](ancestors = 134343912)

java/core/src/main/java/com/microsoft/bond/BondType.java
sigansky: Since this is not an explicit "throw", I want to indicate that this line terminates the method.

---
In reply to: [134045874](https://github.com/Microsoft/bond/pull/572#discussion_r134045874) [](ancestors = 134045874)

PR #574, opened 2017-08-18
[Java] Add Blob wrapper class around byte []

chadwalters:
Added a simple Blob wrapper class that allows for implementation of equals and
hashCode methods based on the contents of the blob.

Update codegen and blob example to reflect this change as well.


java/core/src/main/java/com/microsoft/bond/Blob.java
```diff
@@ -0,0 +1,80 @@
...
+     * Provide access to underlying {@code byte []}
+     * @return the underying {@code byte []}
+     */
+    public byte[] getData()
+    {
+        return this.data;
+    }
+
+    /**
+     * Provide a shallow clone of this {@code Blob} instance. Underlying {@code byte []} will be shared
```
tstein: ~~ Do you need the second space in `{@code byte []}`?

java/core/src/main/java/com/microsoft/bond/Blob.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Creates a new Blob instance wrapping an existing {@code byte []}
+     * @param data
+     * @return new Blob instance wrapping the data array
+     * @throws IllegalArgumentException if the argument is null
+     */
+    public Blob(byte[] data) {
+        if (data == null) {
+            throw new IllegalArgumentException("Argument 'data' must not be null");
+        }
```
sigansky: We have a helper to do that check: ArgumentHelper.ensureNotNull

java/core/src/main/java/com/microsoft/bond/Blob.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Compare for equality by contents of underlying {@code byte []}
+     * @param other object to compare against
+     * @return {@code true} if object is also a Blob and the underlying arrays have equivalent contents
+     */
+    @Override
+    public boolean equals(Object other) {
+        if (this == other) {
+            return true;
+        }
```
sigansky: Arrays.equals already takes care of this minor optimization.

java/core/src/main/java/com/microsoft/bond/Blob.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Provide a shallow clone of this {@code Blob} instance. Underlying {@code byte []} will be shared
+     * @return a shallow clone of this {@code Blob} instance
+     */
+    @Override
+    protected Blob clone() {
+        return new Blob(this.data);
+    }
+
+    private byte[] data;
```
sigansky: >private [](start = 4, length = 7)

final

java/core/src/main/java/com/microsoft/bond/BlobBondType.java
```diff
@@ -72,25 +72,25 @@ public final boolean isGenericType() {
...
-    protected final byte[] newDefaultValue() {
+    protected final Blob newDefaultValue() {
         return DEFAULT_VALUE;
     }
 
     @Override
-    protected final byte[] cloneValue(byte[] value) {
+    protected final Blob cloneValue(Blob value) {
         return value.clone();
     }
```
sigansky: This should return a deep clone, which is not what Blob.clone does. Please see the API of cloneValue in BondType.

examples/java/core/blob/src/main/java/com/microsoft/bond/examples/BlobExample.java
```diff
@@ -3,8 +3,10 @@
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.nio.ByteBuffer;
```
chwarr: Minor: is this import needed anymore?

chadwalters: Removed.

java/core/src/main/java/com/microsoft/bond/Blob.java
chadwalters: Addressed.

java/core/src/main/java/com/microsoft/bond/BlobBondType.java
chadwalters: Addressed.

java/core/src/main/java/com/microsoft/bond/Blob.java
chadwalters: Not exactly. The parameters to Array.equals() are the arrays -- this is the optimization at the level of Object.

java/core/src/main/java/com/microsoft/bond/Blob.java
chadwalters: Addressed.

java/core/src/main/java/com/microsoft/bond/Blob.java
chadwalters: Addressed throughout file.

chadwalters:
CR comments address. Please squash when merging.

chwarr:
Golden tests are failing. Can you re-gen?

PR #577, opened 2017-08-22
Move Java world to org.bondlib.

tstein:
checked:
* gradle and IDEA builds for `core`, `compat`, `gradle-plugin`
* `make check` with GBC and compat tests enabled
* "com.microsoft.bond" doesn't appear anywhere in my tree except for some stale build artifacts

Because the package of the gradle plugin has changed, you'll need to `mvn install` it anew.

In a separate commit, delete the `interop/` directory, which has outlived its usefulness.


chwarr:
There are some .class files under the plug-ins out\ directory. Can you remove them.

java/core/src/test/java/org/bondlib/SomethingFloatTest.java
```diff
@@ -1,9 +1,10 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-package com.microsoft.bond;
+package org.bondlib;
 
-import com.microsoft.bond.helpers.FloatingPointHelper;
+import org.bondlib.helpers.FloatingPointHelper;
+import org.junit.Assert;
```
chwarr: The renaming has also adjusted imports here. Can we do that in a different commit?

java/core/src/main/java/org/bondlib/UntaggedProtocolStreamBonded.java
```diff
@@ -37,7 +37,7 @@ public void serialize(ProtocolWriter protocolWriter) throws IOException {
             throw new InvalidBondDataException("Cannot serialize an unknown struct type within a Bonded instance.");
         }
         // deserialize into object using object-backed temporary instance, then serialize it
-        Bonded.fromObject(this.deserialize(), this.bondType).serialize(protocolWriter);
+        fromObject(this.deserialize(), this.bondType).serialize(protocolWriter);
```
chwarr: If we want to remove the `Bonded.`, can we do it in a different commit?

java/core/src/main/java/org/bondlib/TaggedProtocolStreamBonded.java
```diff
@@ -46,7 +46,7 @@ void serialize(BondType.SerializationContext context) throws IOException {
             throw new InvalidBondDataException("Cannot serialize an unknown struct type within a Bonded instance.");
         }
         // deserialize into object using object-backed temporary instance, then serialize it
-        Bonded.fromObject(this.deserialize(), this.bondType).serialize(context);
+        fromObject(this.deserialize(), this.bondType).serialize(context);
```
chwarr: Can we remove the `Bonded.` in a different commit?

java/gradle-plugin/.idea/modules/bond-gradle.iml
```diff
@@ -34,6 +35,15 @@
         <SOURCES />
       </library>
     </orderEntry>
+    <orderEntry type="module-library" scope="PROVIDED">
+      <library name="Gradle: gradle-installation-beacon-3.5">
+        <CLASSES>
+          <root url="jar:///usr/local/Cellar/gradle/3.5/libexec/lib/gradle-installation-beacon-3.5.jar!/" />
```
chwarr: Are these "Cellar" changes needed?

chwarr:
Also, can you reword the removal commit to be more like "Remove no longed needed, handwritten compat driver"?

java/core/src/test/java/org/bondlib/SomethingFloatTest.java
tstein: Fixed.

java/core/src/main/java/org/bondlib/UntaggedProtocolStreamBonded.java
tstein: Fixed.

java/core/src/test/java/org/bondlib/SomethingFloatTest.java
tstein: Fixed.

java/core/src/main/java/org/bondlib/TaggedProtocolStreamBonded.java
tstein: Fixed.

java/gradle-plugin/.idea/modules/bond-gradle.iml
tstein: Definitely not.

java/gradle-plugin/.idea/modules/bond-gradle.iml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<module external.linked.project.id="bond-gradle" external.linked.project.path="$MODULE_DIR$/../.." external.root.project.path="$MODULE_DIR$/../.." external.system.id="GRADLE" external.system.module.group="com.microsoft.bond" external.system.module.version="1.0" type="JAVA_MODULE" version="4">
+<module external.linked.project.id="bond-gradle" external.linked.project.path="$MODULE_DIR$/../.." external.root.project.path="$MODULE_DIR$/../.." external.system.id="GRADLE" external.system.module.group="org.bondlib" external.system.module.version="1.0" type="JAVA_MODULE" version="4">
   <component name="NewModuleRootManager">
-    <output url="file://$MODULE_DIR$/../../build/classes/main" />
-    <output-test url="file://$MODULE_DIR$/../../build/classes/test" />
+    <output url="file://$MODULE_DIR$/../../out/production/classes" />
```
chadwalters: Is this change from "build" to "out" intentional?

tstein: Wasn't originally, but I'm consciously surrendering and making that change here:
* I have yet to find a way to get IDEA to stop changing it back.
* I don't know how to get IDEA to even temporarily "compile" core's `alphabet.txt` test resource into `build/` instead of `out/`, so there is going to be a `core/out/` either way.
* `java/.gitignore` includes `*/out` as of this PR.

There wasn't a strong reason for trying to share a build directory in the first place, and some internet folks seem to think it's actually a [bad idea](https://youtrack.jetbrains.com/issue/IDEA-91952).

chadwalters: OK, I agree that we should not fight uphill against IDEA. We may want to look at changing things elsewhere as well (e.g.: Java examples) but that shouldn't hold this PR.

PR #579, opened 2017-08-22
travis: CI all Java components and run compat tests.

tstein:
With these changes, Java gets built and tested, and all examples are run. One had a placeholder `assert false` that we should tackle immediately.

Everything works except for three cases in compat related to schemas and JSON. We need to get this green to get the benefit of CI, so I've disabled them in two places in the third commit. (One to avoid creating Java compat tests for `json` and `schema`, and another to do nothing with Java when the `gbc schema` test case comes through.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,14 +22,27 @@ env:
     - FLAVOR="cpp-grpc" BOOST="1.64.0"
     - FLAVOR="cpp-grpc" BOOST="1.63.0"
     - FLAVOR="hs"
+
+# The Java build is unusually difficult to shoehorn into csharp. Changes to
+# JAVA_HOME don't seem to take effect, and even if you manually symlink
+# /usr/lib/jvm/default-java to the right JDK, you get weird Groovy errors from
+# gradle. matrix: include: makes it possible to override the language setting
+# for specific builds.
+matrix:
+    include:
+        - language: java
+          env: FLAVOR="java"
+
 addons:
   apt:
     sources:
         - ubuntu-toolchain-r-test
     packages:
+        - wget
         - g++-4.8
         - ccache
-        - wget
+        - oracle-java8-installer
+        - oracle-java8-set-default
 
 before_install:
     - mkdir -p ~/.local/bin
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -75,13 +88,34 @@ script:
...
+
     - if [ "$FLAVOR" == "hs" ]; then cmake -DBOND_SKIP_CORE_TESTS=TRUE -DBOND_ENABLE_GRPC=FALSE ..; fi
     - if [ "$FLAVOR" == "hs" ]; then make gbc-tests; fi
     - if [ "$FLAVOR" == "hs" ]; then cd ../compiler; fi
     - if [ "$FLAVOR" == "hs" ]; then ../build/compiler/build/gbc-tests/gbc-tests; fi
+
+    - if [ "$FLAVOR" == "java" ]; then cmake -DBOND_SKIP_GBC_TESTS=TRUE -DBOND_SKIP_CORE_TESTS=TRUE -DBOND_ENABLE_GRPC=FALSE ..; fi
+    - if [ "$FLAVOR" == "java" ]; then make gbc; fi
+    - if [ "$FLAVOR" == "java" ]; then export PATH=`pwd`/compiler/build/gbc:$PATH; fi
+    - if [ "$FLAVOR" == "java" ]; then cd ../java/gradle-plugin; gradle build; fi
```
chwarr: Minor: would it be clearer if we used absolute paths, like `cd '$TRAVIS_BUILD_DIR/java/gradle-pluging'`? (where [`$TRAVIS_BUILD_DIR`](https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables) is "The absolute path to the directory where the repository being built has been copied on the worker.")

.travis.yml
```diff
@@ -75,13 +88,34 @@ script:
...
+    - if [ "$FLAVOR" == "java" ]; then mvn install:install-file -Dfile=build/libs/bond-gradle-1.0.jar -DgroupId=org.bondlib -DartifactId=gradle -Dversion=1.0 -Dpackaging=jar; fi
```
chwarr: Is there any way to decouple the version numbers here so we don't have to remember to rev them when we change the version?

.travis.yml
```diff
@@ -75,13 +88,34 @@ script:
...
+    - if [ "$FLAVOR" == "java" ]; then cd ../core; gradle build; fi
+    - if [ "$FLAVOR" == "java" ]; then cd ../compat; gradle build; fi
```
chwarr: What happens if these gradle builds? We `set -e` below, which makes me think that the Travis build won't fail until much, much later.

.travis.yml
tstein: Good question. I intentionally did it all with relative paths because the `cs` and `hs` flavors are written that way - it's all the same problem, but the `java` section is admittedly the biggest. I find it plenty clear as-is, but I wrote it; if you think it would help to make some or all of them absolute via the envvar, I'll make the change.

.travis.yml
tstein: The thing this needs to match is the `version '1.0'` in `gradle-plugin/build.gradle`. The best idea that I can come up with is to give this script a one-liner to grep the version out of the gradle file. Would that work for you?

.travis.yml
tstein: The build script will continue, but the build job will fail. This is how the other flavors work, though Java is obviously longer. This comment has unleashed a torrent of thoughts on shell and travis mechanics to justify those six characters.

AFAICT, the way travis works is that it pumps each of our lines into a single shell instance, wrapped inside a helper function. The wrapper provides the folding, pretty green/red text in the log, and implements `script_success ||= line_success` behavior. With flow control (`if [ "$FLAVOR"...` or semicolon-separated statements, the return code of the line is the return code of the last evaluated statement. (I wanted and had `cd core; gradle build; cd ..`, but then the line returns 0 even if gradle fails.)

The `for` loop to hit all examples was more annoying. Shell loops evaluate to the return code of the last statement in the last loop, so as long as the last example succeeds, the line with the loop succeeds from travis's perspective. `set -e` modifies the top-level shell so that as soon as any command, in any location, returns nonzero, the shell bails out with that retval. The effect is that the first failed example fails the build as you would expect. (Pedantic note: it also deprives the wrapper of its opportunity to do the nice folding and red/green stuff, but the gradle output from the failure will be the very last thing in the log.)

The right answer here is to have a process that programmatically discovers examples, runs them, and implements `examples_success ||= last_example_success`, which is what cmake will do once these examples are properly wired into it. I considered making a throwaway script, but this seemed like an easy and safe stopgap.

.travis.yml
chwarr: Can you add a commit that switches all (the Java ones) over to being absolute. It's something I'll need to see to know for sure.

.travis.yml
chwarr: `grep` works for me.

.travis.yml
tstein: Done. See 8b11042.

.travis.yml
tstein: I did what I meant, not what I `sed`. B|

(see 2598409)

.travis.yml
chwarr: Agreed that the build scripts could stand some refactoring. Not as part of this PR, though.

.travis.yml
chwarr: Yeah, I like the env var approach better. Thanks.

examples/java/core/protocol_versions/src/main/java/org/bondlib/ProtocolVersions.java
```diff
@@ -57,6 +59,16 @@ public static void main(final String[] args) throws IOException {
...
 
-        assert false : "Add marshaling example here";
+        {
+            // Here, we Marshal to CompactBinary v2.
+            final ByteArrayOutputStream output = new ByteArrayOutputStream();
+            final CompactBinaryWriter writer = new CompactBinaryWriter(output, 2);
+            Marshal.marshal(obj, writer);
+
+            final ByteArrayInputStream input = new ByteArrayInputStream(output.toByteArray());
+            final Struct obj2 = Unmarshal.unmarshal(input, Struct.BOND_TYPE).deserialize();
```
chwarr: The C# example has a useful comment above the call to unmarshal. Can we add a similar one?

                // The protocol and version are determined from the payload
                // itself.

tstein: Done.

PR #580, opened 2017-08-24
java core: Add regression test for CBv2 double bonded bug.

tstein:
Locally reverted the bug fix from 8a608fbe and confirmed that the test fails with an `InvalidBondDataException`.

Additionally, accept IDEA's insistence on `out/` in a separate commit.

### DIFF
--- a/java/core/src/test/bond/bonded.bond
+++ b/java/core/src/test/bond/bonded.bond
@@ -2,6 +2,10 @@ import "common.bond"
 
 namespace org.bondlib.test
 
+struct HasBondedField {
+    0: bonded<Base> bondedField;
+}
+
 struct BondedCollection {
     0: list<bonded<Base>> bondeds;
 }
--- a/java/core/src/test/java/org/bondlib/protocol/CompactBinaryWriterTest.java
+++ b/java/core/src/test/java/org/bondlib/protocol/CompactBinaryWriterTest.java
@@ -3,12 +3,20 @@
 
 package org.bondlib.protocol;
 
+import org.bondlib.Bonded;
+import org.bondlib.Deserializer;
 import org.bondlib.ProtocolType;
+import org.bondlib.Serializer;
+import org.bondlib.test.Base;
+import org.bondlib.test.HasBondedField;
+import org.junit.Assert;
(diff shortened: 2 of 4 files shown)


java/core/src/test/java/org/bondlib/protocol/CompactBinaryWriterTest.java
```diff
@@ -111,4 +119,46 @@ public void testFirstPassWriterWriteFieldOmittedDoesNotWrite() {
...
+        final Serializer<BondedBase> serializer = new Serializer<BondedBase>();
+        for (int i = 0; i < objCount; i++) {
+            serializer.serialize(bondedBasesIn.get(i), writer);
+        }
+
+        final ByteArrayInputStream bais = new ByteArrayInputStream(baos.toByteArray());
+        final CompactBinaryReader reader = new CompactBinaryReader(bais, 2);
+        final Deserializer<BondedBase> deserializer = new Deserializer<BondedBase>(BondedBase.BOND_TYPE);
+        final List<Base> basesOut = new ArrayList<Base>();
+        for (int i = 0; i < objCount; i++) {
```
chwarr: Consider reading until there's no more data in the input stream. This will also ensure that we successfully read everything that was written.

If you want to assert that there's nothing left to read, that would work too.

java/core/src/test/bond/bonded.bond
```diff
@@ -2,6 +2,10 @@ import "common.bond"
 
 namespace org.bondlib.test
 
+struct BondedBase {
```
chwarr: Consider "HasBondedField" and "bondedField" instead. The base/derived part isn't relevant here, but "BondedBase" makes me think that this is a test of inheritance.

java/core/src/test/java/org/bondlib/protocol/CompactBinaryWriterTest.java
tstein: Asserted.

java/core/src/test/bond/bonded.bond
tstein: Done.

PR #587, opened 2017-08-31
java codegen: Don't emit backslashes in comments.

tstein:


compiler/Main.hs
```diff
@@ -182,7 +182,17 @@ javaCodegen Java {..} = do
...
-                          else (commonHeader "//" bondFile javaFile <> code)
+                          else (commonHeader "//" safeBondFile safeJavaFile <> code)
+                              where
+                                  -- javac will always treat "\u" as the start
+                                  -- of a unicode escape sequence, and will
+                                  -- error out if it isn't followed by a valid
+                                  -- code. This breaks compilation of generated
+                                  -- code if either path has components that
+                                  -- start with u.
+                                  backToForward s = LT.unpack $ LT.replace (LT.pack "\\") (LT.pack "/") (LT.pack s)
```
chwarr: How about `backToForward s = map (\c -> if c == '\\' then '/'; else c) s` so we don't have to do all the packing and packing?

tstein: Done.

tstein:
Failure was due to a timeout in a cpp-core build. Restarted to get my green, but this change CIs okay.

PR #596, opened 2017-09-06
[C++] Check for memory allocation overflow errors

chadwalters:
Detect overflow errors related to memory allocation and memory access,
specifically in: bond::OutputMemoryStream, bond::blob, and
bond::detail::SimpleArray.


cpp/inc/bond/core/blob.h
```diff
@@ -157,7 +176,10 @@ class blob
     /// the end of the buffer
     blob range(uint32_t offset) const
     {
-        BOOST_ASSERT(offset <= _length);
+        if (offset > _length)
+        {
+            throw std::invalid_argument("Offset too large; must be less than length of blob");
```
lalo: less than or equal if we follow the condition of the if

cpp/inc/bond/stream/output_buffer.h
```diff
@@ -246,6 +250,12 @@ class OutputMemoryStream
                 _blobs.push_back(blob(_buffer, _rangeOffset, _rangeSize));
             }
 
+            // cap buffer at 2GB
+            if (_bufferSize >= ((std::numeric_limits<uint32_t>::max)() >> 1))
```
tstein: Could this be `_bufferSize > (std::numeric_limits<int32_t>::max)()`? Seems like it should be `>`, not `>=`, either way.

cpp/inc/bond/core/blob.h
chadwalters: Fair enough.

cpp/inc/bond/stream/output_buffer.h
chadwalters: `>` makes sense. I'd rather leave things in terms of `uint32_t` -- seems more semantically coherent.

cpp/inc/bond/core/detail/check_overflow.h
```diff
@@ -0,0 +1,32 @@
...
+#include <stdexcept>
+
+namespace bond
+{
+namespace detail
+{
+
+    inline void check_add_overflow(const char* ptr, uint32_t offset)
+    {
+        if ((ptr + offset) < ptr)
```
chwarr: I'm not sure whether this arithmetic can trigger UB or not... Let me dig into this some more.

cpp/inc/bond/core/blob.h
```diff
@@ -54,6 +57,8 @@ class blob
           _content(_buffer.get() + offset),
           _length(length)
     {
+        bond::detail::check_add_overflow(_buffer.get(), offset);
```
chwarr: On line 57 we've already performed the potentially overflowing addition. I think we want a function that takes a pointer and an offset, throws if it would be outof range, and returns a value otherwise. Then, we can do something like this:

```
blob(const boost::shared_ptr<const char[]>& buffer, uint32_t offset, uint32_t length)
    : _buffer(buffer),
      _content(detail::safe_add(buffer.get(), offset)),
     _length(length)
    {
        check_add_overflow(_content, length);
    }
```

cpp/inc/bond/core/detail/check_overflow.h
```diff
@@ -0,0 +1,36 @@
...
+#include <boost/static_assert.hpp>
+
+namespace bond
+{
+namespace detail
+{
+
+    inline void check_add_overflow(const char* ptr, uint32_t offset)
+    {
+        BOOST_STATIC_ASSERT(sizeof(const char *) == sizeof(std::uintptr_t));
```
chwarr: Can you add a comment saying that the numeric limit logic below requires that uintptr_t be exactly the same size as a pointer?

cpp/inc/bond/core/detail/check_overflow.h
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#pragma once
+
+#include <limits>
```
chwarr: Minor: we also need stdint.h/cstdint

cpp/inc/bond/protocol/detail/simple_array.h
```diff
@@ -3,6 +3,9 @@
 
 #pragma once
 
+#include <limits>
+#include <stdexcept>
```
chwarr: Minor: need stdint.h/cstdint too.

cpp/inc/bond/core/detail/check_overflow.h
```diff
@@ -0,0 +1,36 @@
...
+        BOOST_STATIC_ASSERT(sizeof(const char *) == sizeof(std::uintptr_t));
+
+        std::uintptr_t uintptr = reinterpret_cast<std::uintptr_t>(ptr);
+        if (((std::numeric_limits<uintptr_t>::max)() - offset) < uintptr)
+        {
+            throw std::overflow_error("Overflow on addition");
+        }
+    }
+
+    inline void check_add_overflow(uint32_t lhs, uint32_t rhs)
```
chwarr: Can we make this a template for any unsigned T and then drop the static_cast in output_buffer.h?

cpp/inc/bond/core/detail/check_overflow.h
```diff
@@ -0,0 +1,36 @@
...
+{
+
+    inline void check_add_overflow(const char* ptr, uint32_t offset)
+    {
+        BOOST_STATIC_ASSERT(sizeof(const char *) == sizeof(std::uintptr_t));
+
+        std::uintptr_t uintptr = reinterpret_cast<std::uintptr_t>(ptr);
+        if (((std::numeric_limits<uintptr_t>::max)() - offset) < uintptr)
+        {
+            throw std::overflow_error("Overflow on addition");
```
chwarr: Consider: "offset caused pointer to overflow"?

lalo:
Maybe a negative unit test would be helpful seeing by how this implementation is finicky.

cpp/inc/bond/core/blob.h
chadwalters: Good point.

cpp/inc/bond/core/detail/check_overflow.h
chadwalters: Added.

cpp/inc/bond/core/detail/check_overflow.h
chadwalters: OK

cpp/inc/bond/core/detail/check_overflow.h
chadwalters: OK.

cpp/inc/bond/protocol/detail/simple_array.h
chadwalters: OK.

chadwalters:
I made one more tweak so that the two unsigned integers types don't have to be the same (the first one must be larger than the second).

I think this is ready to go now. Please squash the commits together when merging.

Issue #536, opened 2017-07-21
Address Clang build warnings from newer versions of Clang


chwarr:
Newer versions of Clang than what we're running in our CI systems are issuing warnings about "binding dereferenced null pointer to reference has undefined behavior"

We've gotten reports of this from Apple LLVM version 8.1.0 (clang-802.0.42).

The correct fix may involve adding an additional overload of `Apply` that doesn't take an instance of a `T`.

The quick work around is the edit the `Config.cmake` file to add `-Wno-null-dereference` to the compiler-specific block for the compiler involved, [like](https://github.com/Microsoft/bond/blob/19078031691af9f55527be0816969fc5c38c4837/cmake/Config.cmake#L149-L156):

```
diff --git a/cmake/Config.cmake b/cmake/Config.cmake
index cf82165e..90b06868 100644
--- a/cmake/Config.cmake
+++ b/cmake/Config.cmake
@@ -152,6 +152,7 @@ cxx_add_compile_options(AppleClang
     --std=c++11
     -Wall
     -Werror
+    -Wno-null-dereference
     -Wno-unknown-warning-option
     -Wno-unused-local-typedefs)
```

Here are the warning details:

```
/Users/xxx/dev/bond/cpp/inc/bond/core/schema.h:303:52: error: binding dereferenced null pointer to reference has undefined behavior
      [-Werror,-Wnull-dereference]
    Apply(InitSchemaDef(*s), static_cast<const T&>(*static_cast<T*>(0)));
                                                   ^~~~~~~~~~~~~~~~~~~
/Users/xxx/dev/bond/cpp/inc/bond/core/schema.h:303:52: error: binding dereferenced null pointer to reference has undefined behavior
      [-Werror,-Wnull-dereference]
    Apply(InitSchemaDef(*s), static_cast<const T&>(*static_cast<T*>(0)));
                                                   ^~~~~~~~~~~~~~~~~~~
/Users/xxx/dev/bond/cpp/inc/bond/core/schema.h:267:33: note: in instantiation of member function 'bond::detail::SchemaCache<bond::StructDef,
      void>::AppendStructDef' requested here
        detail::SchemaCache<T>::AppendStructDef(&_schema);
                                ^
/Users/xxx/dev/bond/cpp/inc/bond/core/schema.h:251:27: note: in instantiation of function template specialization
      'bond::InitSchemaDef::GetStructDef<bond::StructDef>' requested here
[...]

PR #598, opened 2017-09-07
c++ clang: Compile -Wno-null-dereference.

tstein:
#536 is now breaking travis CI. The right answer is probably a change to the underlying code, but this will suppress what are, for now, spurious build failures.

### DIFF
--- a/cmake/Config.cmake
+++ b/cmake/Config.cmake
@@ -152,6 +152,7 @@ cxx_add_compile_options(Clang
     --std=c++11
     -Wall
     -Werror
+    -Wno-null-dereference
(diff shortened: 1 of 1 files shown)


tstein:
Travis has passed with [clang 3.9.0](https://travis-ci.org/Microsoft/bond/jobs/273026256#L51), which includes the suppressed warning.

chwarr:
Merged into master in commit ec14c17ae66caed11880031b049efb39c4934f40.
I'm trying to get a jvm + a merge from master as of ec14c17ae66caed11880031b049efb39c4934f40 building right now, but GitHub appears to be delayed in webhook requests...

PR #607, opened 2017-09-11
[Java] Added helpers for randomized unit tests with some tests

sigansky:
There will be more test Bond files added later.


java/core/src/test/java/org/bondlib/StructGenerator.java
```diff
@@ -0,0 +1,408 @@
...
+            // shouldn't happen
+            throw new RuntimeException(ex);
+        }
+    }
+
+    // helper to generate collection (vector/list/set) elements
+    @SuppressWarnings("unchecked")
+    private void generateCollectionElements(
+            Collection collection, BondType elementType, GenerationContext context) {
+        if (!context.hasReachedStructRecursionDepthLimit(elementType, context)) {
```
tstein: The idea is that if you're make a `T` inside of a collection instead of as a member, it still counts for depth limits? This doesn't increment the recursion because it's counting on `generateValue` to do that?

java/core/src/test/java/org/bondlib/StructGenerator.java
```diff
@@ -0,0 +1,408 @@
...
+        return this.rand.nextDouble() < probability;
+    }
+
+    // helper to generate a pseudo-random integer with the given number of bits (optionally signed)
+    private long generateIntValue(int bitCount, boolean isSigned) {
+        // Given number N randomly chosen between 1 and bitCount with uniform probability, determine the
+        // magnitude of the result as an unsigned value of N bits where bits 0 to N-1 are randomly chosen
+        // individually (with uniform probability), and all the other bits are set to 0. Then optionally
+        // negate the result with 50% probability.
+        // This algorithm improves coverage of smaller integers for variable-length encoding.
```
tstein: I agree with this algorithm, but this comment didn't help much with getting there. How about putting the fact that it isn't uniform and the justification for that up front?
`
We want uniform distribution over magnitudes of numbers, rather than values, to exercise smaller varints as frequently as larger ones. Therefore, we first pick a random length up to bitCount, then a random number that fits in that number of bits. If isSigned, we negate that number with 50% probability. 
`

Additionally, because this is critical to understanding the meaning and use of `generateIntValue`, this should be a Javadoc. It doesn't matter that it's private - making it a Javadoc will allow other devs to see it in tooltips or whatever the IDE has instead of having to leave the callsite.

java/core/src/test/java/org/bondlib/StructGenerator.java
```diff
@@ -0,0 +1,408 @@
...
+        int n = this.rand.nextInt(bitCount) + 1;
+        long result = 0;
+        for (int i = 0; i < n; ++i) {
+            if (this.rand.nextBoolean()) {
+                result |= (1 << i);
+            }
+        }
```
tstein: Any reason this can't be
```
int resultBits = this.rand.nextInt(bitCount) + 1;
long result = this.rand.nextInt(1 << resultBits);
```
?

java/core/src/test/java/org/bondlib/StructGenerator.java
```diff
@@ -0,0 +1,408 @@
...
+
+    // generates values for and sets fields of the given Bond object; recurses to the base type
+    private void generateStructFields(
+            StructBondType bondType, BondSerializable structValue, GenerationContext context) {
+        // recurse to the base type
+        if (bondType.getBaseStructType() != null) {
+            this.generateStructFields(bondType.getBaseStructType(), structValue, context);
+        }
+        // generate fields for the current type
+        StructBondType.StructField[] bondFields = bondType.geStructFields();
```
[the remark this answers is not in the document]
tstein: Looks like this is an existing typo in `StructBondType`. Please fix it in a separate commit here or a new PR.

java/core/src/test/java/org/bondlib/StructGenerator.java
```diff
@@ -0,0 +1,408 @@
...
+        // This algorithm improves coverage of smaller integers for variable-length encoding.
+        int n = this.rand.nextInt(bitCount) + 1;
+        long result = 0;
+        for (int i = 0; i < n; ++i) {
+            if (this.rand.nextBoolean()) {
+                result |= (1 << i);
+            }
+        }
+        if (isSigned && this.rand.nextBoolean()) {
+            result = -result;
```
tstein: I don't think it's possible for this to generate `INT_MIN`, etc. Defer to you on whether this matters.

java/core/src/test/java/org/bondlib/StructGenerator.java
sigansky: What is the typo?

---
In reply to: [138759802](https://github.com/Microsoft/bond/pull/607#discussion_r138759802) [](ancestors = 138759802)

java/core/src/test/java/org/bondlib/StructGenerator.java
sigansky: generateValue is general and will call generateStruct if the element is a struct, which will increment depth.

---
In reply to: [138755007](https://github.com/Microsoft/bond/pull/607#discussion_r138755007) [](ancestors = 138755007)

java/core/src/test/java/org/bondlib/StructGenerator.java
sigansky: Won't work cleanly for unsigned. The current code relies on bit manipulations rather than signed arithmetic, which does the trick.

---
In reply to: [138759100](https://github.com/Microsoft/bond/pull/607#discussion_r138759100) [](ancestors = 138759100)

java/core/src/test/java/org/bondlib/StructGenerator.java
sigansky: No it's not.

---
In reply to: [138765577](https://github.com/Microsoft/bond/pull/607#discussion_r138765577) [](ancestors = 138765577)

java/core/src/test/java/org/bondlib/StructGenerator.java
```diff
@@ -0,0 +1,408 @@
...
+
+/**
+ * Generates a struct of a given type with randomly set fields.
+ */
+public final class StructGenerator {
+
+    private final StructGeneratorSettings settings;
+    private final Random rand;
+
+    // pool of special float values many of which can't be generated by a random number generator
```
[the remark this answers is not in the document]
sigansky: Not sure what you mean. Are you complaining about a comma missing, or something else?

---
In reply to: [138748209](https://github.com/Microsoft/bond/pull/607#discussion_r138748209) [](ancestors = 138748209)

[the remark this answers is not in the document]
tstein: Yes, just the missing comma. I'm noting that it did actually throw me off.

java/core/src/test/java/org/bondlib/StructGenerator.java
tstein: Not sure I understand, but it's fine as is.

java/core/src/test/java/org/bondlib/RandomRoundTripTest.java
```diff
@@ -0,0 +1,162 @@
...
+        // multiple iterarions
+        for (int i = 0; i < testIterationCount; ++i) {
+            TStruct structValue = structGenerator.generateStruct(bondType);
+
+            // Fast V1
+            try {
+                TestHelper.marshalUnmarshalAndCompare(structValue, ProtocolType.FAST_PROTOCOL, 1);
+            } catch (IOException e) {
+                // exception shouldn't happen
+                Assert.fail("Error round-trip using Fast V1: " + e.toString());
```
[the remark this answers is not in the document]
sigansky: OK

---
In reply to: [138251059](https://github.com/Microsoft/bond/pull/607#discussion_r138251059) [](ancestors = 138251059)

java/core/src/test/java/org/bondlib/StructGenerator.java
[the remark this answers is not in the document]
sigansky: Fixed.

---
In reply to: [138748706](https://github.com/Microsoft/bond/pull/607#discussion_r138748706) [](ancestors = 138748706)

java/core/src/test/java/org/bondlib/StructGenerator.java
```diff
@@ -0,0 +1,408 @@
...
+        }
+        // generate fields for the current type
+        StructBondType.StructField[] bondFields = bondType.geStructFields();
+        for (StructBondType.StructField bondField : bondFields) {
+            if (!bondField.isOptional() || this.testProbability(this.settings.probabilityOfSettingOptionalField)) {
+                Field classField = getClassField(bondField);
+                Object value = this.generateValue(bondField.getFieldType(), context);
+                if (bondField.isDefaultNothing()) {
+                    // wrap the value, use the Class of the field to figure out whether
+                    // the wrapper is for Java primitive types of for Object types
```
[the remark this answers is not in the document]
sigansky: Fixed

---
In reply to: [138748911](https://github.com/Microsoft/bond/pull/607#discussion_r138748911) [](ancestors = 138748911)

java/core/src/test/java/org/bondlib/StructGenerator.java
```diff
@@ -0,0 +1,408 @@
...
+            Map mapValue = mapBondType.newDefaultValue();
+            this.generateMapElements(mapValue, mapKeyBondType, mapValueBondType, context);
+            returnValue = mapValue;
+        } else if (bondTypeClass == BondedBondType.class) {
+            // bonded<>
+            BondedBondType bondedBondType = (BondedBondType) bondType;
+            StructBondType declaredStructBondType = bondedBondType.getValueType();
+            ArrayList<StructBondType> allowedPolymorphicStructBondTypes =
+                    this.getPolymorphicStructBondTypesFor(declaredStructBondType);
+            int index = this.rand.nextInt(allowedPolymorphicStructBondTypes.size());
```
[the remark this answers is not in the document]
sigansky: OK

---
In reply to: [138749998](https://github.com/Microsoft/bond/pull/607#discussion_r138749998) [](ancestors = 138749998)

java/core/src/test/java/org/bondlib/StructGenerator.java
sigansky: Fixed this by mapping 0 to most negative number, when the sign is negative.

---
In reply to: [138771867](https://github.com/Microsoft/bond/pull/607#discussion_r138771867) [](ancestors = 138771867,138765577)

java/core/src/test/java/org/bondlib/StructGenerator.java
sigansky: Made this method (and a couple others) public and added javadoc.

---
In reply to: [138758973](https://github.com/Microsoft/bond/pull/607#discussion_r138758973) [](ancestors = 138758973)

java/core/src/test/java/org/bondlib/StructGenerator.java
sigansky: Added comma.

---
In reply to: [138773422](https://github.com/Microsoft/bond/pull/607#discussion_r138773422) [](ancestors = 138773422)

java/core/src/test/java/org/bondlib/RandomRoundTripTest.java
```diff
@@ -0,0 +1,264 @@
...
+        settings.polymorphicBondTypes.addAll(Arrays.asList(polymorphicBondTypes));
+        testRoundTrip(bondType, randomSeed, settings);
+    }
+
+    private static <TStruct extends BondSerializable> void testRoundTrip(
+            StructBondType<TStruct> bondType, long randomSeed, StructGeneratorSettings settings)
+            throws IOException {
+        StructGenerator structGenerator = new StructGenerator(settings, randomSeed);
+
+        // multiple iterarions
```
tstein: `iterarions` is still here

java/core/src/test/java/org/bondlib/RandomRoundTripTest.java
```diff
@@ -0,0 +1,264 @@
...
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.EnumMap;
+
+import static org.junit.Assert.*;
+
+public class RandomRoundTripTest {
+
+    private static final int testIterationCount = 1;
```
tstein: Just noticed this is 1. What's the point of having a tunable and a loop here if we're setting it to 1 now? When would we increase it and do more rounds?

java/core/src/test/java/org/bondlib/StructGenerator.java
```diff
@@ -0,0 +1,443 @@
...
+        long result = 0;
+        for (int currentBitIndex = 0; currentBitIndex < bitCount; ++currentBitIndex) {
+            if (this.rand.nextBoolean()) {
+                result |= (1 << currentBitIndex);
+            }
+        }
+        if (isSigned && this.rand.nextBoolean()) {
+            if (result == 0) {
+                // use the most negative number of the given width
+                result |= ~0L << bitCount;
```
tstein: A bit odd to `|=` instead of just assign here. Either way.

java/core/src/test/java/org/bondlib/RandomRoundTripTest.java
sigansky: When I make it maxint and head into the weekend.

sigansky: What number would you like?

PR #608, opened 2017-09-12
runtime schema-driven untagged deserialization

tstein:


java/core/src/main/java/org/bondlib/StructBondType.java
```diff
@@ -175,14 +175,16 @@ protected abstract void deserializeStructFields(
...
+     * fields.
      *
      * @param context contains the runtime context of the deserialization
+     * @param typeDef the TypeDef of this struct
      * @param value   the value to deserialize into
      * @throws IOException if an I/O error occurred
      */
     protected abstract void deserializeStructFields(
-        UntaggedDeserializationContext context, TStruct value) throws IOException;
+        UntaggedDeserializationContext context, TypeDef typeDef, TStruct value) throws IOException;
```
sigansky: >TypeDef typeDef,  [](start = 48, length = 17)

I believe this is cleaner with StructDef parameter. TypeDef is too generic and doesn't always map to a struct. Further, generated implementation only used the TypeDef to get StructDef from it.

cs/src/core/RuntimeSchema.cs
```diff
@@ -34,25 +35,37 @@ public RuntimeSchema(SchemaDef schema)
 
         public RuntimeSchema GetBaseSchema()
         {
-            Debug.Assert(IsStruct);
+            if (!IsStruct)
+            {
+                throw new InvalidOperationException();
```
sigansky: >InvalidOperationException(); [](start = 26, length = 28)

I made this comment in Chad's PR, which got lost. Please add exception error message here and in other places in this file.

examples/java/core/runtime_schema/src/main/java/org/bondlib/examples/RuntimeSchemaExample.java
```diff
@@ -0,0 +1,44 @@
...
+        final Serializer<SchemaDef> serializer = new Serializer<>();
+        serializer.serialize(Example.BOND_TYPE.buildSchemaDef(), writer);
+
+        final ByteArrayInputStream input = new ByteArrayInputStream(output.toByteArray());
+
+        final CompactBinaryReader reader = new CompactBinaryReader(input, 1);
+        final Deserializer<SchemaDef> deserializer = new Deserializer<>(SchemaDef.BOND_TYPE);
+
+        final SchemaDef schemaDef = deserializer.deserialize(reader);
+        final RuntimeSchema schema = new RuntimeSchema(schemaDef);
```
sigansky: I thought we're getting rid of RuntimeSchema. It seems to be unused, except in examples.

examples/java/core/untagged_protocols/src/main/java/org/bondlib/examples/UntaggedProtocols.java
```diff
@@ -0,0 +1,43 @@
...
+        final SimpleBinaryWriter writer = new SimpleBinaryWriter(output, 1);
+
+        final Serializer<Example1> serializer = new Serializer<>();
+        serializer.serialize(obj, writer);
+
+        final ByteArrayInputStream input = new ByteArrayInputStream(output.toByteArray());
+
+        final SimpleBinaryReader reader = new SimpleBinaryReader(input, 1);
+        final Deserializer<Example2> deserializer = new Deserializer<>(Example2.BOND_TYPE);
+        final Example2 obj2 = deserializer.deserialize(reader, schema);
```
sigansky: >deserializer.deserialize(reader, schema) [](start = 30, length = 40)

Deserializer doesn't have a method that takes RuntimeSchema. If this compiles (I assume you made sure it does) then there must be some stale build problems. Please clean/rebuild everything. You only need to pass SchemaDef and there's no need for RuntimeSchema.

sigansky:
>         return value;

Missing implementation for Bonded. Need to deserialize (unmarshall) Bonded contents. Do you plan to do this in the next commit?

---
Refers to: java/core/src/main/java/org/bondlib/BondedBondType.java:137 in 95a2ac8. [](commit_id = 95a2ac859d3cbd643866228434759781c810deaa, deletion_comment = False)

java/core/src/main/java/org/bondlib/BlobBondType.java
```diff
@@ -107,7 +107,9 @@ protected final Blob deserializeValue(TaggedDeserializationContext context) thro
     }
 
     @Override
-    protected final Blob deserializeValue(UntaggedDeserializationContext context) throws IOException {
+    protected final Blob deserializeValue(
+        UntaggedDeserializationContext context,
+        TypeDef typeDef) throws IOException {
```
sigansky: >TypeDef typeDef [](start = 8, length = 15)

For Blob and all primitive types, need to check whether the given TypeDef is compatible with the current type and appropriate conversion or raise an error. Do you plan to add this?

java/core/src/main/java/org/bondlib/UInt32BondType.java
```diff
@@ -75,7 +75,8 @@ protected final Integer deserializeValue(TaggedDeserializationContext context) t
     }
 
     @Override
-    protected final Integer deserializeValue(UntaggedDeserializationContext context) throws IOException {
+    protected final Integer deserializeValue(
+        UntaggedDeserializationContext context, TypeDef typeDef) throws IOException {
```
sigansky: > TypeDef typeDef [](start = 47, length = 16)

Minor inconsistency: other types have this on a new line.

java/core/src/main/java/org/bondlib/protocol/UntaggedProtocolReader.java
```diff
@@ -193,6 +196,23 @@
     void skipWString() throws IOException;
 
     /**
+     * Skip based on supplied runtime schema, starting at the root of the SchemaDef
+     * 
+     * @param schema the schema def
+     * @throws IOException
+     */
+    void skip(SchemaDef schema) throws IOException;
```
sigansky: I prefer we don't have this method as it can trivially be implemented in terms of the more general method. Else, every implementation needs to implement an identical method (unless we use Java 8 with default implementation support).

java/core/src/main/java/org/bondlib/StructBondType.java
```diff
@@ -355,17 +357,24 @@ private void deserializeValueAsBase(TaggedDeserializationContext context, TStruc
...
-        this.deserializeValue(context, value);
+        this.deserializeValue(context, typeDef, value);
         return value;
     }
 
-    private void deserializeValue(UntaggedDeserializationContext context, TStruct value) throws IOException {
+    private void deserializeValue(
+            UntaggedDeserializationContext context,
+            TypeDef typeDef,
+            TStruct value) throws IOException {
```
sigansky: Indentation problems. Please auto-format in IDE.

java/core/src/main/java/org/bondlib/StructBondType.java
tstein: The current case uses the `StructDef`, but passing the `TypeDef` permits all cases (not just structs) to do error checking, right? That was the intent.

examples/java/core/runtime_schema/src/main/java/org/bondlib/examples/RuntimeSchemaExample.java
tstein: Yeah, I missed the examples. Correction incoming.

java/core/src/main/java/org/bondlib/BlobBondType.java
tstein: Not in this commit.

java/core/src/main/java/org/bondlib/UInt32BondType.java
tstein: Done.

java/core/src/main/java/org/bondlib/StructBondType.java
tstein: Done.

java/core/src/main/java/org/bondlib/protocol/UntaggedProtocolReader.java
tstein: Done.

examples/java/core/runtime_schema/src/main/bond/schema.bond
```diff
@@ -0,0 +1,13 @@
...
+
+struct Struct
+{
+    0: string String = "this is a string";
+}
+
+[StructAttribute("Value of the attribute")]
+struct Example
+{
+    [FieldAttribute("Value of the attribute")]
```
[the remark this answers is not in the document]
tstein: Done.

tstein:
One of the examples is still broken, and I'm not sure what the correct behavior is. C# exception messages coming in a different fixup from a different machine.

cs/src/core/RuntimeSchema.cs
tstein: Done.

chadwalters:
Failing example was due to some futzing I was doing with it before handed it off. Sorry!

java/core/src/main/java/org/bondlib/StructBondType.java
sigansky: This method is deserializeStructFields, which expects a struct as only structs can have fields. There is no scenario where TypeDef would be useful here.

sigansky:
>         return value;

I didn't see a reply to the comment regarding Bonded yet. It's not clear to me how an untagged payload containing Bonded struct (marshalled) is deserialized. The current code disregards the schema. Can you please add a TODO at least?

---
In reply to: [328975137](https://github.com/Microsoft/bond/pull/608#issuecomment-328975137) [](ancestors = 328975137)

---
Refers to: java/core/src/main/java/org/bondlib/BondedBondType.java:137 in a402c6b. [](commit_id = a402c6bf28e7d2ce17af960dfb96701ae9e30227, deletion_comment = False)

java/core/src/main/java/org/bondlib/StructBondType.java
tstein: Done.

tstein:
Forgot to update golden files. Fixed.

tstein:
Made a mistake during the fixup. Fixed that.

PR #609, opened 2017-09-12
Minor fix (proposal)

kant:


chadwalters:
@kant: I think the current syntax is intentional as a doxygen-related command to create a title. These .md files are not intended for direct consumption as mark down but rather are pumped through doxygen to produce the manual pages, such as https://microsoft.github.io/bond/why_bond.html and https://microsoft.github.io/bond/manual/bond_cpp.html . Does that make sense?

chadwalters:
I am closing this for now.

PR #611, opened 2017-09-13
[Java] Replaced field/struct metadata interfaces with SchemaDef's metadata

sigansky:
Changes:

1. Removed FieldMetadata and StructMetadata interfaces and replaced them with the Metadata Bond struct in protocol writers and BondType hierarchy.
2. Added SchemaHelpers class to abstract SchemaDef's intricacies, specifically default field values within a Variant struct.
3. Simplified SimpleBinaryProtocol's logic for getting field default values from schema.
4. Added SchemaDef property to StructBondType with caching (unable to do it at initialization time since that causes a circularity when initializing SchemaDef's own BondType). 
5. Fixed 2 bugs that are currently in other PR's: FieldDef was not added to field collection and initializing Variant with nothing led to null dereference.

### DIFF
--- a/java/core/src/main/java/org/bondlib/protocol/SchemaHelper.java
+++ b/java/core/src/main/java/org/bondlib/protocol/SchemaHelper.java
@@ -1,0 +1,66 @@
+package org.bondlib.protocol;
+
+import org.bondlib.Metadata;
+import org.bondlib.Modifier;
+
+/**
+ * Contains helper methods for working with schema.
+ */
+public final class SchemaHelper {
+
+    private SchemaHelper() {
+    }
+
+    public static byte getDefaultUInt8FieldValue(Metadata fieldMetadata) {
+        return (byte) fieldMetadata.default_value.uint_value;
+    }
+
+    public static short getDefaultUInt16FieldValue(Metadata fieldMetadata) {
+        return (short) fieldMetadata.default_value.uint_value;
+    }
+
+    public static int getDefaultUInt32FieldValue(Metadata fieldMetadata) {
+        return (int) fieldMetadata.default_value.uint_value;
+    }
+
+    public static long getDefaultUInt64FieldValue(Metadata fieldMetadata) {
+        return fieldMetadata.default_value.uint_value;
+    }
+
+    public static byte getDefaultInt8FieldValue(Metadata fieldMetadata) {
+        return (byte) fieldMetadata.default_value.int_value;
+    }
+
+    public static short getDefaultInt16FieldValue(Metadata fieldMetadata) {
+        return (short) fieldMetadata.default_value.int_value;
+    }
+
+    public static int getDefaultInt32FieldValue(Metadata fieldMetadata) {
+        return (int) fieldMetadata.default_value.int_value;
+    }
+
+    public static long getDefaultInt64FieldValue(Metadata fieldMetadata) {
+        return fieldMetadata.default_value.int_value;
+    }
+
+    public static boolean getDefaultBoolFieldValue(Metadata fieldMetadata) {
+        // false is represented by 0; everything else is considered true
+        return fieldMetadata.default_value.int_value != 0;
+    }
+
+    public static float getDefaultFloatFieldValue(Metadata fieldMetadata) {
+        return (float) fieldMetadata.default_value.double_value;
+    }
+
+    public static double getDefaultDoubleFieldValue(Metadata fieldMetadata) {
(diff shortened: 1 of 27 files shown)


java/core/src/main/java/org/bondlib/protocol/SimpleBinaryWriter.java
```diff
@@ -48,72 +49,58 @@ public void writeVersion() throws IOException {
...
-        //
-        // Fields that are = nothing have no default, so it is an unrecoverable error to attempt to
-        // serialize such a field if it is actually set to nothing (e.g., a SomethingObject which is
-        // null).
-
-        if (metadata.isDefaultNothing()) {
-            throw new IllegalArgumentException("can't omit fields with default = nothing in SimpleBinary");
-        }
-
+        // for optional fields. Please note that serialization framework never calls writeFieldOmitted
```
tstein: Is never supposed to. IIRC, this was a guard against creating garbage data if we incorrectly call this function, which we can't statically prevent. Has something changed?

java/core/src/main/java/org/bondlib/protocol/SchemaHelper.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public static boolean isRequiredField(Metadata fieldMetadata) {
+        return fieldMetadata.modifier.value == Modifier.Values.Required;
+    }
+
+    public static boolean isRequiredOptionalField(Metadata fieldMetadata) {
+        return fieldMetadata.modifier.value == Modifier.Values.RequiredOptional;
+    }
+
+    public static byte getDefaultUInt8FieldValue(Metadata fieldMetadata) {
```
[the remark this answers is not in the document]
sigansky: Abstraction of the schema so that we don't have to deal with Variant type. The logic is not intuitive and heavily influenced by C++ (e.g. piggy back on int to get bool default, while string/wstring have separate values). 

---
In reply to: [138743899](https://github.com/Microsoft/bond/pull/611#discussion_r138743899) [](ancestors = 138743899)

java/core/src/main/java/org/bondlib/protocol/SimpleBinaryWriter.java
sigansky: If we have such guards, let's have them in all public methods as I can think of dozens places where unexpected things can happen (e.g. call structEnd without structBegin). This check was not necessary in my opinion as the code guarantees that it won't happen.

An assert could be more appropriate here. I am strictly against replacing would-be-asserts by such exceptions.

---
In reply to: [138762754](https://github.com/Microsoft/bond/pull/611#discussion_r138762754) [](ancestors = 138762754)

java/core/src/main/java/org/bondlib/protocol/SchemaHelper.java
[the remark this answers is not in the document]
tstein: [C# has this logic inlined](https://github.com/Microsoft/bond/blob/master/cs/src/core/protocols/SimpleBinary.cs#L137), and it's exactly as foreign to C# as it is to Java. I don't see a useful abstraction here - you still have to know which of these to call, it's still going to blow up if you call the wrong one.

java/core/src/main/java/org/bondlib/protocol/SimpleBinaryWriter.java
tstein: This is a larger discussion and not relevant to the actual topic of this PR. Could you please revert this change in this PR, and, if you really want to change it, create a new one? If I'm correctly remembering that @chwarr and I added this together, he may have an opinion on it, too.

java/core/src/main/java/org/bondlib/protocol/SchemaHelper.java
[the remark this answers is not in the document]
sigansky: Again, I am following best practices and not C#'s implementation. Knowing which one to call is straightforward while knowing how to get the default value for a field of type bool or int is not. I'd like to keep it as is.

java/core/src/main/java/org/bondlib/protocol/SimpleBinaryWriter.java
sigansky: I'll have that discussion with Christopher when he's back. Meanwhile, it bothers me to have code that tests condition that never occurs; again I prefer an assert there. I am reverting this change for now, pending further discussion.

PR #613, opened 2017-09-14
[C++] Additional security fixes in SimpleArray

chadwalters:
Also, extend checked_add tests.

### DIFF
--- a/cpp/inc/bond/protocol/detail/simple_array.h
+++ b/cpp/inc/bond/protocol/detail/simple_array.h
@@ -5,6 +5,7 @@
 
 #include <limits>
 #include <stdexcept>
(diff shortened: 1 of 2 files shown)


cpp/inc/bond/protocol/detail/simple_array.h
```diff
@@ -38,6 +38,10 @@ class SimpleArray
 
     T pop()
     {
+        if (_size == 0)
+        {
+            throw std::underflow_error("Can't pop empty array");
```
sigansky: >Can't pop empty array [](start = 40, length = 21)

Perhaps "Can't pop from an empty array"?

PR #616, opened 2017-09-15
cmake integration for Java

tstein:


tstein:
cmake now understands a `java` target, which will build the plugin and install it into the local maven repo, build core, and build compat. Those targets all have sensible names (`gradle-plugin`, `gradle-plugin-install`, `java-core`, `java-compat`) and sensibly depend on each other and on building `gbc`. This PR updates `.travis.yml` to use this target.

Immediate next steps:
* Builds still happen in-tree, in the old `build/` dirs. e.g., core will build in `$bond_git_root/java/core/build`. I wish gradle took a `--build-directory`, but it doesn't seem to be that simple.
* Examples aren't covered. It should be easy to adapt the `add_gradle_build` cmake function added in this PR to `gradle run` each example, and it should probably be structured similiarly to `java/core` - use `add_subdirectory` to recur into `examples/core/java` and then put a simple `CMakeLists.txt` in the dir for each example. I wouldn't have the `java` target depend on the examples, which it will if you use `add_gradle_build` without changing it.
* I didn't have a chance to test what happens if you aren't set up for/don't care about Java.

Possible next steps:
* As written, the cmake stuff in this change completely ignores source file tracking or changes and delegates those decisions to gradle. I think this is the right call, but you guys should consider it.
* Because gradle will still use various in-tree build dirs, `make clean` won't actually clean the Java build. If we can't figure out how to do an out-of-tree build in the shared build dir easily, might be worth making a java-clean and having clean depend on it or something.

cmake/FindJavaBuildTools.cmake
```diff
@@ -0,0 +1,26 @@
...
+endif()
+
+set (MAVEN_FOUND FALSE)
+if (MAVEN_EXECUTABLE)
+    set (MAVEN_FOUND TRUE)
+    execute_process (
+        COMMAND ${MAVEN_EXECUTABLE} --version
+        OUTPUT_VARIABLE maven_version)
+    message (STATUS "maven found at ${MAVEN_EXECUTABLE}")
+    message (STATUS "${maven_version}")
```
lalo: Will this output maven before the number version or just a random number?

tstein: This will spit out a whole block of text. I couldn't find a option that gives a useful single line.
```
-- maven found at /usr/local/bin/mvn
-- Apache Maven 3.5.0 (ff8f5e7444045639af65f6095c62210b5713f426; 2017-04-03T19:39:06Z)
Maven home: /usr/local/maven-3.5.0
Java version: 1.8.0_144, vendor: Oracle Corporation
Java home: /usr/lib/jvm/java-8-oracle/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "4.4.0-93-generic", arch: "amd64", family: "unix"
```

gradle is similar:
```
-- gradle found at /usr/local/bin/gradle
-- 
------------------------------------------------------------
Gradle 4.0.1
------------------------------------------------------------
Build time:   2017-07-07 14:02:41 UTC
Revision:     38e5dc0f772daecca1d2681885d3d85414eb6826
Groovy:       2.4.11
Ant:          Apache Ant(TM) version 1.9.6 compiled on June 29 2015
JVM:          1.8.0_144 (Oracle Corporation 25.144-b01)
OS:           Linux 4.4.0-93-generic amd64
```

tstein:
This should be good to go. Please rebase and merge.

tstein:
This should work now.

cmake/FindJavaBuildTools.cmake
lalo: ok this is fine

cmake/Java.cmake
```diff
@@ -0,0 +1,57 @@
+cmake_minimum_required (VERSION 2.8.12)
+
+#set (CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../cmake)
```
lalo: what is this comment? for debugging purpose or you forgot to remove?

PR #618, opened 2017-09-20
Attribute fixes for Bond-over-gRPC

chadwalters:
Add service attribute codegen tests for C#.
Fix service/method attributes in C# and C++.


compiler/Main.hs
```diff
@@ -87,7 +87,7 @@ cppCodegen options@Cpp {..} = do
     templates = concat $ map snd $ filter fst codegen_templates
     codegen_templates = [ (core_enabled, core_files)
                         , (comm_enabled, [comm_h export_attribute, comm_cpp])
-                        , (grpc_enabled, [grpc_h export_attribute])
+                        , (grpc_enabled, [grpc_h export_attribute, grpc_cpp])
```
lalo: We need to update the cmake to include the generated *grpc.cpp files, in this line: https://github.com/Microsoft/bond/blob/master/cmake/Bond.cmake#L60

chwarr: We also need to compile them. Later on in the same file, in `add_bond_executable`, something similar to what's done for Comm is needed now.

compiler/src/Language/Bond/Codegen/Cpp/Grpc_cpp.hs
```diff
@@ -0,0 +1,34 @@
...
+import Language.Bond.Codegen.Util
+import qualified Language.Bond.Codegen.Cpp.Util as CPP
+
+-- | Codegen template for generating /base_name/_grpc.cpp containing
+-- definitions of helper functions and schema metadata static variables.
+grpc_cpp :: MappingContext -> String -> [Import] -> [Declaration] -> (String, Text)
+grpc_cpp cpp file _imports declarations = ("_grpc.cpp", [lt|
+#include "#{file}_reflection.h"
+#include "#{file}_grpc.h"
+#include <bond/core/exception.h>
```
chwarr: Do we directly generate code that throws Bond exceptions? I think we can omit this header.

compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
```diff
@@ -141,10 +144,68 @@ inline #{className}::#{proxyName}<TThreadPool>::#{proxyName}(
...
+
+        pushMethod (method, i) =
+            [lt|private: typedef #{typename}boost::mpl::push_front<methods#{i}, #{typename}service::#{method}>::type methods#{i + 1};|]
+
+        -- constructor, generated only for service templates
+        constructor = onlyTemplate [lt|Schema()
+        {
+            // Force instantiation of template statics
+            (void)metadata;
+            #{newlineSep 4 static serviceMethods}
```
chwarr: Minor: The indentation is off on these. I think we want an indentation level of 3, not 4. (See the golden files.)

cpp/inc/bond/core/reflection.h
```diff
@@ -200,6 +200,37 @@ const typename method_pointer<Service, Input, Result>::type
...
+    static const Metadata& metadata;
+};
+
+
+template <
+    typename Service,
+    typename Input,
+    typename Result,
+    const bond::Metadata* metadata_ptr>
+    const bond::Metadata&
```
chwarr: Minor: indentation not consistent with `MethodTemplate`

cpp/inc/bond/core/reflection.h
```diff
@@ -200,6 +200,37 @@ const typename method_pointer<Service, Input, Result>::type
     MethodTemplate<Service, Input, Result, method_ptr, metadata_ptr>::method = method_ptr;
 
 
+/// @brief Method description in compile-time schema
+template <
+    typename Service,
+    typename Input,
+    typename Result,
+    const Metadata* metadata_ptr>
+    struct MethodTemplateNoFunc
```
chwarr: The name here isn't great. That appears to be because we're avoiding breaking back compat, by renaming `MethodTemplate`; however, only Comm uses `MethodTemplate`, so if we want to swap names, we can.

Or, possibly call this `MethodTemplateBase` and have `MethodTemplate` inherit from it?

compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
```diff
@@ -141,10 +144,68 @@ inline #{className}::#{proxyName}<TThreadPool>::#{proxyName}(
...
+};
+#{onlyTemplate $ CPP.schemaMetadata cpp s}
 |]
       where
         className = CPP.className s
         template = CPP.template s
+        onlyTemplate x = if null declParams then mempty else x
+        typename = onlyTemplate [lt|typename |]
+
+        export_attr = optional (\a -> [lt|#{a}
```
chwarr: We should update the DLL example to access metadata to make sure this works as we expect.

compiler/Main.hs
chadwalters: Got it.

compiler/src/Language/Bond/Codegen/Cpp/Grpc_cpp.hs
chadwalters: It's a holdover from Comm_cpp.hs, where it also may not be needed. I was erring on the side of caution but I'll look to remove it.

cpp/inc/bond/core/reflection.h
chadwalters: I think I'd rather pull this out into a separate bond::ext::grpc::reflection.h header. Sound OK?

compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
chadwalters: Will do.

cpp/inc/bond/core/reflection.h
chadwalters: Fixed.

cpp/inc/bond/core/reflection.h
chwarr: I like it even better.

PR #620, opened 2017-09-22
[Java] Transcoder and TextProtocolReader API

sigansky:


java/core/src/main/java/org/bondlib/TextProtocolReader.java
```diff
@@ -0,0 +1,164 @@
...
+
+/**
+ * Reads from serialized payload encoded using a structured text protocol such as JSON or XML.
+ */
+public interface TextProtocolReader {
+
+    /**
+     * Start reading a struct.
+     */
+    void readStructBegin() throws IOException;
```
sigansky: >void [](start = 4, length = 4)

Return struct name if supported by the protocol.

java/core/src/main/java/org/bondlib/Transcoder.java
```diff
@@ -0,0 +1,68 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package org.bondlib;
+
+import java.io.IOException;
+
+/**
+ * Responsible for transcoding between protocols.
+ */
+public final class Transcoder {
```
sigansky: >Transcoder [](start = 19, length = 10)

Make a static class.

PR #621, opened 2017-09-23
[c++ grpc] Drop client callback after executing

chwarr:
After executing the client callback upon receiving a response, clear the
stored callback to prevent a shared_ptr cycle.

Previously, when the callback function was not cleared, if the callback
happened to store the unary_call_result that it was invoked with (like
wait_callback does), there would be a shared_ptr cycle: the callback
would have a shared_ptr to the unary_call_result, which was keeping its
parent client_unary_call_data alive. This client_unary_call_data had a
copy of the callback function, which had a shared_ptr to the
unary_call_result.

This change resets the callback to empty after is has been executed,
breaking the cycle.

### DIFF
--- a/cpp/inc/bond/ext/grpc/detail/client_call_data.h
+++ b/cpp/inc/bond/ext/grpc/detail/client_call_data.h
@@ -51,14 +51,14 @@ struct client_unary_call_data
     std::shared_ptr<TThreadPool> _threadPool;
     /// A response reader.
     std::unique_ptr<grpc::ClientAsyncResponseReader<bond::bonded<TResponse>>> _responseReader;
-    /// The arguments to pass back into the client callback. Also doubles as 
+    /// The arguments to pass back into the client callback. Also doubles as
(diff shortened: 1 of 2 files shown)


chwarr:
This change appears to have introduced some crashes. Investigating.

cpp/inc/bond/ext/grpc/detail/client_call_data.h
```diff
@@ -71,8 +71,8 @@ struct client_unary_call_data
           _threadPool(std::move(threadPool)),
           _responseReader(),
           _callbackArgs(context),
-          _self(),
-          _cb(cb)
+          _cb(cb),
```
lalo: does the order matter here?

chwarr: Members are initialized in the order they are declared in. I like to keep the initializer list in the same order. It reduces cognitive burden, and some compilers warn if the orders differ.

PR #629, opened 2017-09-29
[c#] Use null XmlResolver when possible

chwarr:
By setting XmlReaderSettings.XmlResolver to null, we can avoid having
XmlReader resolve external XML entities. Attempting to resolve these
external entities can potentially access the network, which is not
something we want to do during deserialization.

This API only exists when targeting the full .NET 4.5 framework, hence
the need for a net45-nonportable version of Bond.dll.

### DIFF
--- a/cs/src/core/protocols/SimpleXmlReader.cs
+++ b/cs/src/core/protocols/SimpleXmlReader.cs
@@ -26,22 +26,22 @@ namespace Bond.Protocols
                     IgnoreComments = true,
                     IgnoreProcessingInstructions = true,
 #if SUPPORTS_XMLRESOLVER
-                // do not attempt to resolve any external resources
-                XmlResolver = null,
+                    // do not attempt to resolve any external resources
+                    XmlResolver = null,
 #endif
                 }))
         { }
 
         public SimpleXmlReader(TextReader textReader)
-    : this(XmlReader.Create(textReader, new XmlReaderSettings
-    {
-        IgnoreComments = true,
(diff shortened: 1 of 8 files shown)


cs/src/core/protocols/SimpleXmlReader.cs
```diff
@@ -21,13 +21,29 @@ public SimpleXmlReader(XmlReader reader)
...
 
         public SimpleXmlReader(Stream stream)
-            : this(XmlReader.Create(stream, new XmlReaderSettings 
+            : this(XmlReader.Create(stream, new XmlReaderSettings
                 {
                     IgnoreComments = true,
-                    IgnoreProcessingInstructions = true
+                    IgnoreProcessingInstructions = true,
+#if SUPPORTS_XMLRESOLVER
+                // do not attempt to resolve any external resources
```
chadwalters: Minor: Inconsistent indentation?

cs/src/core/protocols/SimpleXmlReader.cs
```diff
@@ -21,13 +21,29 @@ public SimpleXmlReader(XmlReader reader)
...
+#endif
                 }))
         { }
 
+        public SimpleXmlReader(TextReader textReader)
+    : this(XmlReader.Create(textReader, new XmlReaderSettings
+    {
+        IgnoreComments = true,
+        IgnoreProcessingInstructions = true,
+#if SUPPORTS_XMLRESOLVER
```
chadwalters: Minor: Inconsistent indentation?

PR #630, opened 2017-10-05
[C# grpc] Include grpc generated files automatically.

hcoona:
This change would include "%(filename)_grpc.cs" as well as "%(filename)_types.cs" only for the `BondCodegen` items with addition `Options` meta and the meta contains '--grpc'.


.travis.yml
```diff
@@ -72,8 +72,10 @@ script:
     - if [ "$FLAVOR" == "cs" ]; then make DESTDIR=$HOME install; fi
     - if [ "$FLAVOR" == "cs" ]; then cd ..; fi
     - if [ "$FLAVOR" == "cs" ]; then export BOND_COMPILER_PATH=$HOME/usr/local/bin; fi
-    - if [ "$FLAVOR" == "cs" ]; then xbuild /p:Configuration=Debug cs/cs.sln; fi
-    - if [ "$FLAVOR" == "cs" ]; then xbuild /p:Configuration=Fields cs/cs.sln; fi
+    # exclude the Compiler.vcxproj from cs/cs.sln on Linux (xbuild ignore it with a warning but msbuild fail on it)
+    - if [ "$FLAVOR" == "cs" ]; then sed '/21E175D5-BBDD-4B63-8FB7-38899BF2F9D1/d' cs/cs.sln > cs/cs_no_cpp.sln; fi
+    - if [ "$FLAVOR" == "cs" ]; then msbuild /p:Configuration=Debug cs/cs_no_cpp.sln; fi
```
chwarr: Switching to msbuild is understandable, but shouldn't be part of this PR. Can you revert the xbuild to msbuild changes?

cs/build/nuget/Common.targets
```diff
@@ -127,6 +127,12 @@
         <DependentUpon>%(BondCodegen.Identity)</DependentUpon>
       </_BondGeneratedFiles>
 
+      <_BondGeneratedFiles Include="@(BondCodegen -> '$(BondOutputDirectory)\%(FileName)_grpc.cs')"
+                           Condition="$([System.String]::new('%(BondCodegen.Options)').Contains('--grpc'))">
```
chwarr: This also needs to handle the case when the property `$(BondOptions)` contains "--grpc". Note that `$(BondOptions)` is only used for [`BondCodegen` items that don't have overridden Options metadata](https://github.com/Microsoft/bond/blob/ca0d1c63da2760ad5753a6a8bf29b00f8cb01585/cs/build/nuget/Common.targets#L103-L104). One of the examples can be changed to use `$(BondOptions)` as a test for this behavior.

chwarr:
Go ahead and add new commits (instead of rebasing). That will make follow ups to the PR easier. One of the maintainers can squash/rebase for the final merge.

.travis.yml
hcoona: xbuild didn't get the [property function](https://docs.microsoft.com/en-us/visualstudio/msbuild/property-functions) work fine, so I have to switch to msbuild here. I'll make another PR for the msbuild switching.

hcoona: @chwarr Please see PR #633

chwarr:
Since this relies on MSBuild functionality, it would, when released, break anyone using xbuild. Let me see whether I can assess how large that population is. (I suspect that such a breaking change would be OK.)

chwarr:
Requiring MSBuild is fine, so we can rely on MSBuild-only functionality. After https://github.com/Microsoft/bond/pull/633 is finished, this PR can be resumed atop those changes.

chwarr:
#633 has been merged (commit 85995fe), so this can be reworked atop that. You may want to keep an eye on https://github.com/Microsoft/bond/pull/649, which is also touching .travis.yml, but doesn't materially affect the C# build.

chwarr:
I added support for `--structs=false` in PR https://github.com/Microsoft/bond/pull/653, which includes the changes in this one as well. This PR will get closed automatically when #653 is merged: please leave it open until then.

PR #631, opened 2017-10-05
Update missing package dependency for gRPC build in README

ara-ayvazyan:


README.md
```diff
@@ -86,6 +86,7 @@ different options:
 
 ```bash
 sudo apt-get install \
+    golang \
```
chwarr: Minor: this list is already alpha sorted. Let's keep it that way.

PR #635, opened 2017-10-06
Ignore .vs folder

ara-ayvazyan:


.gitignore
```diff
@@ -12,5 +12,6 @@
 *.swo
 *.swm
 
-# VS Code
-/.vscode/
+# VS
+/.vs
```
chwarr: Let's just ignore a `.vs` directory anywhere in the tree, so can we change to `.vs/` here and remove the one we have in `cs\.gitignore`.

Similarly for the `.vscode` directory: `.vscode/`

PR #639, opened 2017-10-10
[c++] Avoid compiling unused default ctor

ara-ayvazyan:


compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
```diff
@@ -206,10 +206,25 @@ namespace std
...
 
+        needAlloc alloc = isJust structBase || any (allocParameterized alloc . fieldType) structFields
+        allocParameterized alloc t = (isStruct t) || (L.isInfixOf (L.pack alloc) . toLazyText $ cppType t)
+
         -- default constructor
         defaultCtor = [lt|
-        #{declName}()#{initList}#{ctorBody}|]
+        #{dummyTemplateTag}#{declName}(#{vc12WorkaroundParam})#{initList}#{ctorBody}|]
           where
+            needAllocParam = case allocator of
```
chwarr: Consider `needAllocParam = maybe False needAlloc`

Issue #590, opened 2017-09-04
nullable<alias> does generate nullable type for decimal


DanielFabian:
I have a schema, where I define decimal as in the example and it correctly generates the `decimal` in C#. But if the property is a `nullable<decimal>` it does not work for some reason. The same works just fine for `DateTime`. This schema:

    namespace Repro
    
    using decimal = blob;
    
    struct ReproStruct
    {    
        0: decimal MyProp;
        1: nullable<decimal> MyBuggyProp;
    }

generates this faulty C#:

    namespace Repro
    {
        using System.Collections.Generic;

        [global::Bond.Schema]
        [System.CodeDom.Compiler.GeneratedCode("gbc", "0.10.0.0")]
        public partial class ReproStruct
        {
            [global::Bond.Id(0), global::Bond.Type(typeof(global::Bond.Tag.blob))]
            public decimal MyProp { get; set; }

            [global::Bond.Id(1), global::Bond.Type(typeof(global::Bond.Tag.nullable<global::Bond.Tag.blob>))]
            public decimal MyBuggyProp { get; set; }

            public ReproStruct()
                : this("Repro.ReproStruct", "ReproStruct")
            {}

            protected ReproStruct(string fullName, string name)
            {
                MyProp = new decimal();
            }
        }
    } // Repro

chwarr:
Yup, I see the problem. Thanks for the report.

When [determining the C# type to emit for a nullable field](https://github.com/Microsoft/bond/blob/fcd492dfd7870774e604130ab9f2d8dc1070da8b/compiler/src/Language/Bond/Codegen/TypeMapping.hs#L422), either `type` (for a C# reference type) or `type?` (for a C# value type), gbc uses the [`isScalar`](https://github.com/Microsoft/bond/blob/939c298ff6d619dbc0bc9f5ca25c2088de86d83b/compiler/src/Language/Bond/Syntax/Util.hs#L49-L65) function, and appends the `?` for scalar types. (For aliases, `isScalar` returns the scalarness of the underlying Bond type.) `isScalar` operates in the Bond type system, independent of the target language. This works out when `isScalar` and "Is a C# value type?" are the same, which is the case when no type mapping is going on.

[...]

PR #642, opened 2017-10-12
Remove extra blob conversions during serialization

chadwalters:
There were extra conversions of aliased blobs during serialization. Now each blob is only converted once.

Also, updated the expressions tests to include aliased blobs.

### DIFF
--- a/cs/src/core/expressions/ObjectParser.cs
+++ b/cs/src/core/expressions/ObjectParser.cs
@@ -100,7 +100,27 @@ namespace Bond.Expressions
             var fieldId = Expression.Constant(id);
             var fieldType = Expression.Constant(fieldSchemaType.GetBondDataType());
             var fieldValue = DataExpression.PropertyOrField(structVar, schemaField.Name);
-            var parser = new ObjectParser(this, fieldValue, fieldSchemaType);
+
+            ObjectParser parser = null;
+
+            Expression blob = null;
+            ParameterExpression convertedBlob = null;
+
+            // To avoid calling Convert multiple times on the same aliased Blob
+            // we must construct a new ObjectParser with the expected return type of
+            // of Convert
+            if (fieldSchemaType.IsBondBlob())
+            {
+                blob = typeAlias.Convert(fieldValue, fieldSchemaType);
+                convertedBlob = Expression.Variable(blob.Type, "convertedBlob");
+
+                if (blob.Type != fieldValue.Type)
+                {
+                    parser = new ObjectParser(this, convertedBlob, convertedBlob.Type);
+                }
+            }
+
+            parser = parser ?? new ObjectParser(this, fieldValue, fieldSchemaType);
 
             var processField = field != null
                 ? field.Value(parser, fieldType)
@@ -121,9 +141,14 @@ namespace Bond.Expressions
 
                 if (fieldSchemaType.IsBondBlob())
                 {
-                    cannotOmit = Expression.NotEqual(
-                        typeAlias.Convert(fieldValue, fieldSchemaType), 
+                    var notEqual = Expression.NotEqual(
+                        convertedBlob,
                         Expression.Default(typeof(ArraySegment<byte>)));
+
+                    return Expression.Block(
+                        new[] { convertedBlob },
+                        Expression.Assign(convertedBlob, blob),
+                        PrunedExpression.IfThenElse(notEqual,processField,omitField));
                 }
                 else if (fieldSchemaType.IsBondContainer())
                 {
@@ -154,7 +179,8 @@ namespace Bond.Expressions
                 new ObjectParser(this, item, itemType),
                 Expression.Constant(itemType.GetBondDataType()),
                 next,
-                count);
+                count,
+                null);
 
             if (value.Type.IsArray)
                 return ArrayContainer(itemHandler);
@@ -332,7 +358,8 @@ namespace Bond.Expressions
                 new ObjectParser(this, value, valueType),
                 Expression.Constant(valueType.GetBondDataType()),
                 Expression.NotEqual(Expression.PostDecrementAssign(count), Expression.Constant(0)),
-                count);
+                count,
+                null);
 
             return Expression.Block(
                 new[] { count },
@@ -355,7 +382,8 @@ namespace Bond.Expressions
                 new ObjectParser(this, item, typeof(sbyte)),
                 Expression.Constant(BondDataType.BT_INT8),
                 Expression.LessThan(index, end),
-                count);
+                count,
+                arraySegment);
 
             return Expression.Block(
                 new[] { arraySegment, count, index, end },
--- a/cs/src/core/expressions/ObjectParser.cs
+++ b/cs/src/core/expressions/ObjectParser.cs
@@ -148,7 +148,7 @@ namespace Bond.Expressions
                     return Expression.Block(
                         new[] { convertedBlob },
                         Expression.Assign(convertedBlob, blob),
-                        PrunedExpression.IfThenElse(notEqual,processField,omitField));
+                        PrunedExpression.IfThenElse(notEqual, processField, omitField));
                 }
                 else if (fieldSchemaType.IsBondContainer())
                 {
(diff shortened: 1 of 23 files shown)


cs/test/core/ConvertTests.cs
```diff
@@ -0,0 +1,112 @@
...
+            ConvertToArraySegmentCount = 0;
+        }
+    }
+
+
+    [TestFixture]
+    public class TypeAliasTests
+    {
+        [SetUp]
+        public void Setuo()
```
lalo: Setup? (typo right?)

cs/test/core/ConvertTests.cs
```diff
@@ -0,0 +1,112 @@
...
+        public static void Clear()
+        {
+            ConvertToDecimalCount = 0;
+            ConvertToArraySegmentCount = 0;
+        }
+    }
+
+
+    [TestFixture]
+    public class TypeAliasTests
```
lalo: ConvertTests instead of TypeAliasTests to match file name

cs/test/core/ConvertTests.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            var outputStream = new OutputBuffer();
+            var writer = new CompactBinaryWriter<OutputBuffer>(outputStream);
+            var serializer = new Serializer<CompactBinaryWriter<OutputBuffer>>(typeof(Decimals));
+
+            serializer.Serialize(foo, writer);
+
+            Assert.AreEqual(1+foo.Numbers.Count, BondTypeAliasConverter.ConvertToArraySegmentCount);
+            Assert.AreEqual(0, BondTypeAliasConverter.ConvertToDecimalCount);
+        }
```
lalo: We could also add tests that count how many converts happen on **deserialize**, just to make sure we never regress.

cs/test/expressions/Program.cs
```diff
@@ -108,6 +108,29 @@ public SerializerDebugView(Type type)
         string IDebugView.DebugView { get { return debugView; } }
     }
 
+    public static class BondTypeAliasConverter
+    {
+        public static int ConvertToDecimalCount = 0;
```
lalo: do we need this counters on test/expressions/Program? Don't see it being used.

cs/test/core/ConvertTests.cs
chadwalters: Yes, I plan to expand test coverage to cover maps, vectors, etc., and also to add deserialization tests.

cs/test/expressions/Program.cs
chadwalters: Removed.

cs/test/core/ConvertTests.cs
chadwalters: Done.

cs/test/core/ConvertTests.cs
chadwalters: Done.

cs/test/core/ConvertTests.cs
```diff
@@ -0,0 +1,112 @@
...
+            var serializer = new Serializer<CompactBinaryWriter<OutputBuffer>>(typeof(Decimals));
+
+            serializer.Serialize(foo, writer);
+
+            Assert.AreEqual(1, BondTypeAliasConverter.ConvertToArraySegmentCount);
+            Assert.AreEqual(0, BondTypeAliasConverter.ConvertToDecimalCount);
+        }
+
+        [Test]
+        // this test currently fails
```
lalo: outdated comment

chadwalters: Addressed.

chadwalters:
Updated with expanded test coverage. There is still an issue with nullable<blob> that might need to be handled in a separate request.

cs/src/core/expressions/ObjectParser.cs
```diff
@@ -121,9 +141,14 @@ Expression Field(ITransform transform, Expression structVar, UInt16 id, ISchemaF
...
-                    cannotOmit = Expression.NotEqual(
-                        typeAlias.Convert(fieldValue, fieldSchemaType), 
+                    var notEqual = Expression.NotEqual(
+                        convertedBlob,
                         Expression.Default(typeof(ArraySegment<byte>)));
+
+                    return Expression.Block(
+                        new[] { convertedBlob },
+                        Expression.Assign(convertedBlob, blob),
+                        PrunedExpression.IfThenElse(notEqual,processField,omitField));
```
chwarr: Minor: spaces around commas

cs/test/core/Core.csproj
```diff
@@ -54,6 +54,7 @@
     <Compile Include="SerializerGeneratorFactoryTests.cs" />
     <Compile Include="StreamTests.cs" />
     <Compile Include="Structs.cs" />
+    <Compile Include="ConvertTests.cs" />
```
chwarr: Minor: this list is alpha sorted (or vaguely alpha sorted), so can we move ConvertTests up?

cs/test/expressions/SerializeCB.expressions
```diff
@@ -327,26 +327,221 @@
...
+                    .Call $writer.WriteFieldBegin(
+                        .Constant<Bond.BondDataType>(BT_LIST),
+                        .Constant<System.UInt16>(50),
+                        .Constant<Bond.Metadata>(b));
+                    .Block(
+                        System.ArraySegment`1[System.Byte] $arraySegment,
+                        System.Int32 $count,
+                        System.Int32 $index,
+                        System.Int32 $end) {
+                        $arraySegment = $Example.b;
```
chwarr: Is there any simplification that can be done to get rid of this older optimization for a similar problem?

cs/test/expressions/schemas.bond
```diff
@@ -12,6 +12,8 @@ struct Nested
     0: double _double;
 }
 
+using decimal = int32;
```
chwarr: A decimal can't always fit in an int32. I would have expected a blob here... Am I missing something?

cs/test/core/ConvertTests.cs
```diff
@@ -0,0 +1,150 @@
...
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.Linq;
+    using System.Reflection;
+    using System.Threading;
+    using Bond;
+    using NUnit.Framework;
```
chwarr: Minor: put the Bond usings near each other.

cs/test/core/ConvertTests.cs
```diff
@@ -0,0 +1,150 @@
...
+    using Bond.IO.Safe;
+    using Bond.Protocols;
+
+    [global::Bond.Schema]
+    public partial class Decimals
+    {
+        [global::Bond.Id(0), global::Bond.Type(typeof(global::Bond.Tag.blob))]
+        public System.Decimal _dec { get; set; }
```
chwarr: Is `System.Decimal` instead of `decimal` intentional here?

cs/test/core/ConvertTests.cs
```diff
@@ -0,0 +1,150 @@
...
+    using System.Linq;
+    using System.Reflection;
+    using System.Threading;
+    using Bond;
+    using NUnit.Framework;
+    using Bond.IO.Safe;
+    using Bond.Protocols;
+
+    [global::Bond.Schema]
+    public partial class Decimals
```
chwarr: If we're not merging with generated code, can we drop the `partial`?

cs/test/core/ConvertTests.cs
```diff
@@ -0,0 +1,150 @@
...
+            var bits = decimal.GetBits(value);
+            var data = new byte[bits.Length * sizeof(int)];
+            Buffer.BlockCopy(bits, 0, data, 0, data.Length);
+
+            Interlocked.Increment(ref ConvertToArraySegmentCount);
+
+            return new ArraySegment<byte>(data);
+        }
+
+        public static void Clear()
```
chwarr: Consider `ResetCounts()`.

cs/src/core/expressions/ObjectParser.cs
chadwalters: Done.

cs/test/core/Core.csproj
chadwalters: Done.

cs/test/expressions/schemas.bond
chadwalters: Nope, just a mistake on my part. This was something I was playing with to compare behavior when aliasing a primitive vs a blob and I didn't mean to include it in the PR. Reverted.

cs/test/core/ConvertTests.cs
chadwalters: Done.

cs/test/core/ConvertTests.cs
chadwalters: Done.

cs/test/core/ConvertTests.cs
chadwalters: Probably not. Corrected.

cs/test/core/ConvertTests.cs
chadwalters: Done.

cs/test/expressions/SerializeCB.expressions
chadwalters: Investigating this.

chadwalters:
Pushed a new version with CR comments addressed.

Nullable has also been addressed although this does touch on some of the issues raised in https://github.com/Microsoft/bond/issues/590.

chadwalters:
N.B.: When this is ready for merge, please squash and edit commit message (especially portion related to nullable<blob>).

cs/test/expressions/SerializeCB.expressions
chadwalters: Removing this adds quite a bit of complexity between the code paths where blobs require conversion and where they don't. Not sure it is worth addressing this unless expression optimizer isn't capable of the appropriate peephole optimizations. Thoughts?

chwarr: Thanks for looking. If it's not super simple, I'd skip the removal.

chadwalters:
Updated with tests for mapping to reference types as well as value types. Please squash and edit commit message (nullables are now handled) when ready.

PR #644, opened 2017-10-16
[c++] Surround macro parameters with parenthesis

chwarr:
Macro parameters with commas in them, like "std::map<int, std::string>"
need to be surrounded with parenthesis so they are treated as one
parameter instead of as two parameters to the macro.

Standardize where the ! is in parthensized parameters.

### DIFF
--- a/cpp/inc/bond/core/traits.h
+++ b/cpp/inc/bond/core/traits.h
@@ -23,6 +23,8 @@
 #include <memory>
 #endif
 
+#include <utility>
+
 namespace bond
 {
 
--- a/cpp/inc/bond/core/traits.h
+++ b/cpp/inc/bond/core/traits.h
@@ -4,8 +4,9 @@
 #pragma once
 
 #include "config.h"
-#include "scalar_interface.h"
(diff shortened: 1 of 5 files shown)


cpp/inc/bond/core/traits.h
```diff
@@ -193,7 +193,7 @@ is_reader<Input, T, typename boost::enable_if<is_class<typename Input::Parser> >
 template <typename T> struct
 buffer_magic
 {
-    BOOST_STATIC_ASSERT_MSG(!is_same<T, T>::value, "Undefined buffer.");
+    BOOST_STATIC_ASSERT_MSG((!is_same<T, T>::value), "Undefined buffer.");
```
ara-ayvazyan: Maybe we should add `template <typename T> struct always_false : std::false_type {};` and use as `always_false<T>::value` which will not have `,` in the macro and will be cleaner?

chwarr: Addressed

chwarr:
Please squash when merging. Thanks.

PR #645, opened 2017-10-17
[c++ grpc] Add shared_unary_call and example

chwarr:
<strike>Please squash when merging, and take the commit message primarily from 06e9fe41.</strike>
Please squash when merging.


cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -149,37 +164,64 @@ class unary_call final
...
+        return std::make_shared<unary_call>(std::move(*this));
+    }
+
     /// @brief Responds to the client with the given message and status.
     ///
     /// Only the first call to \p Finish or \p FinishWithError will be
     /// honored.
+    ///
+    /// @note May only be called on unary_call objects that haven't been
+    /// moved-from.
```
lalo: I don't understand this comment. My understanding is that if you std::move the object to another call, like in your example (line 67), you later call Finish (line 53). The sentence has the opposite meaning for me.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -100,28 +101,35 @@ namespace detail {
...
 /// @brief The details of a single async, unary call.
 ///
 /// Call \p Finish or \p FinishWithError to send a response back to the
 /// client.
 ///
 /// If no explicit call to Finish/FinishWithError has been made before this
-/// unary_call instance is destroyed, a generic internal server error is
-/// sent.
+/// unary_call instance is destroyed or moved from, a generic internal
+/// server error is sent.
```
lalo: Should we add a comment on the codegen on the server stub similar to this one where we define the expected behavior of the life of the unary_call that is received?

cpp/inc/bond/ext/grpc/unary_call.h
chwarr: Is this clearer phrasing? "May not be called on this instance if it has been moved-from."

The two unary_call instances on lines 53 and 67 are different instances. When the instance from 67 is used to create the one that used on line 53, the one on 67 enters the "moved-from" state.

cpp/inc/bond/ext/grpc/unary_call.h
chwarr: If we start lifting documentation to the generated code, we're going to get a lot of duplication, so I'd rather not say anything, hopefully causing a search for (and a finding of) unary_call's documentation.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -149,37 +159,58 @@ class unary_call final
...
         return _impl->_request;
     }
 
+    /// @brief Creates a std::shared_ptr<unary_call> from this instance.
+    ///
+    /// This instance may not be used after to_shared has been called; the
+    /// returned shared instance must be used instead.
+    ///
+    /// @note May not be called on this instance if it has been moved-from.
+    std::shared_ptr<unary_call> to_shared()
```
ara-ayvazyan: Rename to `share` and mark as `&&`?

examples/cpp/grpc/async-server/async-server.cpp
```diff
@@ -0,0 +1,123 @@
...
+private:
+    using CallType = bond::ext::gRPC::unary_call<bond::bonded<HelloRequest>, HelloReply>;
+
+    std::shared_ptr<bond::ext::gRPC::thread_pool> _tp;
+
+    class PerRequestState
+    {
+    public:
+        explicit PerRequestState(CallType call)
+            : _call(call.to_shared())
```
ara-ayvazyan: This will become `std::move(call).share()`.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -118,18 +118,13 @@ class unary_call final
...
         : _impl(std::move(other._impl))
     {
         other._impl.reset();
     }
 
-    /// @brief Assigned to this unary_call by moving from another
-    /// unary_call.
-    ///
-    /// The moved-from unary_call will be empty.
+    /// @brief Assigne to this unary_call by moving from another unary_call.
```
lalo: is this assignee or a typo?

chwarr: Typo. Fixing it.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -149,37 +159,58 @@ class unary_call final
     }
 
     /// @brief Get the server context for this call.
+    ///
+    /// @note May not be called on this instance if it has been moved-from.
```
ara-ayvazyan: Put a single comment about the effects of moving on move copy/assign operator only?

cpp/inc/bond/ext/grpc/unary_call.h
ara-ayvazyan: Are we sure that we want to return a smart pointer instead of something like `shared_unary_call`? That will change the access semantics for the user.

cpp/inc/bond/ext/grpc/detail/service_call_data.h
```diff
@@ -69,7 +69,7 @@ struct service_unary_call_data : io_manager_tag
           _cq(cq),
           _threadPool(threadPool),
           _cb(cb),
-          _receivedCall(std::make_shared<unary_call_impl<TRequest, TResponse>>())
+          _receivedCall(new unary_call_impl<TRequest, TResponse>)
```
ara-ayvazyan: `make_unique`?

cpp/inc/bond/ext/grpc/detail/service_call_data.h
```diff
@@ -81,30 +81,32 @@ struct service_unary_call_data : io_manager_tag
...
+            // the unary_call to pass to the user callback, the unary_call
+            // and unary_call_impl start collaborating to manage the
+            // lifetime of the unary_call_impl.
             {
-                std::shared_ptr<unary_call_impl<TRequest, TResponse>> receivedCall = _receivedCall;
+                // We have to manually release the pointer from
+                // _receivedCall to capture it in the lambda. When we can
+                // use C++14, we can simplify this by using lambda's
+                // capture-by-move.
+                unary_call_impl<TRequest, TResponse>* receivedCall = _receivedCall.release();
```
ara-ayvazyan: `auto`?

cpp/inc/bond/ext/grpc/detail/service_call_data.h
```diff
@@ -81,30 +81,32 @@ struct service_unary_call_data : io_manager_tag
...
+                // If _threadPool->schedule throws an exception, we don't
+                // delete the receivedCall that was just released; however
+                // that exception is going to leak into io_manager, which
+                // can't handle unhandled exceptions, and will terminate the
+                // program.
             }
 
             // create new state for the next request that will be received
-            _receivedCall = std::make_shared<unary_call_impl<TRequest, TResponse>>();
+            _receivedCall.reset(new unary_call_impl<TRequest, TResponse>);
```
ara-ayvazyan: `make_unique` here as well?

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -34,50 +35,96 @@ namespace detail {
...
         std::atomic_flag _responseSentFlag;
-        /// A pointer to ourselves used to keep us alive while the response is
-        /// being sent.
-        std::shared_ptr<unary_call_impl> _self;
+        std::atomic<size_t> _refCount;
 
         unary_call_impl()
             : _context(),
             _request(),
             _responder(&_context),
```
ara-ayvazyan: Comment that `_context` must be defined before `_responder`?

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -34,50 +35,96 @@ namespace detail {
...
+                // still ONE user reference. In this case, the user
+                // reference could go away at any time and delete us, so we
+                // can't touch any part of this object. But we don't need to
+                // send anything anyway.
+                if (trySend == ReleaseTrySend::Yes)
+                {
+                    FinishWithError(
+                        grpc::Status{
+                            grpc::StatusCode::INTERNAL,
+                                "An internal server error has occurred." });
```
ara-ayvazyan: Minor: align text with status argument?

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -34,50 +35,96 @@ namespace detail {
...
         unary_call_impl(const unary_call_impl&) = delete;
         unary_call_impl& operator=(const unary_call_impl&) = delete;
 
+        enum class ReleaseTrySend
+        {
+            No,
+            Yes
+        };
+
+        void AddRef()
```
ara-ayvazyan: Both `AddRef` and `Release` should be private methods?

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -87,163 +134,282 @@ namespace detail {
...
+            // The response has been sent, so we no longer need to keep
+            // ourselves alive: release the implicit initial refcount that
+            // this instance was constructed with.
+            Release(ReleaseTrySend::No);
         }
     };
 
-}
+    template <typename TRequest, typename TResponse>
+    void intrusive_ptr_add_ref(unary_call_impl<TRequest, TResponse>* call)
```
ara-ayvazyan: Both `intrusive_ptr_add_ref` and `intrusive_ptr_release` should become `friend`s?

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -87,163 +134,282 @@ namespace detail {
...
-class unary_call final
-{
-public:
-    unary_call(unary_call&& other) noexcept
-        : _impl(std::move(other._impl))
+    template <typename TRequest, typename TResponse>
+    void intrusive_ptr_release(unary_call_impl<TRequest, TResponse>* call)
     {
-        other._impl.reset();
+        call->Release(unary_call_impl<TRequest, TResponse>::ReleaseTrySend::Yes);
```
ara-ayvazyan: You can use as `ReleaseTrySend` after you make it in-class.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -87,163 +134,282 @@ namespace detail {
...
+    public:
+        unary_call_base() = default;
+
+        unary_call_base(const unary_call_base&) = default;
+        unary_call_base& operator=(const unary_call_base&) = default;
+
+        unary_call_base(unary_call_base&& other) = default;
+        unary_call_base& operator=(unary_call_base&& rhs) = default;
+
+        ~unary_call_base()
```
ara-ayvazyan: Remove the dtor completely?

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -87,163 +134,282 @@ namespace detail {
...
+    /// shared_unary_call.
+    template <typename TRequest, typename TResponse>
+    class unary_call_base
     {
-        if (this != &rhs)
+    private:
+        using impl_type = unary_call_impl<TRequest, TResponse>;
+
+    public:
+        unary_call_base() = default;
```
ara-ayvazyan: I don't think we need to explicitly `=default` everything, especially there are no custom constructors. You might want to put the default constructed under `protected` section.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -87,163 +134,282 @@ namespace detail {
...
+        {
+            BOOST_ASSERT(_impl);
+            return *_impl;
+        }
+    };
+
+} // namespace detail
+
+template <typename TRequest, typename TResponse>
+class shared_unary_call;
```
ara-ayvazyan: Please separate each class into its own header.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -87,163 +134,282 @@ namespace detail {
...
+    /// @brief Assign to this unary_call by moving from another unary_call.
     ///
-    /// Only the first call to \p Finish or \p FinishWithError will be
-    /// honored.
-    void Finish(const bond::bonded<TResponse>& msg, const grpc::Status& status = grpc::Status::OK)
+    /// The moved-from instance is valid only for destruction and
+    /// assignment.
+    unary_call& operator=(unary_call&& rhs) = default;
+
+    ~unary_call() = default;
```
ara-ayvazyan: I think you have too many unnecessary `=default`s :)

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -87,163 +134,282 @@ namespace detail {
...
+        }
+
+    protected:
+        explicit unary_call_base(impl_type*&& impl) noexcept
+            : _impl(std::move(impl))
+        {
+            BOOST_ASSERT(_impl);
+        }
+
+        explicit operator bool() const
```
ara-ayvazyan: `BOND_NOEXCEPT`, there seem to be more functions that can be marked that way.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -87,163 +134,282 @@ namespace detail {
...
-    /// Only the first call to \p Finish or \p FinishWithError will be
-    /// honored.
-    void Finish(const bond::bonded<TResponse>& msg, const grpc::Status& status = grpc::Status::OK)
+    /// The moved-from instance is valid only for destruction and
+    /// assignment.
+    unary_call& operator=(unary_call&& rhs) = default;
+
+    ~unary_call() = default;
+
+    void swap(unary_call& rhs) noexcept
```
ara-ayvazyan: `using base_type::swap;`

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -87,163 +134,282 @@ namespace detail {
...
+        return shared_unary_call<TRequest, TResponse>(std::move(*this));
     }
 
     template <typename SUCDRequest, typename SUCDResponse, typename SUCDThreadPool>
     friend struct detail::service_unary_call_data;
 
 private:
-    std::shared_ptr<detail::unary_call_impl<TRequest, TResponse>> _impl;
+    explicit unary_call(impl_type*&& impl) noexcept
+        : base_type(std::move(impl))
```
ara-ayvazyan: Please remove the `&&` and do not `move`. Or even better, do `using base_type::base_type;` if we don't care about VC12 here.

cpp/inc/bond/ext/grpc/unary_call.h
chwarr: And, if not, they need to be inline.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -87,163 +134,282 @@ namespace detail {
...
+    shared_unary_call& operator=(shared_unary_call&&) = default;
+
+    ~shared_unary_call() = default;
+
+    /// @brief Returns true if this shared_unary_call is non-empty;
+    /// otherwise false.
+    ///
+    /// @warning An empty shared_unary_call can only be destroyed or
+    /// assigned to.
+    explicit operator bool() const
```
ara-ayvazyan: `using base_type::operator bool;` if you really want to be that explicit.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -87,163 +134,282 @@ namespace detail {
...
+    /// @warning An empty shared_unary_call can only be destroyed or
+    /// assigned to.
+    shared_unary_call() = default;
+
+    /// @brief Create a shared_unary_call by moving from an existing
+    /// unary_call.
+    ///
+    /// The shared_unary_call assumes ownership, and the provided unary_call
+    /// will be in the moved-from state.
+    explicit shared_unary_call(unary_call<TRequest, TResponse>&& other) noexcept
```
ara-ayvazyan: Keep it simple and remove the `&&`.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -87,163 +134,282 @@ namespace detail {
...
     {
-        BOOST_ASSERT(_impl);
+        return base_type::operator bool();
     }
 
-    /// Resets the underlying detail::unary_call_impl that this is wrapping.
-    /// If the previous \p _impl was non-null, attempts to send an internal
-    /// server error.
-    void reset(std::shared_ptr<detail::unary_call_impl<TRequest, TResponse>>&& newValue)
+    void swap(shared_unary_call& rhs) noexcept
```
ara-ayvazyan: `using base_type::swap;`

examples/cpp/grpc/async-server/async-server.cpp
```diff
@@ -0,0 +1,129 @@
...
+    // program.
+    explicit GreeterServiceImpl(std::shared_ptr<bond::ext::gRPC::thread_pool> tp)
+        : _tp(std::move(tp))
+    { }
+
+private:
+    using CallType = bond::ext::gRPC::unary_call<bond::bonded<HelloRequest>, HelloReply>;
+    using SharedCallType = bond::ext::gRPC::shared_unary_call<bond::bonded<HelloRequest>, HelloReply>;
+
+    std::shared_ptr<bond::ext::gRPC::thread_pool> _tp;
```
ara-ayvazyan: Prefer "types, member functions, data members" order?

examples/cpp/grpc/async-server/async-server.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+    class PerRequestState
+    {
+    public:
+        explicit PerRequestState(CallType call)
+        : _call(std::move(call).share())
```
ara-ayvazyan: Minor: align?

examples/cpp/grpc/async-server/async-server.cpp
```diff
@@ -0,0 +1,129 @@
...
+        // Callbacks to be copyable, so we switch to using
+        // shared_unary_call, which is copyable.
+        SharedCallType _call;
+
+        // Other state could be added as needed.
+    };
+
+    void SayHello(CallType call) override
+    {
+        PerRequestState perRequestState { std::move(call) };
```
ara-ayvazyan: Do you really need a named variable?

examples/cpp/grpc/async-server/async-server.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+int main()
+{
+    auto ioManager = std::make_shared<bond::ext::gRPC::io_manager>();
+    auto threadPool = std::make_shared<bond::ext::gRPC::thread_pool>();
+
+    GreeterServiceImpl service(threadPool);
+
+    bond::ext::gRPC::server_builder builder;
+    builder.SetThreadPool(threadPool);
```
ara-ayvazyan: Minor: concatenate the calls for a builder?

examples/cpp/grpc/async-server/async-server.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+    const std::string user("world");
+
+    HelloRequest request;
+    request.name = user;
+
+    bond::ext::gRPC::wait_callback<HelloReply> cb;
+    greeter.AsyncSayHello(request, cb);
+
+    bool waitResult = cb.wait_for(std::chrono::seconds(10));
```
ara-ayvazyan: Too many descriptive variables sometimes has the opposite effect on readability in my opinion. The `if (!cb.wait_for(...))` seems cleaner.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -34,50 +35,96 @@ namespace detail {
...
         TRequest _request;
         grpc::ServerAsyncResponseWriter<bond::bonded<TResponse>> _responder;
         /// Tracks whether any response has been sent yet.
         std::atomic_flag _responseSentFlag;
-        /// A pointer to ourselves used to keep us alive while the response is
-        /// being sent.
-        std::shared_ptr<unary_call_impl> _self;
+        std::atomic<size_t> _refCount;
 
         unary_call_impl()
```
ara-ayvazyan: If we don't care about VC12 here, then might be better to have in-class initializations and `=default` this constructor.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -34,50 +35,96 @@ namespace detail {
...
+                // Finish/FinishWithError and completion of sending that
+                // response has just happened via invoke(), AND there is
+                // still ONE user reference. In this case, the user
+                // reference could go away at any time and delete us, so we
+                // can't touch any part of this object. But we don't need to
+                // send anything anyway.
+                if (trySend == ReleaseTrySend::Yes)
+                {
+                    FinishWithError(
+                        grpc::Status{
```
ara-ayvazyan: I think you can skip the `grpc::Status` part.

cpp/inc/bond/ext/grpc/unary_call.h
ara-ayvazyan: The in-class defined functions are implicitly inline.

cpp/inc/bond/ext/grpc/detail/service_call_data.h
chwarr: `std::make_unique` was only added in C++14.

cpp/inc/bond/ext/grpc/detail/service_call_data.h
chwarr: `std::make_unique` was only added in C++14.

cpp/inc/bond/ext/grpc/unary_call.h
chwarr: Will do after resolving the bulk of the other comments, as GitHub has trouble tracking comments across moves.

cpp/inc/bond/ext/grpc/unary_call.h
chwarr: Thanks. That's a copy/paste mistake.

cpp/inc/bond/ext/grpc/detail/service_call_data.h
ara-ayvazyan: Oh, right.

cpp/inc/bond/ext/grpc/detail/service_call_data.h
```diff
@@ -81,30 +81,32 @@ struct service_unary_call_data : io_manager_tag
...
+            // the unary_call to pass to the user callback, the unary_call
+            // and unary_call_impl start collaborating to manage the
+            // lifetime of the unary_call_impl.
             {
-                std::shared_ptr<unary_call_impl<TRequest, TResponse>> receivedCall = _receivedCall;
+                // We have to manually release the pointer from
+                // _receivedCall to capture it in the lambda. When we can
+                // use C++14, we can simplify this by using lambda's
+                // capture-by-move.
+                auto receivedCall = _receivedCall.release();
```
ara-ayvazyan: I would avoid doing this, because it will require all the scheduled functions to be executed in order to avoid memory leak, or ever worse functional misbehavior. Given that `unary_call_impl` already has a reference counter you can wrap it into `boost::intrusive_ptr` (which is copyable) and capture it that way. This will make sure that the lifetime of that object does not rely on the function invocation. Also note that even with C++14 support you might not be able to move-capture if the scheduler uses `std::function` to store the functions which is unable to store move-only objects.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -34,40 +35,33 @@ namespace detail {
...
+    /// usual 0 that many boost::intrusive_ptr implementations use. When the
+    /// notification that the response has been sent is dequeued from the
+    /// completion queue, %invoke() calls %Release() on itself, decrementing
+    /// the ref count, and allowing the remaining unary_call and
+    /// shared_unary_call objects, if any, to control lifetime.
     template <typename TRequest, typename TResponse>
-    struct unary_call_impl final
-        : std::enable_shared_from_this<unary_call_impl<TRequest, TResponse>>,
-          io_manager_tag
+    struct unary_call_impl final : io_manager_tag
```
ara-ayvazyan: Minor: this looks little more complex than a simply struct (data record).

cpp/test/grpc/unary_call.cpp
```diff
@@ -0,0 +1,53 @@
...
+    static void DefaultShared_IsEmpty()
+    {
+        shared_unary_callBox c{};
+        UT_AssertIsFalse(static_cast<bool>(c));
+    }
+
+    static void Initialize()
+    {
+        UnitTestSuite suite("unary_call");
+        suite.AddTestCase(&DefaultShared_IsEmpty, "DefaultShared_IsEmpty");
```
ara-ayvazyan: Use Boost.Test for new ones?

cpp/inc/bond/ext/grpc/detail/service_call_data.h
chwarr: > it will require all the scheduled functions to be executed in order to avoid memory leak

At the moment, it's [already a requirement that the scheduler invoke all callbacks scheduled to avoid memory leaks](https://github.com/Microsoft/bond/blob/70ced821b31a38a93582f5885d7b22ca9ab7638c/cpp/inc/bond/ext/grpc/thread_pool.h#L35-L38). Should this be revisited? Yes. In this PR, I'd rather not: we're not making anything any worse, and we are making the unary_call interface better. Not fixing this now also doesn't prevent or make it harder to fix it later.

> Given that `unary_call_impl` already has a reference counter you can wrap it into `boost::intrusive_ptr` (which is copyable) and capture it that way.

Let me see if I can rework this. I think I need to adjust the refcounting to make this work, but it will end up even better than before.

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ different versioning scheme, following the Haskell community's
   arguments.)
 * Fixed some macro uses that did not have sufficient parenthesis around
   parameters and resulted in compiler errors.
+* Added the `bond::ext::gRPC::shared_unary_call` type. This type can be used
```
ara-ayvazyan: Truncated line?

cpp/inc/bond/ext/grpc/detail/service_call_data.h
```diff
@@ -81,30 +84,26 @@ struct service_unary_call_data : io_manager_tag
...
+            // the unary_call to pass to the user callback, the unary_call
+            // and unary_call_impl start collaborating to manage the
+            // lifetime of the unary_call_impl.
             {
-                std::shared_ptr<unary_call_impl<TRequest, TResponse>> receivedCall = _receivedCall;
+                // We have to manually release the pointer from
+                // _receivedCall to capture it in the lambda. When we can
+                // use C++14, we can simplify this by using lambda's
+                // capture-by-move.
+                auto receivedCall = boost::intrusive_ptr<uc_impl>{ _receivedCall.release() };
```
ara-ayvazyan: Should `_receivedCall` be changed to `boost::intrusive_ptr` as well?

chwarr: Not easily that I can see. `uc_impl::Release` is responsible for the guaranteed send behavior, but that is only needed _after_ a call has been received. By keeping the `uc_impl` in a `unique_ptr` in `service_unary_call_data`, if a call is never received, the we just delete the `uc_impl` when the service instance is destroyed. Its only when a call has been received that we start refcounting and triggering the guaranteed send behavior.

I tried moving this has-call-been-received behavior inside of `uc_impl`, but it got gross quickly.

CHANGELOG.md
chwarr: Fixed. Thanks.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -34,40 +35,28 @@ namespace detail {
...
+    /// usual 0 that many boost::intrusive_ptr implementations use. When the
+    /// notification that the response has been sent is dequeued from the
+    /// completion queue, %invoke() calls %Release() on itself, decrementing
+    /// the ref count, and allowing the remaining unary_call and
+    /// shared_unary_call objects, if any, to control lifetime.
     template <typename TRequest, typename TResponse>
-    struct unary_call_impl final
-        : std::enable_shared_from_this<unary_call_impl<TRequest, TResponse>>,
-          io_manager_tag
+    class unary_call_impl final : io_manager_tag
```
ara-ayvazyan: Should this be derived as `public`?

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -34,40 +35,28 @@ namespace detail {
...
-            _responder(&_context),
-            _responseSentFlag(),
-            _self()
-        { }
+    public:
+        // A pointer to the context is passed to _responder when
+        // constructing it, so this needs to be declared before _responder.
+        grpc::ServerContext _context{};
+        TRequest _request{};
+        grpc::ServerAsyncResponseWriter<bond::bonded<TResponse>> _responder{ &_context };
```
ara-ayvazyan: Why are these still `public`? Add access function for `_responder` as well and make these private.

cpp/inc/bond/ext/grpc/unary_call.h
chwarr: No.

cpp/inc/bond/ext/grpc/unary_call.h
chwarr: They're used by `service_unary_call_data`, a close collaborator. Both are detail classes. This is intentional.

cpp/inc/bond/ext/grpc/unary_call.h
ara-ayvazyan: OK, then probably better to explicitly mark as `private`.

cpp/inc/bond/ext/grpc/unary_call.h
chwarr: Switched to accessor methods. See commit 2718690.

chwarr:
Split `unary_call` &c into multiple files and rebased atop the current master (commit 85995fe6).

chwarr:
Just fixed a Clang compiler error.

Issue #641, opened 2017-10-11
Build Haskell with customized proj file instead of vcxproj file


hcoona:
Use `vcxproj` file compile Haskell codes lead to a hard dependency to Visual C++ which is not available on other platform now. I suggest to use a customized `proj` file instead to eliminate this unnecessary dependency.

Related PR #633

hcoona:
I didn't know the impact to developer. Dropped vcxproj, the developer cannot list the Haskell files in Visual Studio. If I use `csproj`, they can list them but without "Virtual Folder" function.

chwarr:
It's fine if the .hs files don't show up in Visual Studio when using the cs.sln.

chwarr:
Fixed in commit 0822d27.

PR #647, opened 2017-10-18
[build] Improve build system for haskell project without hard dependency to VC

hcoona:
See #641 for details discussion.

> Use `vcxproj` file compile Haskell codes lead to a hard dependency to Visual C++ which is not available on other platform now. I suggest to use a customized `proj` file instead to eliminate this unnecessary dependency.

The current changes would use a `csproj` to invoke the `cmake` to build the `gbc` binary. This would get rid of the hard dependency to Visual C++.

The current changes keeps the same behaviors of origin `vcxproj`.
1. It builds only on Windows
1. It skips on other platform
1. It invoke `cmake` to build gbc.
1. It works with xbuild
1. It works with msbuild
1. It works with the `sln` file in particular with the `postProject` metadata

I think we can further improve it in the future to get rid of `cmake` & build scripts in `.travis.yml`, make this `csproj` work on multiple platforms.


cs/Compiler.csproj
```diff
@@ -0,0 +1,96 @@
...
+
+    <MakeDir Directories="$(IntermediateOutputPath)" />
+
+    <Exec Command="cmake $(ProjectDir)\..\compiler -Wno-dev" WorkingDirectory="$(IntermediateOutputPath)" />
+    <Exec Command="cmake --build . --target gbc" WorkingDirectory="$(IntermediateOutputPath)" />
+    
+    <Copy SourceFiles="$(IntermediateOutputPath)\build\gbc\gbc.exe" DestinationFolder="$(ProjectDir)\tools\" />
+  </Target>
+
+  <Target Name="ExternalBuild" DependsOnTargets="$(ExternalBuildDependsOn)" />
```
chwarr: It looks like ExternalBuild and CoreExternalBuild can be merged together, reducing the number of Targets. Any reason to keep them split?

cs/Compiler.csproj
```diff
@@ -0,0 +1,96 @@
...
+
+  <Target Name="SkipExternalBuild">
+    <Message Text="Skip external building because it's only for Windows" Importance="High" />
```
chwarr: Minor: can the message text be changed to "**Skipping** external build because it's only for Windows"?

cs/Compiler.csproj
```diff
@@ -0,0 +1,96 @@
...
+
+  <Import Project="$(MSBuildThisFileDirectory)\build\internal\Common.Internal.targets" />
+
+  <PropertyGroup>
+    <ExternalBuildDependsOn>
+      CoreExternalBuild;
+    </ExternalBuildDependsOn>
+  </PropertyGroup>
+
+  <Choose>
```
chwarr: I find that Choose makes sense when there are more things to set than just one property. For just this one property, this is shorter:

```
<PropertyGroup>
    <BuildDependsOn>SkipExternalBuild</BuildDependsOn>
    <BuildDependsOn Condition=" '$(OS)' == 'Windows_NT' ">ExternalBuild</BuildDependsOn>
</PropertyGroup>
```

hcoona:
@chwarr I fixed the comments, please help to review it again.

chwarr:
Thank you very much for this change!

PR #648, opened 2017-10-18
java core: Implement j.io.Serializable for concrete types.

tstein:
The actual serialization is CBv1 marshalling. With this commit, attempting to put a generic Bond-generated type through an API expecting a Serializable will throw.

I checked whether we could get rid of `serialVersionUID` - looks like the header and the fqcn/sVUID matching logic are the same between `Serializable` and `Externalizable`. See more [here](http://bugs.java.com/bugdatabase/view_bug.do?bug_id=4094702).

### DIFF
--- a/compiler/src/Language/Bond/Codegen/Java/Class_java.hs
+++ b/compiler/src/Language/Bond/Codegen/Java/Class_java.hs
@@ -494,7 +494,7 @@ javaNativeSerializationGlue declName = [lt|
     @Override
     public void writeExternal(java.io.ObjectOutput out) throws java.io.IOException {
         final java.io.ByteArrayOutputStream outStream = new java.io.ByteArrayOutputStream();
-        final org.bondlib.ProtocolWriter writer = new org.bondlib.CompactBinaryWriter(outStream, 2);
+        final org.bondlib.ProtocolWriter writer = new org.bondlib.CompactBinaryWriter(outStream, 1);
         org.bondlib.Marshal.marshal(this, writer);
 
         final byte[] marshalled = outStream.toByteArray();
@@ -506,7 +506,7 @@ javaNativeSerializationGlue declName = [lt|
     public void readExternal(java.io.ObjectInput in) throws java.io.IOException, java.lang.ClassNotFoundException {
         final int marshalledLength = in.readInt();
         final byte[] marshalled = new byte[marshalledLength];
-        final int bytesRead = in.read(marshalled);
+        in.readFully(marshalled);
 
         final java.io.ByteArrayInputStream inStream = new java.io.ByteArrayInputStream(marshalled);
         this.__deserializedInstance = org.bondlib.Unmarshal.unmarshal(inStream, getBondType()).deserialize();
--- a/compiler/src/Language/Bond/Codegen/Java/Class_java.hs
+++ b/compiler/src/Language/Bond/Codegen/Java/Class_java.hs
@@ -498,12 +498,14 @@ javaNativeSerializationGlue declName = [lt|
         org.bondlib.Marshal.marshal(this, writer);
 
         final byte[] marshalled = outStream.toByteArray();
+        out.write(0);   // This type is not generic and has zero type parameters.
         out.writeInt(marshalled.length);
         out.write(marshalled);
     }
 
     @Override
     public void readExternal(java.io.ObjectInput in) throws java.io.IOException, java.lang.ClassNotFoundException {
+        if (!(in.read() == 0)) throw new java.io.IOException("type is not generic, but serialized data has type parameters.");
         final int marshalledLength = in.readInt();
         final byte[] marshalled = new byte[marshalledLength];
         in.readFully(marshalled);
--- a/compiler/src/Language/Bond/Codegen/Java/Class_java.hs
+++ b/compiler/src/Language/Bond/Codegen/Java/Class_java.hs
@@ -505,7 +505,7 @@ javaNativeSerializationGlue declName = [lt|
 
     @Override
     public void readExternal(java.io.ObjectInput in) throws java.io.IOException, java.lang.ClassNotFoundException {
-        if (!(in.read() == 0)) throw new java.io.IOException("type is not generic, but serialized data has type parameters.");
+        if (in.read() != 0) throw new java.io.IOException("type is not generic, but serialized data has type parameters.");
         final int marshalledLength = in.readInt();
         final byte[] marshalled = new byte[marshalledLength];
         in.readFully(marshalled);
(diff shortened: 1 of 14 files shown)


compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -473,6 +473,70 @@ object_hashCode fields structBase = [lt|@Override
...
+javaNativeSerializationGlue :: String -> Text
+javaNativeSerializationGlue declName = [lt|
+    // Java native serialization
+    private static final long serialVersionUID = 0L;
+    private #{declName} __deserializedInstance;
+
+    @Override
+    public void writeExternal(java.io.ObjectOutput out) throws java.io.IOException {
+        final java.io.ByteArrayOutputStream outStream = new java.io.ByteArrayOutputStream();
+        final org.bondlib.ProtocolWriter writer = new org.bondlib.CompactBinaryWriter(outStream, 2);
```
chwarr: The commit message says CBv1, but isn't this CBv2?

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -473,6 +473,70 @@ object_hashCode fields structBase = [lt|@Override
...
+        org.bondlib.Marshal.marshal(this, writer);
+
+        final byte[] marshalled = outStream.toByteArray();
+        out.writeInt(marshalled.length);
+        out.write(marshalled);
```
chwarr: If only we had an ObjectOutputStream. We could use `outStream.writeTo`...

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -473,6 +473,70 @@ object_hashCode fields structBase = [lt|@Override
...
+    }
+
+    @Override
+    public void readExternal(java.io.ObjectInput in) throws java.io.IOException, java.lang.ClassNotFoundException {
+        final int marshalledLength = in.readInt();
+        final byte[] marshalled = new byte[marshalledLength];
+        final int bytesRead = in.read(marshalled);
```
chwarr: From the API documentation for [`ObjectInput.read`](http://docs.oracle.com/javase/7/docs/api/java/io/ObjectInput.html#read(byte[])), it looks like `read` needs to be called in a loop until the requested number of bytes have been read or EOF is signaled.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: Good catch.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: You know what the sad thing is? I dug into it with the debugger and it looks like it's actually going to be an `ObjectOutputStream` IRL. They just... didn't want to give you the stream guarantee?

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: I don't get that implication from that doc, but I dug into this mess and found that [`DataInput.readFully()`](http://docs.oracle.com/javase/6/docs/api/java/io/DataInput.html#readFully(byte[])) guarantees the behavior you want. `ObjectInput extends DataInput`, though the doc that suggests it is only in the method of `ObjectInputStream` with the [exact same signature](http://docs.oracle.com/javase/6/docs/api/java/io/ObjectInputStream.html#read(byte[],%20int,%20int)).

☕️🤷‍♂️

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -535,6 +599,8 @@ public class #{typeNameWithParams declName declParams}#{maybe interface baseClas
     }
     #{bondTypeDescriptorInstanceVariableDecl}
 
+    #{ifThenElse (null declParams) (javaNativeSerializationGlue declName) javaNativeSerializationUnimpl}
```
sigansky: >#{ifThenElse (null declParams) (javaNativeSerializationGlue declName) javaNativeSerializationUnimpl} [](start = 4, length = 100)

Pls add support for generic types.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -473,6 +473,70 @@ object_hashCode fields structBase = [lt|@Override
...
+    }
+
+    @Override
+    public void readExternal(java.io.ObjectInput in) throws java.io.IOException, java.lang.ClassNotFoundException {
+        final int marshalledLength = in.readInt();
+        final byte[] marshalled = new byte[marshalledLength];
+        in.readFully(marshalled);
+
+        final java.io.ByteArrayInputStream inStream = new java.io.ByteArrayInputStream(marshalled);
+        this.__deserializedInstance = org.bondlib.Unmarshal.unmarshal(inStream, getBondType()).deserialize();
```
sigansky: >this.__deserializedInstance = org.bondlib.Unmarshal.unmarshal(inStream, getBondType()).deserialize(); [](start = 8, length = 101)

I suggest against polluting every object's state (per hierarchy) by this temporary instance variable. It should be easy to implement an "unmarshall" version that targets a particular object instead of creating a new one.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: Added a byte in front of all of this to represent the number of type parameters, which is a fixed 0 for concrete types.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -498,12 +498,14 @@ javaNativeSerializationGlue declName = [lt|
...
 
         final byte[] marshalled = outStream.toByteArray();
+        out.write(0);   // This type is not generic and has zero type parameters.
         out.writeInt(marshalled.length);
         out.write(marshalled);
     }
 
     @Override
     public void readExternal(java.io.ObjectInput in) throws java.io.IOException, java.lang.ClassNotFoundException {
+        if (!(in.read() == 0)) throw new java.io.IOException("type is not generic, but serialized data has type parameters.");
```
chwarr: `!=` ?

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -494,19 +494,21 @@ javaNativeSerializationGlue declName = [lt|
...
```
chwarr: It looks like the golden tests haven't been updated.

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
tstein: As discussed offline, I'd like to make the trade of having this state to avoid writing a new top-level API in this commit.

PR #649, opened 2017-10-19
[build] Use a docker image for building c++

ara-ayvazyan:

PR #650, opened 2017-10-20
[gbc] Switch from Parsec to Megaparsec

lalo:
- gbc can only be built with GHC >= 8.0.1
- stack resolver has been updated

- Not working on travis, will look after https://github.com/Microsoft/bond/pull/649 gets merged.
- This page is handy to see the definitions of the parsec parsers that we used to use (specially number related) in case you want to deeply code review the Lexer.hs https://hackage.haskell.org/package/parsec-3.1.11/docs/Text-Parsec-Token.html


lalo:
This is the travis error I'm hitting https://travis-ci.org/Microsoft/bond/jobs/290239737#L2191

chwarr:
To debug the Travis error, try [unsetting the CC and CXX variables that we point to ccache](https://github.com/Microsoft/bond/blob/85995fe6372d423ce9973d20bbdd4e39de07cc80/.travis.yml#L56-L57). Based on the results of that, we can figure out our next step.

ara-ayvazyan:
I think I've figured out the problem. We should not be doing `apt-get install haskell-stack` and instead should install the latest version by doing `wget -qO- https://get.haskellstack.org/ | sh` as described [here](https://docs.haskellstack.org/en/stable/install_and_upgrade/#linux). Also will need to pass `--allow-different-user` to `stack setup` and `stack build`.

lalo:
I'm 80% certain we need a more recent Stack build, waiting on https://github.com/Microsoft/bond/pull/673 to be merged and dockerfile to be udpated

compiler/src/Language/Bond/Codegen/CustomMapping.hs
```diff
@@ -11,11 +11,13 @@ module Language.Bond.Codegen.CustomMapping
     , parseNamespaceMapping
```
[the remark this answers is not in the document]
lalo: No idea why this happened, I'll fix it.

compiler/src/Language/Bond/Codegen/CustomMapping.hs
```diff
@@ -34,46 +36,52 @@ data NamespaceMapping = NamespaceMapping
...
     }
 
-type Parser a = Parsec SourceName () a
+type Parser = Parsec Void String
 
-whitespace :: Parser String
-whitespace = many (char ' ') <?> "whitespace"
 identifier :: Parser String
-identifier = many1 (alphaNum <|> char '_') <?> "identifier"
+identifier = some (alphaNumChar <|> char '_') <?> "identifier"
```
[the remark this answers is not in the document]
lalo: Should I fix it or should we go bug for bug and then start revamping the parser?

compiler/src/Language/Bond/Codegen/CustomMapping.hs
```diff
@@ -34,46 +36,52 @@ data NamespaceMapping = NamespaceMapping
...
+symbol = L.symbol sc
+
 equal :: Parser String
 equal = symbol "="
+
 integer :: Parser Integer
-integer = decimal <$> many1 digit <?> "decimal number"
-  where
-    decimal = foldl (\x d -> 10 * x + toInteger (digitToInt d)) 0
+integer = L.decimal
```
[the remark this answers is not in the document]
lalo: yes, so integer should be positive and negative

compiler/src/Language/Bond/Codegen/CustomMapping.hs
```diff
@@ -34,46 +36,52 @@ data NamespaceMapping = NamespaceMapping
...
-identifier = many1 (alphaNum <|> char '_') <?> "identifier"
+identifier = some (alphaNumChar <|> char '_') <?> "identifier"
+
 qualifiedName :: Parser [String]
 qualifiedName = sepBy1 identifier (char '.') <?> "qualified name"
+
+sc :: Parser ()
+sc = L.space space1 lineCmnt blockCmnt
+  where
+    lineCmnt  = L.skipLineComment "//"
```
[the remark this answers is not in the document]
lalo: right

compiler/stack.yaml
```diff
@@ -2,7 +2,15 @@
 #
 # Some commonly used options have been documented as comments in this file.
 # For advanced use and comprehensive documentation of the format, please see:
-# http://docs.haskellstack.org/en/stable/yaml_configuration/
+# https://docs.haskellstack.org/en/stable/yaml_configuration/
+
+# A warning or info to be displayed to the user on config load.
```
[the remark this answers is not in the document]
lalo: I wanted to keep it during iterations because it was a reminder to make sure stack did the right thing. I had some complications before updating my stack.

compiler/src/Language/Bond/Codegen/CustomMapping.hs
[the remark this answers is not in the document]
lalo: this made me think by reading the parsec documentation, will integer accept negative hex and oct, I'll check that out

compiler/src/Language/Bond/Codegen/CustomMapping.hs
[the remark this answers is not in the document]
lalo: fixed

compiler/src/Language/Bond/Codegen/CustomMapping.hs
```diff
@@ -34,46 +36,65 @@ data NamespaceMapping = NamespaceMapping
...
+octal = lexeme . try $ char '0' >> char' 'o' >> L.octal
+
+natural :: Parser Integer
+natural = hexadecimal <|> octal <|> decimal
+
 integer :: Parser Integer
-integer = decimal <$> many1 digit <?> "decimal number"
-  where
-    decimal = foldl (\x d -> 10 * x + toInteger (digitToInt d)) 0
+integer = natural <|> L.signed sc decimal
```
lalo: this should also accept signed naturals a.k.a. -0xFF and -0o17, at least that's what parsec integer did. Will fix.

compiler/src/Language/Bond/Codegen/CustomMapping.hs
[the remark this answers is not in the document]
chwarr: An issue is fine.

compiler/src/Language/Bond/Parser.hs
```diff
@@ -93,31 +72,38 @@ processImport :: Import -> Parser()
...
             void $ local (\e -> e { currentFile = path }) bond
 
 -- parser for struct, enum or type alias declaration/definition
 declaration :: Parser Declaration
 declaration = do
+    -- When adding a new Declaration parser, order matters in the following command.
+    -- Parsers must fail to consume ANY token for the next parser to be able to succesfully work
+    -- unless the parser is encapsulated in a try statement. For more info on try and <|> see:
+    -- https://hackage.haskell.org/package/megaparsec-6.2.0/docs/Text-Megaparsec.html#v:try
     decl <- try forward
```
lalo: We want to avoid these try statements that encompass whole parsers, this will avoid backtracking and losing information for better error messages. Try statements should be used when parsing 'branches' or where the syntax diverges. This way if when we successfully parse the token 'service', and then later we hit a parsing error we don't backtrack all the way to here, instead just fail iniside the service parser. This is trickier with struct and struct view that share some common syntax.

chwarr: Are you planning to fold forward into structDeclaration?

compiler/extra/gbc-check.py
```diff
@@ -0,0 +1,94 @@
...
+        shutil.rmtree(second_dir)
+
+def main():
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument('--stable_gbc', required=True, help="Path to stable base gbc")
+    parser.add_argument('--updated_gbc', required=True, help="Path to updated gbc")
+    parser.add_argument('--output_dir', required=True, help="Directory where script will write to")
+    parser.add_argument('--input_dir', required=True, help="Directory of bond files to be tested, can have nested dirs with bond files")
+    parser.add_argument('--gbc_args', required=True, help="Args to gbc i.e. 'cpp' or 'cs' to be appended on every gbc command")
```
chwarr: Minor: "e.g." instead of "i.e."

compiler/extra/gbc-check.py
```diff
@@ -0,0 +1,94 @@
+
+# Simple script to test easily multiple bond files and compare the
+# output of two versions of gbc. Output includes all codegen output,
+# stdout and stderr. Also compares quantity of codegen files generated
+# and exit code. A diff will be outputed per file.
```
chwarr: Minor: "output"

compiler/extra/gbc-check.py
```diff
@@ -0,0 +1,94 @@
...
+
+def generate_bond(gbc, gbc_flags, output_dir, bond_file):
+    bond_command = "{} {} -o {} {}".format(gbc, gbc_flags, output_dir, bond_file)
+
+    if not os.path.exists(output_dir):
+        os.makedirs(output_dir)
+
+    with open(os.path.join(output_dir,"stdout.txt"), "w+") as stdout_file:
+        with open(os.path.join(output_dir,"stderr.txt"), "w+") as stderr_file:
+            return_code = subprocess.call(bond_command, shell=True, stdout=stdout_file, stderr=stderr_file)
```
chwarr: It looks like the following will also work and will be more robust to things like spaces in paths:
```
subprocess.call(gbc, gbc_flags, "-o", output_dir, bond_file, shell=False, stdout=stdout_file, stderr=stderr_file)
```

compiler/extra/gbc-check.py
```diff
@@ -0,0 +1,94 @@
...
+                        file_2.readlines(),
+                        fromfile='stable_gbc',
+                        tofile='updated_gbc',
+                    )
+                    lines = False
+                    for line in diff:
+                        lines = True
+                        print(line)
+                    if lines:
+                        print("ERROR File {}: has generated diff on generated file/output {}".format(bond_file, a))
```
chwarr: Minor: "..has different generated output..."?

compiler/extra/gbc-check.py
```diff
@@ -0,0 +1,94 @@
...
+        for (a, b) in zip(list_files_1, list_files_2):
+            with open(os.path.join(first_dir, a), 'r') as file_1:
+                with open(os.path.join(second_dir, b), 'r') as file_2:
+                    diff = difflib.unified_diff(
+                        file_1.readlines(),
+                        file_2.readlines(),
+                        fromfile='stable_gbc',
+                        tofile='updated_gbc',
+                    )
+                    lines = False
```
chwarr: anyDiffs instead of lines?

compiler/src/Language/Bond/Parser.hs
lalo: No since it would be the same code. If I move it to structDecleration

```
forward = Forward <$> asks currentNamespaces <*> name <*> parameters <* semi <?> "forward declaration"
  where
name = keyword "struct" *> identifier
```
would end up sharing everything except the ```semi``` which would end up being a ```try semi``` and no benefit would be achieved since there is nothing else to parse after that. After that try statement we would end up running the other parsers all the time since if its not a semi it will likely be a struct or a view of.

lalo: Plus forwards don't have attributes and it would complicate the code for no benefit at all.

PR #651, opened 2017-10-21
[c#] Fix cs.sln load in Visual Studio 2015/2017

chwarr:
* The project type GUID was still for a C++ project. It has been
  switched to a C# project.
    * This caused a platform change from Win32 to Any CPU
* The cloning example was missing a solution-level dependency on
  Bond.IO.dll. One has been added.


chwarr:
FYI: @hcoona, I had to adjust the cs.sln to get it to load again in Visual Studio.

hcoona:
Thank you, I didn't notice that yet, and I didn't know the GUID would identify the project type.

PR #652, opened 2017-10-24
[build] Use a docker image for all builds

ara-ayvazyan:


.travis.yml
```diff
@@ -24,77 +23,10 @@ env:
...
-        export CC="ccache clang -Qunused-arguments --system-header-prefix=boost/";
-      fi
-
-    # nunit installation
-    - if [ "$FLAVOR" == "cs" ]; then travis_retry nuget install NUnit.Runners -version 2.6.4; fi
-
-before_script:
-    # restore nuget packages for solution
-    - if [ "$FLAVOR" == "cs" ]; then travis_retry nuget restore cs/cs.sln; fi
+    - travis_retry curl -L https://bondbinaries.blob.core.windows.net/bondbinaries/docker-images/bond-xenial-new.tar.xz | time docker load
```
ara-ayvazyan: Will switch to original file without `-new` after the review

tools/image-builder/build_bond.zsh
```diff
@@ -0,0 +1,82 @@
...
+            "cpp-comm") CPP_CMAKE_ARGS="-DBOND_ENABLE_COMM=TRUE -DBOND_SKIP_GBC_TESTS=TRUE -DBOND_SKIP_CORE_TESTS=TRUE -DBOND_ENABLE_GRPC=FALSE";;
+            "cpp-grpc") CPP_CMAKE_ARGS="-DBOND_SKIP_GBC_TESTS=TRUE -DBOND_SKIP_CORE_TESTS=TRUE -DgRPC_ZLIB_PROVIDER=package";;
+            *) echo "Unknown FLAVOR=$FLAVOR"; exit 1;;
+        esac
+
+        export BOOST_ROOT=/opt/boosts/boost_`echo $BOOST | tr . _`
+        export CXX="ccache clang++ -Qunused-arguments --system-header-prefix=boost/"
+        export CC="ccache clang -Qunused-arguments --system-header-prefix=boost/"
+
+        ln -s /root/.ccache $HOME/.ccache
```
tstein: No harm in making the symlink, even if we never run `ccache`. Can we do all the `ln -s`es together at the top?

tools/image-builder/build_bond.zsh
```diff
@@ -0,0 +1,82 @@
...
+
+mkdir -p $HOME $BUILD_PATH
+ln -s /root/.stack $HOME/.stack
+
+cd $BUILD_PATH
+
+case "$FLAVOR" in
+    cpp-*)
+        case "$FLAVOR" in
+            "cpp-core") CPP_CMAKE_ARGS="-DBOND_SKIP_GBC_TESTS=TRUE -DBOND_ENABLE_GRPC=FALSE";;
```
tstein: ~~ I guess the quotes here are unnecessary. Either way is fine with me.

tools/image-builder/build_bond.zsh
```diff
@@ -0,0 +1,82 @@
+#!/bin/zsh
+
+set -eux
+setopt shwordsplit
+
+HOME=$1
+FLAVOR=$2
+BOOST=${3:-}
+
+BUILD_PATH=/root/build/$FLAVOR
```
tstein: What's the purpose of having separate build dirs for each flavor? One of the goals of containerizing this, at least for me, is creating a stateless build - every time you run the container with one flavor, `/root/build` shouldn't exist.

tools/image-builder/build_bond.zsh
```diff
@@ -0,0 +1,82 @@
...
+
+        export BOOST_ROOT=/opt/boosts/boost_`echo $BOOST | tr . _`
+        export CXX="ccache clang++ -Qunused-arguments --system-header-prefix=boost/"
+        export CC="ccache clang -Qunused-arguments --system-header-prefix=boost/"
+
+        ln -s /root/.ccache $HOME/.ccache
+
+        cmake $CPP_CMAKE_ARGS /root/bond
+
+        make --jobs 2 check
```
tstein: Absolutely not a blocker for this change, but let's spend a few minutes checking whether we want to replace this 2 with some logic.

tools/image-builder/build_bond.zsh
```diff
@@ -0,0 +1,82 @@
...
+        export CXX="ccache clang++ -Qunused-arguments --system-header-prefix=boost/"
+        export CC="ccache clang -Qunused-arguments --system-header-prefix=boost/"
+        ln -s /root/.ccache $HOME/.ccache
+
+        nuget restore /root/bond/cs/cs.sln
+
+        cmake -DBOND_SKIP_GBC_TESTS=TRUE -DBOND_SKIP_CORE_TESTS=TRUE -DBOND_ENABLE_GRPC=FALSE /root/bond
+
+        make gbc
+        make DESTDIR=/root install
```
tstein: The boosts have mutually exclusive install locations, so they make sense in `/opt`, and the NUnit runner doesn't really belong... anywhere, but it seems like letting `make install` put `gbc` in `/usr/local/bin` would be appropriate. That gets `gbc` into the default `$PATH`, which will definitely simplify Java and probably simplify C#.

tools/image-builder/build_bond.zsh
```diff
@@ -0,0 +1,82 @@
...
+
+        ln -s /root/.ccache $HOME/.ccache
+
+        cmake $CPP_CMAKE_ARGS /root/bond
+
+        make --jobs 2 check
+        ;;
+
+    "cs")
+        # TODO: Remove build dependency on C++
```
tstein: If you don't want to lose track of this improvement, how about making a github issue instead of a TODO?

tools/image-builder/build_bond.zsh
```diff
@@ -0,0 +1,82 @@
+#!/bin/zsh
+
+set -eux
+setopt shwordsplit
```
tstein: Please add a comment explaining in what way/why we're depending on this behavior.

tools/image-builder/build_bond.zsh
```diff
@@ -0,0 +1,82 @@
...
+
+HOME=$1
+FLAVOR=$2
+BOOST=${3:-}
+
+BUILD_PATH=/root/build/$FLAVOR
+if [ $BOOST != "" ]; then BUILD_PATH=$BUILD_PATH/$BOOST; fi
```
chwarr: I think this is causing an (non-fatal?) error during execution of the script:

For the [C# build for this PR](https://travis-ci.org/Microsoft/bond/jobs/291902105#L633):
```
+/root/build_bond.zsh:11> [ '!=' '' ']'
/root/build_bond.zsh:11: parse error: condition expected: !=
```

tools/image-builder/build_bond.zsh
```diff
@@ -0,0 +1,82 @@
...
+        nuget restore /root/bond/cs/cs.sln
+
+        cmake -DBOND_SKIP_GBC_TESTS=TRUE -DBOND_SKIP_CORE_TESTS=TRUE -DBOND_ENABLE_GRPC=FALSE /root/bond
+
+        make gbc
+        make DESTDIR=/root install
+
+        export BOND_COMPILER_PATH=/root/usr/local/bin
+
+        msbuild /p:Configuration=Debug /root/bond/cs/cs.sln
```
chwarr: For a future commit: we should look at passing `/m` to enable parallel builds.

tools/image-builder/build_bond.zsh
tstein: `if [ "x$BOOST" != "x" ]`. shell scripting ¯\_(ツ)_/¯

tstein: Better answer: use `[[` every time.

chwarr: 🤦‍

tools/image-builder/build_bond.zsh
```diff
@@ -0,0 +1,82 @@
+#!/bin/zsh
+
+set -eux
+setopt shwordsplit
+
+HOME=$1
```
tstein: This is shadowing a common envvar. `SYMLINKED_HOME`?

tools/image-builder/build_bond.zsh
```diff
@@ -0,0 +1,82 @@
...
+FLAVOR=$2
+BOOST=${3:-}
```
tstein: Any thoughts on setting an explicit default here? i.e., `BOOST=${3:-1.63.0}`.

tools/image-builder/build_bond.zsh
ara-ayvazyan: Switched to a `/root/build` for all flavors.

tools/image-builder/build_bond.zsh
ara-ayvazyan: Updated

tools/image-builder/build_bond.zsh
ara-ayvazyan: Talked offline. Keeping the way it is.

tools/image-builder/build_bond.zsh
ara-ayvazyan: Better to eliminate the issue rather than address. the line is gone :)

tools/image-builder/build_bond.zsh
ara-ayvazyan: Fixed.

tools/image-builder/build_bond.zsh
ara-ayvazyan: Talked offline. Will not add default and instead will have explicit check for cpp.

tools/image-builder/build_bond.zsh
ara-ayvazyan: Makes sense. Updated.

PR #661, opened 2017-10-25
[gbc c++] Add missing move assignment operator

ara-ayvazyan:


compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
```diff
@@ -309,12 +309,13 @@ namespace std
             -- default OK when there are no meta fields
             implicitlyDeclared = CPP.ifndef CPP.defaultedFunctions [lt|
         // Compiler generated operator= OK
-        #{declName}& operator=(const #{declName}&) = default;|]
+        #{declName}& operator=(const #{declName}&) = default;
+        #{declName}& operator=(#{declName}&&) = default;|]
```
ara-ayvazyan: Looking at VC12 compatibility.

compiler/tests/generated/alias_key_types.h
```diff
@@ -45,9 +45,16 @@ namespace test
 #endif
         
         
-#ifndef BOND_NO_CXX11_DEFAULTED_FUNCTIONS
+#if !defined(BOND_NO_CXX11_DEFAULTED_MOVE_CTOR)
         // Compiler generated operator= OK
         foo& operator=(const foo&) = default;
+        foo& operator=(foo&&) = default;
+#else
+        foo& operator=(foo other)
```
chwarr: What is the result of `std::is_(nothrow_)move_assignable` for a type that just has this kind of assignment operator?

ara-ayvazyan: The `std::is_move_assignable` will of course be `true`, and the `std::is_nothrow_move_assignable` will be `false`. Note, that STL containers gained `noexcept` only in C++17 for their move-assignment operators.

lalo:
@ara-ayvazyan do we have tests that exercise and compile this c++ code?

Issue #660, opened 2017-10-25
Cannot serialize required binary guid using version 7 (C#)


nrandell:
I just upgraded to the latest libraries and some of my unit tests started failing around the serialization of Guid to array segments. 

I took the guid example, cloned it to (https://github.com/nrandell/bond-guid.git) and extended to have record with required and one without. The one without required works fine, the other throws a TypeInitializationException

```
Unhandled Exception: System.TypeInitializationException: The type initializer for 'Cache`2' threw an exception. ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.InvalidOperationException: variable 'convertedBlob' of type 'System.ArraySegment`1[System.Byte]' referenced from scope '', but it is not defined
   at System.Linq.Expressions.Compiler.VariableBinder.Reference(ParameterExpression node, VariableStorageKind storage)
   at System.Linq.Expressions.Compiler.VariableBinder.VisitParameter(ParameterExpression node)
   at System.Linq.Expressions.ParameterExpression.Accept(ExpressionVisitor visitor)
   at System.Linq.Expressions.Compiler.VariableBinder.Visit(Expression node)
   at System.Linq.Expressions.ExpressionVisitor.VisitBinary(BinaryExpression node)
   at System.Linq.Expressions.BinaryExpression.Accept(ExpressionVisitor visitor)
   at System.Linq.Expressions.Compiler.VariableBinder.Visit(Expression node)
   at System.Linq.Expressions.ExpressionVisitor.Visit(ReadOnlyCollection`1 nodes)
   at System.Linq.Expressions.Compiler.VariableBinder.VisitBlock(BlockExpression node)
   at System.Linq.Expressions.BlockExpression.Accept(ExpressionVisitor visitor)
   at System.Linq.Expressions.Compiler.VariableBinder.Visit(Expression node)
   at System.Linq.Expressions.ExpressionVisitor.Visit(ReadOnlyCollection`1 nodes)
   at System.Linq.Expressions.Compiler.VariableBinder.VisitBlock(BlockExpression node)
   at System.Linq.Expressions.BlockExpression.Accept(ExpressionVisitor visitor)
   at System.Linq.Expressions.Compiler.VariableBinder.Visit(Expression node)
   at System.Linq.Expressions.ExpressionVisitor.Visit(ReadOnlyCollection`1 nodes)
   at System.Linq.Expressions.Compiler.VariableBinder.VisitLambda[T](Expression`1 node)
[...]

PR #662, opened 2017-10-25
[c#] Fix alias missing convert

lalo:
- Fixes issue https://github.com/Microsoft/bond/issues/660

### DIFF
--- a/cs/src/core/expressions/ObjectParser.cs
+++ b/cs/src/core/expressions/ObjectParser.cs
@@ -134,6 +134,14 @@ namespace Bond.Expressions
             if (fieldSchemaType.IsBondStruct() || fieldSchemaType.IsBonded() || schemaField.GetModifier() != Modifier.Optional)
             {
                 cannotOmit = Expression.Constant(true);
+
+                if (fieldSchemaType.IsBondBlob())
+                {
+                    return Expression.Block(
+                        new[] { convertedBlob },
+                        Expression.Assign(convertedBlob, blob),
+                        PrunedExpression.IfThenElse(cannotOmit, processField, omitField));
+                }
             }
             else
             {
--- a/cs/src/core/expressions/ObjectParser.cs
+++ b/cs/src/core/expressions/ObjectParser.cs
@@ -140,7 +140,7 @@ namespace Bond.Expressions
                     return Expression.Block(
                         new[] { convertedBlob },
                         Expression.Assign(convertedBlob, blob),
-                        PrunedExpression.IfThenElse(cannotOmit, processField, omitField));
+                        processField);
                 }
             }
             else
--- a/cs/test/core/ConvertTests.cs
+++ b/cs/test/core/ConvertTests.cs
@@ -27,6 +27,9 @@
         [global::Bond.Id(4), global::Bond.Type(typeof(global::Bond.Tag.nullable<global::Bond.Tag.blob>))]
         public decimal? _decNullable { get; set; }
 
+        [global::Bond.Id(5), global::Bond.Type(typeof(global::Bond.Tag.blob)), global::Bond.Required]
+        public decimal _decRequired { get; set; }
+
         public Decimals()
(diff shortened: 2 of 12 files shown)


cs/src/core/expressions/ObjectParser.cs
```diff
@@ -134,6 +134,14 @@ Expression Field(ITransform transform, Expression structVar, UInt16 id, ISchemaF
             if (fieldSchemaType.IsBondStruct() || fieldSchemaType.IsBonded() || schemaField.GetModifier() != Modifier.Optional)
```
lalo: This bug also affects bonded and the other modifiers that are not optional.

chwarr: struct fields cannot be aliased, nor can bonded fields, so really is just affects blob fields that aren't optional.

cs/test/core/ConvertTests.cs
```diff
@@ -27,6 +27,9 @@ public class Decimals
         [global::Bond.Id(4), global::Bond.Type(typeof(global::Bond.Tag.nullable<global::Bond.Tag.blob>))]
```
chwarr: We should also ensure that we have test coverage for aliases required, optional, required_optional fields that use a type other than blob as their underlying type. Since blob is special cased in many places, we also want to test the non-special case code paths.

cs/test/core/ConvertTests.cs
```diff
@@ -38,11 +41,12 @@ protected Decimals(string fullName, string name)
             _decList = new LinkedList<decimal>();
             _decMap = new Dictionary<int, decimal>();
             _decNullable = null;
+            _decRequired = new decimal();
         }
 
         internal int CountInstances()
         {
-            return 1 + _decVector.Count + _decList.Count + _decMap.Count + (_decNullable.HasValue ? 1 : 0);
+            return 1 + _decVector.Count + _decList.Count + _decMap.Count + (_decNullable.HasValue ? 1 : 0) + 1;
```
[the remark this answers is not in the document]
lalo: Wanted to keep symmetry with the declarations

cs/src/core/expressions/ObjectParser.cs
```diff
@@ -134,6 +134,14 @@ Expression Field(ITransform transform, Expression structVar, UInt16 id, ISchemaF
...
             {
                 cannotOmit = Expression.Constant(true);
+
+                if (fieldSchemaType.IsBondBlob())
+                {
+                    return Expression.Block(
+                        new[] { convertedBlob },
+                        Expression.Assign(convertedBlob, blob),
+                        PrunedExpression.IfThenElse(cannotOmit, processField, omitField));
```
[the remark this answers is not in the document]
lalo: true

cs/test/core/ConvertTests.cs
```diff
@@ -30,6 +79,12 @@ public class Decimals
         [global::Bond.Id(5), global::Bond.Type(typeof(global::Bond.Tag.blob)), global::Bond.Required]
         public decimal _decRequired { get; set; }
 
+        [global::Bond.Id(6), global::Bond.Type(typeof(global::Bond.Tag.blob)), global::Bond.RequiredOptional]
+        public decimal _decRequiredOptional { get; set; }
+
+        [global::Bond.Id(7), global::Bond.Type(typeof(global::Bond.Tag.blob))]
+        public decimal _decOptional { get; set; }
```
chwarr: `_decOptional` looks to be the same as `_dec`, so one of them can be removed. Same thing for `_date` and `_ref`.

cs/test/core/ConvertTests.cs
```diff
@@ -197,8 +280,80 @@ public void Setup()
             BondTypeAliasConverter.ResetCounts();
         }
 
+        private static DateTime MakeUtcDateTime(int year, int month, int day)
```
chwarr: Minor: move this below all the test methods.

PR #663, opened 2017-10-25
[gbc] Update maintainer in bond.cabal

chwarr:


lalo:
I'll leave it open for a while in case somebody wants to comment on before merging.

chwarr:
Travis CI runs were canceled as @lalo is debugging some other failures. They will need to be re-run before merging.

PR #664, opened 2017-10-25
[gbc c++] Generate copy and move constructors with allocator

ara-ayvazyan:


CHANGELOG.md
```diff
@@ -19,6 +19,11 @@ different versioning scheme, following the Haskell community's
 * C# NuGet version: TBD
 * C# Comm NuGet version: TBD
 
+### `gbc` and Bond compiler library ###
+
+* C++ codegen now generates additional copy and move constructors with allocator
```
chwarr: Minor: consider "C++ codegen now generates copy and move constructors with an allocator argument when a custom allocator is used."

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
```diff
@@ -276,6 +278,28 @@ namespace std
                 getAllocator BT_MetaFullName =  [lt|.get_allocator()|]
                 getAllocator _ = mempty
 
+        -- copy constructor with allocator
+        allocatorCopyCtor alloc = [lt|
```
chwarr: If we combine this into one function called something like `allocatorCopyMoveCtors`, we can share a ton of the identical code and just have `copyFieldInit` and `moveFieldInit` to handle the differences between wrapping with `std::move` and not.

Perhaps something like this will help too:

```
withMove x = [lt|std::move(#{x})|] -- and for not moving, just use id
```

compiler/tests/generated/alias_with_allocator_types.h
```diff
@@ -46,6 +46,22 @@ namespace test
         foo(const foo&) = default;
 #endif
         
+        foo(const foo& other, const arena& allocator)
```
chwarr: Should these constructors be templated so that they are only expanded if they are actually used?

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs

CHANGELOG.md
ara-ayvazyan: Updated.

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
chwarr: The functions still look nearly identical to me. I was sort of expecting something more along the lines of this:

```
allocatorCtorBuilder argumentAdjuster = ...
allocatorCopyCtor = allocatorCtorBuilder id
allocatorMoveCtor = allocatorCtorBuilder withMove
    where withMove x = ...
```

ara-ayvazyan: I've not sent the updated commit yet :)

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
```diff
@@ -278,17 +278,20 @@ namespace std
                 getAllocator BT_MetaFullName =  [lt|.get_allocator()|]
                 getAllocator _ = mempty
 
-        -- copy constructor with allocator
-        allocatorCopyCtor alloc = [lt|
-        #{declName}(const #{declName}&#{otherParam}, const #{alloc}&#{allocParam})#{initList}#{ctorBody}
-        |]
+        -- copy/move constructor with allocator
+        allocatorCopyOrMoveCtor fieldDecl paramValue alloc = [lt|
```
chwarr: `fieldDecl` make me think this is handling each field one-by-one. Perhaps `paramDecl` or `otherParamDecl`? (If `otherParamDecl` is picked, then consider `otherParamValue` instead of `paramValue`.)

chwarr:
Please try to avoid rebasing during a review. It forces reviewers to re-review everything, including stuff they've already looked at, as GitHub can't handle tracking comments across rebases well.

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
```diff
@@ -118,8 +119,8 @@ namespace std
     {
         #{newlineSepEnd 2 field structFields}#{defaultCtor}
 
-        #{copyCtor}
-        #{moveCtor}
+        #{copyCtor}#{optional allocatorCopyCtor allocator}
+        #{moveCtor}#{optional allocatorMoveCtor allocator}
```
chwarr: Consider moving the `optional` into `allocatorCopyOrMoveCtor` so we do the two checks at the same place.

cpp/inc/bond/core/maybe.h
```diff
@@ -45,17 +53,23 @@ class maybe
...
     template <typename Allocator>
-    explicit maybe(Allocator& alloc)
+    explicit maybe(const Allocator& alloc)
         : _value(alloc),
           _nothing(true)
     {}
 
     template <typename Compare, typename Allocator>
-    explicit maybe(const Compare& comp, Allocator& alloc)
+    maybe(const Compare& comp, const Allocator& alloc)
```
chwarr: Consider checking where this is called to see whether we need such an overload at all.

ara-ayvazyan: Looks like we do not need this anymore since we moved to `C++11`. Will send a separate PR after I confirm it works on all supported compilers.

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs

compiler/Options.hs
```diff
@@ -95,6 +96,7 @@ cpp = Cpp
     , core_enabled = True &= explicit &= name "core" &= help "Generate core serialization definitions (true by default, --core=false to disable)"
     , comm_enabled = False &= explicit &= name "comm" &= help "Generate comm definitions"
     , grpc_enabled = False &= explicit &= name "grpc" &= help "Generate gRPC definitions"
+    , alloc_ctors_enabled = False &= explicit &= name "alloc_ctors" &= help "Generate constructors with allocator argument"
```
ara-ayvazyan: Should be `alloc-ctors` (with `-`).

compiler/src/Language/Bond/Codegen/Java/Class_java.hs
```diff
@@ -235,13 +235,16 @@ makeStructBondTypeMember_StructBondTypeBuilderImpl declName declParams = [lt|
...
-                return new StructBondTypeImpl(#{ifThenElse (null declParams) "null" "new org.bondlib.GenericTypeSpecialization(genericTypeArguments)"});
+                return new StructBondTypeImpl(#{ifThenElse (null declParams) nullStr "new org.bondlib.GenericTypeSpecialization(genericTypeArguments)"});
             }
 
             static void register() {
                 registerStructType(#{declName}.class, new StructBondTypeBuilderImpl());
             }
         }|]
+    where
+        nullStr :: Text
```
ara-ayvazyan: `nullText`

chwarr: You should be able to shorten this to one line:

```
nullText = "null" :: Text
```

compiler/src/Language/Bond/Util.hs
```diff
@@ -70,6 +71,11 @@ sepBy s f (x:xs)
 optional :: (Monoid m) => (a -> m) -> Maybe a -> m
 optional = maybe mempty
 
+-- takes a predicate and two items and outputs one based on predicate
```
chwarr: Minor: this doesn't take a predicate. That would be more like `(b -> Bool) -> a -> a`.
Maybe just "if/then/else as a function"?

ara-ayvazyan: Updated to: `if-then-else as a function`

PR #668, opened 2017-10-26
[build c++] Update build script to accept different compilers

ara-ayvazyan:


ara-ayvazyan:
I've canceled the CI build intentionally since the image is not updated.

PR #669, opened 2017-10-26
[c++] Fix broken build on gcc

ara-ayvazyan:


cmake/Config.cmake
```diff
@@ -169,6 +169,7 @@ cxx_add_compile_options(GNU
     --std=c++11
     -Wall
     -Werror
+    -Wno-deprecated-declarations
```
chwarr: What are we using that's deprecated?

ara-ayvazyan: Boost uses deprecated `std::auto_ptr`. Will add a comment here.

PR #670, opened 2017-10-26
[build c#] Enable parallel build

ara-ayvazyan:


tools/image-builder/build_bond.zsh
```diff
@@ -51,8 +51,8 @@ case "$FLAVOR" in
         make gbc
         make install
 
-        msbuild /p:Configuration=Debug /root/bond/cs/cs.sln
-        msbuild /p:Configuration=Fields /root/bond/cs/cs.sln
+        msbuild /p:Configuration=Debug /root/bond/cs/cs.sln /m
```
chwarr: Minor: let's move the arguments to before the solution: `msbuild /p:... /m /root/bond/cs/cs.sln`

ara-ayvazyan: Updated.

PR #671, opened 2017-10-27
merge jvm into master and build Java CI in the container

tstein:


chwarr:
The build appears to be failing. I think this is because the build_bond.zsh is built-in to the Docker container. To get this to pass, either we need to

1. update the container with these changes before merging or
1. teach the container build to pull the build script from the repository and not be baked into the container.

Option 2 seems cleaner to me.

chwarr:
Minor: can we reword the message for commit 7dd04d4? It just says "tmp" right now.

tstein:
This is very much WIP. Sorry for not marking it. I'll update when it's ready to look at.

tstein:
Yes, the `build_bond.zsh` that's baked into the container doesn't know about Java. The direction I'd like to go is stripping it down until it's just a dispatcher for the copy of `build_$FLAVOR` in the wokring tree, which is how Java is intended to work in this commit. From that perspective, I'd like to just update the container this time, and then quickly move to a place where we're only touching `build_bond.zsh` when we're doing something as monumental as adding a language.

ara-ayvazyan:
#679

tstein:
Requires #682.

tstein:
Is already rebased on #679, assuming it goes in unmodified.

PR #673, opened 2017-10-27
[gbc] Use latest stack

ara-ayvazyan:


ara-ayvazyan:
I've canceled the CI builds intentionally since the image is not updated.

ara-ayvazyan:
Looks like this did not actually fix the issue 😞

compiler/stack_build.cmake
```diff
@@ -11,7 +11,7 @@
 cmake_policy (SET CMP0012 NEW)
 
 execute_process (
-    COMMAND ${STACK_EXECUTABLE} setup
+    COMMAND ${STACK_EXECUTABLE} setup --allow-different-user
```
chwarr: It looks to me that this is solving a problem specific to our use of Travis CI + a Docket container. It doesn't seem appropriate to pass `--allow-different-user` unconditionally here: for developer/customer machines that isn't the right default.

Is there a way we can fix this just from the Travis build script?

ara-ayvazyan: That flag is no more needed.

ara-ayvazyan: I've sent separate PRs #681 and #682.

PR #674, opened 2017-10-27
[c++] Cleanup C++11 feature checks

ara-ayvazyan:


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ different versioning scheme, following the Haskell community's
 
 * C++ codegen now properly generates move assignment operator which was broken
   for some cases.
+* C++ codegen no more generates checks for C++11, except for MSVC 2013 workarounds.
```
chwarr: Minor: "...no longer generates"

compiler/src/Language/Bond/Codegen/Cpp/Util.hs
```diff
@@ -16,9 +16,6 @@ module Language.Bond.Codegen.Cpp.Util
     , attributeInit
     , schemaMetadata
     , ifndef
-    , defaultedFunctions
```
chwarr: This is a small breaking change (removal of exported definitions), so instead of the version change to 0.10.2.0, we need to go to 0.11.0.0.

cpp/inc/bond/core/config.h
```diff
@@ -5,63 +5,10 @@
...
-#if defined(BOOST_NO_CXX11_DEFAULTED_FUNCTIONS) \
-    || defined(BOOST_NO_CXX11_RVALUE_REFERENCES) \
-    || defined(BOOST_NO_RVALUE_REFERENCES)
-// We need support for both defaulted functions and rvalue references to use
-// default move ctors
-#define BOND_NO_CXX11_DEFAULTED_MOVE_CTOR
-#else
-    #if defined(_MSC_VER) && (_MSC_VER < 1900)
-    // Versions of MSVC prior to 1900 do not support = default for move ctors
-    #define BOND_NO_CXX11_DEFAULTED_MOVE_CTOR
```
chwarr: The code may be cleaner if we leave BOND_NO_CXX11_DEFAULTED_MOVE_CTOR in config.h instead of inlining it the generated code. Try it and see. I'll defer to your judgment here.

ara-ayvazyan: I agree that it will be cleaner. I hope that going forward we will have less such checks and not more (this change removes most). Also, one argument against that is that it makes `gbc` more coupled with core library code. I'll probable choose to go this way given that there is only one such check left in `gbc`.

ara-ayvazyan: Looks like `ifndef` can also be removed.

compiler/src/Language/Bond/Codegen/Cpp/Util.hs
ara-ayvazyan: Fixed.

CHANGELOG.md
ara-ayvazyan: Thanks.

PR #675, opened 2017-10-27
[c++ grpc] Fix multi-slice deser. memory leak

chwarr:
The caller of grpc_byte_buffer_reader_next is responsible for calling
grpc_slice_unref on the slice. However, SerializationTraits::Deserialize
was only calling grpc_slice_unref on the last slice. Now, it is called
on each slice.

Also fixes a potential bug that would call grpc_slice_unref on an
uninitialized slice.

### DIFF
--- a/cpp/inc/bond/ext/grpc/bond_utils.h
+++ b/cpp/inc/bond/ext/grpc/bond_utils.h
@@ -81,8 +81,7 @@ namespace grpc {
 
             char* dest = buff.get();
 
-            grpc_slice s;
-            while (grpc_byte_buffer_reader_next(&reader, &s) != 0)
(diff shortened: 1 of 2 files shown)


cpp/inc/bond/ext/grpc/bond_utils.h
```diff
@@ -86,11 +86,11 @@ namespace grpc {
             {
```
ara-ayvazyan: Change to: `for (grpc_slice s; grpc_byte_buffer_reader_next(&reader, &s) != 0; )`

chwarr:
Since the CI systems are backed up, I ran builds and tests locally. They're passing.

Issue #666, opened 2017-10-26
Consider using scheduled builds to reducing daily build times and improving coverage


ara-ayvazyan:
We can consider building less and more important configurations for pull requests and have scheduled builds for covering the rest. For example, we are not building C++ version with `gcc` (only `clang` is used), while we are covering 8 versions of Boost.
- [ ] [Scheduled builds](https://www.appveyor.com/docs/build-configuration/#scheduled-builds) for appveyor CI.
- [ ] [Cron jobs](https://docs.travis-ci.com/user/cron-jobs/) for Travis CI.

PR #677, opened 2017-10-27
[travis c++] Compile with Boost 1.65.1 and GCC

ara-ayvazyan:


chwarr:
In the commit message, we'll want to talk about the plan to only build some representative configurations on each PR/commit and that the fuller matrix will be built with the cron jobs you suggested in issue https://github.com/Microsoft/bond/issues/666.

ara-ayvazyan:
@tstein, I agree, but this will require changes in `build_bond.zsh` so that it accepts named parameters. Will keep this way for now, and I've added #678 to address it separately.

PR #679, opened 2017-10-28
[build] Use build_bond.zsh from sources

ara-ayvazyan:


ara-ayvazyan:
The `tools/ci-scripts` does not reflect the linux vs windows, while `build_bond.zsh` is specifically for linux. How about `tools/ci-scripts/linux/build.zsh`?

PR #681, opened 2017-10-29
[build c++] Fix how ccache is used

ara-ayvazyan:


tools/image-builder/build_bond.zsh
```diff
@@ -11,18 +9,22 @@ COMPILER=${4:-clang}
 
 case "$COMPILER" in
     clang)
-        CXX_COMPILER="clang++ -Qunused-arguments --system-header-prefix=boost/"
-        CC_COMPILER="clang -Qunused-arguments --system-header-prefix=boost/"
+        CXX_COMPILER=clang++
+        CXX_FLAGS="-Qunused-arguments --system-header-prefix=boost/"
+        CC_COMPILER=clang
+        CC_FLAGS="-Qunused-arguments --system-header-prefix=boost/"
```
tstein: If these variables were exported and named `CXX/CXXFLAGS/CC/CFLAGS`, they would affect all subprocesses and would not need to be explicitly passed to `cmake` as `-DCMAKE_C{,XX}_FLAGS`. Because this script's entire job is to setup and execute a build, I'm more inclined to set and export the standard vars here than have the extra logic. Is there a reason you wanted it like this?

ara-ayvazyan:
@chwarr, primarily yes, but seems like this is generally better way of handing `cache` from what I can understand.

PR #682, opened 2017-10-29
[build gbc] Add custom options for Stack

ara-ayvazyan:


compiler/stack_build.cmake
```diff
@@ -24,7 +24,7 @@ endif()
 set (buildGhcOptions "-O2")
 
 execute_process (
-    COMMAND ${STACK_EXECUTABLE} build :${target} --no-run-tests ${stack_options} --ghc-options=${buildGhcOptions}
+    COMMAND ${STACK_EXECUTABLE} build :${target} --no-run-tests ${stack_options} --ghc-options=${buildGhcOptions} ${BOND_STACK_OPTIONS}
```
chwarr: Neither `stack build` nor `stack path` take `--allow-different-user`, though neither seems to complain if `--allow-different-user` is passed to it.

compiler/stack_build.cmake
```diff
@@ -11,7 +11,7 @@
 cmake_policy (SET CMP0012 NEW)
 
 execute_process (
-    COMMAND ${STACK_EXECUTABLE} setup
+    COMMAND ${STACK_EXECUTABLE} setup ${BOND_STACK_OPTIONS}
```
chwarr: `stack` has two different kinds of options: those that go between `stack` and the command (like `build` or `setup`) and those that go with a corresponding command (like `--allow-different-user` that the `setup` command [among others?] takes). This appears to only be for command-specific options, but is used across many different commands.

Which stack commands need `--allow-different-user`?

ara-ayvazyan: I think `--allow-different-user` is not specific to `setup`. I can see it in `stack --help`.

chwarr: Yup. I misread the help. This should then be:

```
COMMAND ${STACK_EXECUTABLE} ${BOND_STACK_OPTIONS} setup
```

compiler/stack_build.cmake
ara-ayvazyan: Both need it, because that is not a command-specific flag.

compiler/stack_build.cmake
ara-ayvazyan: Updated.

tstein:
Last travis build job timed out at the last second. Merging.

PR #684, opened 2017-10-31
[build] Fix broken build on gcc

ara-ayvazyan:
This fixes the bug introduced in #681.

### DIFF
--- a/tools/image-builder/build_bond.zsh
+++ b/tools/image-builder/build_bond.zsh
@@ -17,7 +17,9 @@ case "$COMPILER" in
     
     gcc)
         CXX_COMPILER=g++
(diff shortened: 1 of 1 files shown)


tools/image-builder/build_bond.zsh
```diff
@@ -17,7 +17,9 @@ case "$COMPILER" in
     
     gcc)
         CXX_COMPILER=g++
+        CXX_FLAGS=
         CC_COMPILER=gcc
+        CC_FLAGS=
```
chwarr: Complaining about undefined variables?

ara-ayvazyan: Yep.

PR #688, opened 2017-11-01
[travis] Restore ownership before caching

lalo:


lalo:
gcc build timed-out, which has happened on multiple branches including master branch

PR #689, opened 2017-11-02
[c++] Cleanup c++11 allocator traits checks

ara-ayvazyan:


ara-ayvazyan:
Judging based on the CI builds, boost 1.58 is also fine. We are using g++ 5.4 which [looks fine](https://travis-ci.org/Microsoft/bond/jobs/296076187) (failure is due to a timeout).

Issue #691, opened 2017-11-02
Bond build failed with error MSB6006: "cmd.exe" exited with code 9009


shanshan0309:
I build Bond with Visual studio 2017 compiler on Windows Server 2016, but it failed with error MSB6006: "cmd.exe" exited with code 9009. This issue can be reproduced from master revision [d157d75](https://github.com/Microsoft/bond/commit/d157d7598c2bb9c1d0a91e9d5c71f9508041916b). Could you help have a look about this issue? Thanks in advance!

**Failures:**
CustomBuild:
  Building Custom Rule D:/Bond/src/java/gradle-plugin/CMakeLists.txt
  CMake does not need to re-run because D:/Bond/build_x64/java/gradle-plugin/CMakeFiles/generate.stamp is up-to-date.
  Generating build/java/gradle-plugin
  'GRADLE_EXECUTABLE-NOTFOUND' is not recognized as an internal or external command,
  operable program or batch file.
C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\VC\VCTargets\Microsoft.CppCommon.targets(171,5): error MSB6006: "cmd.exe" exited with code 9009. [D:\Bond\build_x64\java\gradle-plugin\gradle-plugin.vcxproj]

**Repro steps:**
1. Open VS2017 x64 Native tools command tool
2. git clone --recursive https://github.com/Microsoft/bond.git D:\Bond\src
3. cd D:\Bond
4. mkdir build_x64 && pushd build_x64
5. cmake -G "Visual Studio 15 Win64" -DCMAKE_SYSTEM_VERSION=10.0.15063.0 ..\src
6. msbuild bond.sln /p:Configuration=Release;Platform=x64 /t:Rebuild

ara-ayvazyan:
Looks like the readme is missing instructions on how to setup for building Java (#692). If you do not need Java, then can you try to unload related projects from the solution (#693)?

tstein:
With the above change, Java targets won't be created unless you explicitly ask for them when you run the `cmake` configure step (with `-G`). It may be enough to re-run `cmake -G` to fix the issue in your existing build, but I recommend that you remove your entire `D:\Bond\build_x64` and start fresh. Please let us know if the problem isn't resolved.

shanshan0309:
@tstein, thank you so much for fixing this issue. I've verified your fix works well.

PR #694, opened 2017-11-02
cmake: Add BOND_ENABLE_JAVA (default=False).

tstein:
Intended to resolve https://github.com/Microsoft/bond/issues/691.


cmake/Config.cmake
```diff
@@ -209,6 +209,10 @@ set (BOND_ENABLE_COMM
     "FALSE"
     CACHE BOOL "If FALSE, then do not build Comm")
 
+set (BOND_ENABLE_JAVA
```
chwarr: The `GRADLE_FOUND` and `MAVEN_FOUND` variables in FindJavaBuildTools.cmake don't appear to be used. Can they be removed?

chwarr: Can we make FindJavaBuildTools.cmake fail now, as they'll only be invoked if BOND_ENABLE_JAVA is true?

tstein: Done.

tstein: Done.

PR #695, opened 2017-11-03
[gbc] Add unit tests that covers negative values

lalo:
- Missing unit tests that cover negative hex and oct values are added


compiler/tests/generated/allocator/defaults_reflection.h
```diff
@@ -413,47 +415,69 @@ namespace tests
...
-        private: typedef boost::mpl::push_front<fields27, var::m_int16_5>::type fields28;
-        private: typedef boost::mpl::push_front<fields28, var::m_int16_4>::type fields29;
-        private: typedef boost::mpl::push_front<fields29, var::m_int8_5>::type fields30;
-        private: typedef boost::mpl::push_front<fields30, var::m_int8_4>::type fields31;
-        private: typedef boost::mpl::push_front<fields31, var::m_str_2>::type fields32;
-        private: typedef boost::mpl::push_front<fields32, var::m_str_1>::type fields33;
-        private: typedef boost::mpl::push_front<fields33, var::m_bool_3>::type fields34;
-        private: typedef boost::mpl::push_front<fields34, var::m_bool_2>::type fields35;
-        private: typedef boost::mpl::push_front<fields35, var::m_bool_1>::type fields36;
+        private: typedef boost::mpl::push_front<fields0, var::m_int64_neg_oct>::type fields1;
```
lalo: I don't get it why they were pushed in front of this list

ara-ayvazyan: Because `boost::mpl::push_front` is used which is faster (to compile/run inside compiler) than `push_back`.

lalo: Thanks, sorry I deleted the comment right before you posted. For transparency, I asked that I did not understand why it was added at the top.

PR #697, opened 2017-11-03
[travis] Cache warm-up stage

ara-ayvazyan:


tools/ci-scripts/linux/build_travis.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -eux
+
+pushd $HOME/docker-images
+wget -N https://bondbinaries.blob.core.windows.net/bondbinaries/docker-images/bond-xenial.tar.gz
+popd
+
+time docker load -i $HOME/docker-images/bond-xenial.tar.gz
+docker run -v $HOME/.ccache:/root/.ccache -v $HOME/.stack:/root/.stack -v `pwd`:/root/bond bond-xenial $HOME "$@"
```
ara-ayvazyan: We should mount a sub-folder in `.ccache` per flavor to avoid corruptions.

lalo:
This will increase over all time right? Everything blocks until warm up is finished?

PR #699, opened 2017-11-03
[gbc c++] Generate type aliases

### LINKED ISSUE #685: [gbc c++] Generate type aliases for the ones in IDL

ara-ayvazyan:
Resolves #685.


compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
```diff
@@ -30,8 +30,9 @@ types_h :: [String]     -- ^ list of optional header files to be @#include@'ed b
         -> Bool         -- ^ 'True' to generate enum definitions into a separate file /base_name/_enum.h
         -> Maybe String -- ^ optional custom allocator to be used in the generated code
         -> Bool         -- ^ 'True' to generate constructors with allocator
+        -> Bool         -- ^ 'True' to generate type aliases
```
lalo: I believe this is an API change and we have to bump gbc library

compiler/tests/Tests/Codegen.hs
```diff
@@ -96,7 +96,7 @@ verifyCppCommCodegen args baseName =
 verifyCppGrpcCodegen :: [String] -> FilePath -> TestTree
 verifyCppGrpcCodegen args baseName =
     testGroup baseName $
-        map (verifyFile options baseName cppTypeMapping "") templates
+        map (verifyFile options baseName (cppExpandAliases (type_aliases_enabled options) cppTypeMapping) "") templates
```
lalo: No test is currently exercising this in grpc or comm. I don't know if we need to.

lalo: We also need a test that actually compiles and exercises the cpp changes.

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
ara-ayvazyan: There is already `(minor bump already done in bond.cabal)` note in changelog.

compiler/tests/generated/type_aliases/aliases_types.h
```diff
@@ -0,0 +1,264 @@
...
+#endif
+
+#include <bond/core/config.h>
+#include <bond/core/containers.h>
+
+
+
+namespace tests
+{
+    template <typename N, typename T>
```
ara-ayvazyan: Looks like we have a problem here. We cannot use `typename` for `N`.

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
chwarr: This is a breaking change unless you also add a function that takes one fewer parameter and delegates to this one. (A breaking change in the Bond Haskell library is fine for a change like this.)

ara-ayvazyan: Looks like that is already done. I'll fix the changelog note to `major bump`.

compiler/tests/Tests/Codegen.hs
ara-ayvazyan: Will update existing examples that use aliases.

PR #700, opened 2017-11-03
[c++ grpc] Upgrade to gRPC v1.7.1

chwarr:
Pick up a bug fix in commit gRPC 7269667f (initially implemented for C++
in gRPC commit be9b8142) for channel disconnections we've been
observing.


chwarr:
I've tested this change against our release process, and it's working with gRPC v1.7.1. When this passes CI, it's ready to be squash merged.

PR #703, opened 2017-11-07
[gbc c++] Add option to use std::scoped_allocator_adaptor

ara-ayvazyan:


compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
```diff
@@ -86,11 +87,16 @@ types_h userHeaders enumHeader allocator alloc_ctors_enabled cpp file imports de
 
     anyNullable = Any . isNullable
 
+    anyStringOrContainer t
```
chwarr: Consider `anyStringOrContainer f = Any (isString f || isMetaName f || isContainer f)`

CHANGELOG.md
```diff
@@ -28,6 +28,9 @@ different versioning scheme, following the Haskell community's
   accepting `Comparer` for associative containers.
 * C++ codegen now can generate copy and move constructors with an allocator argument
   when a custom allocator is used and `--alloc-ctors` is passed to `gbc`.
+* C++ codegen now can use [`std::scoped_allocator_adaptor`](http://en.cppreference.com/w/cpp/memory/scoped_allocator_adaptor)
+  for strings and containers when custom allocator is used and `--scoped-alloc` flag
```
chwarr: Are sub-structs intentionally excluded?

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
```diff
@@ -86,11 +87,16 @@ types_h userHeaders enumHeader allocator alloc_ctors_enabled cpp file imports de
...
+        | isString t || isMetaName t || isContainer t = Any True
+        | otherwise = Any False
+
     bondHeaders :: [(Bool, String)]
     bondHeaders = [
         (have anyNullable, "<bond/core/nullable.h>"),
         (have anyBonded, "<bond/core/bonded.h>"),
-        (have anyBlob, "<bond/core/blob.h>")]
+        (have anyBlob, "<bond/core/blob.h>"),
+        (scoped_alloc_enabled && have anyStringOrContainer, "<scoped_allocator>")]
```
chwarr: What about nullable/maybe fields? Should they use the scoped allocator or not?

ara-ayvazyan: The `bond::maybe` is forwarding the allocator to its element and does not directly use it, so no changes there. The `bond::nullable` however uses it in some cases, but since `std::scoped_allocator_adaptor` is compatible with its outer one, no changes there too.

CHANGELOG.md
ara-ayvazyan: Discussed offline.

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs

CHANGELOG.md
chwarr: To summarize the discussion: the sub-structs will get the top-level allocator and will use that to create a `std::scoped_allocator_adapter` around the top-level allocator for their own fields. The intent of this functionality is to make things like `someStruct.collectionField.emplace_back(arg1, arg2)` transparently use the allocator, less so to use the levels functionality of `std::scope_allocator_adapter`.

PR #707, opened 2017-11-09
[c++] Use Kind enum in polymorph. cont. visitor

chwarr:
Switch the polymorphic_container_visitor example to use a Kind enum
instead of the deprecated bond_meta::full_name.

TODO:
* [x] Build and test with G++ and Clang++
* [x] Wait for MSVC 2013 CI builds and fix and problems found in them.

### DIFF
--- a/examples/cpp/core/polymorphic_container_visitor/polymorphic_container_visitor.bond
+++ b/examples/cpp/core/polymorphic_container_visitor/polymorphic_container_visitor.bond
@@ -1,8 +1,15 @@
 namespace examples.polymorphic_containers
 
+enum StructKind
+{
+    StructKind_Unknown;
+    StructKind_Struct1;
+    StructKind_Struct2;
+}
+
 struct Base
 {
-    0: bond_meta::full_name name;
+    0: StructKind kind = StructKind_Unknown;
 }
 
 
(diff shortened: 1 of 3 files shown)


examples/cpp/core/polymorphic_container_visitor/polymorphic_container_visitor.bond
```diff
@@ -1,8 +1,15 @@
 namespace examples.polymorphic_containers
 
+enum StructKind
+{
+    StructKind_Unknown;
```
ara-ayvazyan: Any reason for using `StructKind_` with `c++11`?

chwarr: Bond doesn't yet generate enum classes, so `Circle` would be ambiguous between the enum member and the struct.

ara-ayvazyan: I thought even old one used to generate in a way that `Enum::Value` works using a `namespace` trick?

chwarr: `Enum::Value` works, but just `Value` is also a visible name. Since--in this case--there is also a struct called `Value` that is visible, just `Value` is ambiguous.

ara-ayvazyan: Ah, OK.

PR #709, opened 2017-11-10
[c++] Add detail::SkipElements to remove duplicated code

ara-ayvazyan:


cpp/inc/bond/core/detail/typeid_value.h
```diff
@@ -203,6 +203,20 @@ inline void BasicTypeContainer(T& var, BondDataType type, Reader& input, uint32_
 }
 
 
+template <typename E>
+inline void SkipElements(const E& element, uint32_t size)
+{
+    while (size--)
```
chwarr: Minor: add { and } since we now only have a handful of copies of this function?

PR #711, opened 2017-11-10
[gbc c++] Move ToEnum and FromEnum to .cpp

ara-ayvazyan:


chwarr:
Something funky is going on with Travis. If this hasn't resolved by Monday, we'll need to figure out the problem...

chwarr:
Travis problems cleared up.

PR #712, opened 2017-11-11
[gbc c++] Use init. list for populating enum maps

ara-ayvazyan:
Currently, `boost::assign::map_list_of` is used to initialize C++ enum maps. This change makes use of [list initialization](http://en.cppreference.com/w/cpp/language/list_initialization) available in `c++11` which is more appropriate and avoids extra dependency on template-heavy boost headers.

### DIFF
--- a/compiler/src/Language/Bond/Codegen/Cpp/Types_cpp.hs
+++ b/compiler/src/Language/Bond/Codegen/Cpp/Types_cpp.hs
@@ -8,8 +8,10 @@ module Language.Bond.Codegen.Cpp.Types_cpp (types_cpp) where
 import Data.Monoid
 import Prelude
 import Data.Text.Lazy (Text)
+import Data.List
 import Text.Shakespeare.Text
 import Language.Bond.Syntax.Types
+import Language.Bond.Syntax.Util
 import Language.Bond.Codegen.TypeMapping
 import Language.Bond.Codegen.Util
 import qualified Language.Bond.Codegen.Cpp.Util as CPP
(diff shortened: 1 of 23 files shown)


compiler/tests/generated/alloc_ctors/attributes_types.cpp
```diff
@@ -10,9 +10,10 @@ namespace tests
     namespace Enum
     {
         const
-        std::map<std::string, enum Enum> _name_to_value_Enum =
-            boost::assign::map_list_of<std::string, enum Enum>
-                ("Value1", Value1);
+        std::map<std::string, enum Enum> _name_to_value_Enum
+            {
+                { "Value1", Value1 }
```
chwarr: Inserting into a map is going to be a bit faster if the keys are already in order. Can we sort them in Haskell?

compiler/tests/generated/alloc_ctors/attributes_types.cpp
```diff
@@ -10,9 +10,10 @@ namespace tests
...
+            };
 
         const
         std::map<enum Enum, std::string> _value_to_name_Enum =
```
chwarr: Can we generated the reverse map in Haskell as well?

ara-ayvazyan:
The test failure is in unrelated grpc test.

compiler/src/Language/Bond/Syntax/Util.hs
```diff
@@ -231,3 +232,12 @@ resolveAlias Alias {..} args = fmapType resolveParam aliasType
     paramsArgs = zip declParams args
 resolveAlias _ _ = error "resolveAlias: impossible happened."
 
+
+
+-- | Fill in values for constants w/o explicitly specified value
+constNameValues :: [Constant] -> [(String, Int)]
+constNameValues constants = nameValues 0 constants
```
chwarr: `reifyEnumValues` ?

compiler/src/Language/Bond/Codegen/Cpp/Types_cpp.hs
```diff
@@ -89,9 +91,13 @@ types_cpp cpp file _imports declarations = ("_types.cpp", [lt|
...
             return true;
         }
+
     } // namespace #{declName}
     } // namespace _bond_enumerators|]
       where
-        constant Constant {..} = [lt|{ "#{constantName}", #{constantName} }|]
+        nameValueConst Constant {..} = [lt|{ "#{constantName}", #{constantName} }|]
+        valueNameConst (name, _) = [lt|{ #{name}, "#{name}" }|]
+        enumConstByName = sortBy (\x y -> constantName x `compare` constantName y) enumConstants
```
chwarr: Does `sortBy (comparing constantName) enumConstants` or `sortOn constantName enumConstants` work?

compiler/src/Language/Bond/Codegen/Cpp/Types_cpp.hs
```diff
@@ -89,9 +91,13 @@ types_cpp cpp file _imports declarations = ("_types.cpp", [lt|
...
+        enumConstByValue = sortBy (\x y -> snd x `compare` snd y) $ constNameValues enumConstants
```
chwarr: Similarly, does `sortBy (comparing snd) $ constNameValues enumConstants` or `sortOn snd $ constNameValues enumConstants` work?

compiler/src/Language/Bond/Syntax/Util.hs
ara-ayvazyan: Updated.

PR #713, opened 2017-11-11
[c#] Add cloning to expressions test

chwarr:


cs/test/expressions/Clone.expressions
```diff
@@ -0,0 +1,817 @@
+.Lambda #Lambda1<System.Func`2[System.Object,System.Object]>(System.Object $obj) {
+    .Block(ExpressionsTest.Example $Example) {
+        .Block() {
+            $Example = .New ExpressionsTest.Example();
+            .Block(ExpressionsTest.Example $Example) {
+                $Example = (ExpressionsTest.Example)$obj;
```
chwarr: Because of the names we've given them, the source and destination both get called "Example" here. Let me see whether I can fix that easily.

chwarr: Done.

chwarr:
Right, these aren't golden tests yet. Opened https://github.com/Microsoft/bond/issues/719 for that.

chwarr:
Build failure looks like a network error installing haskell-stack. Rebuilding.

PR #714, opened 2017-11-13
[gbc c++] More flexible codegen for enums

ara-ayvazyan:
Currently C++ enums are generated using `std::map` and are always constructed, even when not used. This change will:
- allow the user to provide custom map type (e.g. `boost::container::flat_map`)
- lazily construct the map on first access
- change the default generated conversion functions to use `std::unordered_map`


compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
```diff
@@ -386,30 +387,33 @@ namespace std
     namespace #{declName}
     {
         #{enumDefinition}
-        extern const std::map<enum #{declName}, std::string> _value_to_name_#{declName};
-        extern const std::map<std::string, enum #{declName}> _name_to_value_#{declName};
-
-        inline
-        const char* GetTypeName(enum #{declName})
+        inline const char* GetTypeName(enum #{declName})
```
ara-ayvazyan: Can be marked as `BOND_CONSTEXPR`.

ara-ayvazyan:
Magic statics are not supported on VC12 😞

compiler/src/Language/Bond/Codegen/Cpp/Types_cpp.hs
```diff
@@ -20,16 +21,18 @@ types_cpp :: MappingContext -> String -> [Import] -> [Declaration] -> (String, T
...
 #{CPP.closeNamespace cpp}
 |])
   where
+    unorderedMapInclude = [lt|#include <unordered_map>
+|]
+
     -- definitions of Schema statics for non-generic structs
     statics s@Struct {..} =
-        if null declParams then CPP.schemaMetadata cpp s else mempty
+        ifThenElse (null declParams) (CPP.schemaMetadata cpp s) mempty
```
chwarr: I find it clearer with the `if ... then ... else` like before. We don't need a function call here, so the expression form is fine.

compiler/src/Language/Bond/Codegen/Cpp/Types_cpp.hs
```diff
@@ -99,3 +102,7 @@ types_cpp cpp file _imports declarations = ("_types.cpp", [lt|
     } // namespace _bond_enumerators|]
 
     statics _ = mempty
+
+    anyEnum [] = False
```
chwarr: There's [`isEnum`](https://github.com/Microsoft/bond/blob/939c298ff6d619dbc0bc9f5ca25c2088de86d83b/compiler/src/Language/Bond/Syntax/Util.hs#L132-L135) in Utils.hs. It handles aliases as well.

With that and [`any`](http://hackage.haskell.org/package/base-4.10.0.0/docs/Prelude.html#v:any)
```
anyEnum :: [Type] -> Bool
anyEnum = any isEnum
```
or, alternatively if you just need a value
```
hasEnumDecl :: Bool
hasEnumDecl = any isEnum declarations
```
or just do it all together:

```
 unorderedMapInclude = if (not (any isEnum declarations)) then mempty else [lt|#include <unordered_map>
|lt]
```

compiler/src/Language/Bond/Codegen/Cpp/Types_cpp.hs
ara-ayvazyan: Reverted.

compiler/src/Language/Bond/Codegen/Cpp/Types_cpp.hs
ara-ayvazyan: Here we care about declarations and not types, so cannot use `isEnum` you mentioned and we don't care about aliases. I added `isEnumDelcaration` and will use `any` (thanks!).

ara-ayvazyan:
@chwarr, will need to discuss a bit more on this, so please hold on merging.

CHANGELOG.md
```diff
@@ -33,6 +33,8 @@ different versioning scheme, following the Haskell community's
 * C++ codegen now can use [`std::scoped_allocator_adaptor`](http://en.cppreference.com/w/cpp/memory/scoped_allocator_adaptor)
   for strings and containers when custom allocator is used and `--scoped-alloc` flag
   is passed to `gbc`.
+* C++ codegen now generates lazily constructible enum maps and accepts user-defined
```
chwarr: Minor: how about "C++ codegen now generated lazily constructed enum name-to-value and value-to-name maps. Additionally, a user-defined map type can now be provided to `GetNameToValueMap` and `GetValueToNameMap`."?

The unordered_map change is an implementation detail, so it can be omitted from the changelog. If you want to talk about it, maybe something like "Enum name-to-value and value-to-name look ups now work in amortized constant time."?

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
```diff
@@ -386,33 +386,51 @@ namespace std
...
+
+        inline const std::map<std::string, enum #{declName}>& GetNameToValueMap(enum #{declName})
         {
             return _name_to_value_#{declName};
         }
+#else
+        template <typename Map = std::map<enum #{declName}, std::string> >
+        inline const Map& GetValueToNameMap(enum #{declName}, ::bond::detail::mpl::identity<Map> = {})
+        {
+            static const Map _map_#{declName}
```
chwarr: Minor: can't we just use "map" or "valueNameMap" here? There are non members to conflict with, nor any named parameters.

CHANGELOG.md
ara-ayvazyan: Updated.

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs

PR #715, opened 2017-11-14
[c++] Reduce included headers for JSON

ara-ayvazyan:
This change introduces a `BOND_LIB_TYPE` macro and moves some utility functions used by JSON protocol to corresponding source file which reduces the number of included files by over 240.

The `BOND_LIB_TYPE` macro will allow:
- to keep some expensive-to-compile headers away from public ones
- pre-compile some common parts of the library


ara-ayvazyan:
@tstein, the macro is for consumers who need to consume Bond as a [header-only](https://en.wikipedia.org/wiki/Header-only) (it does not imply a single, fat source file) library. I could not come up with simpler and more reflective name for it. Please suggest a better one if you have in mind.

tstein:
I couldn't think of one. :) That's about what I thought it implied, though, so let's do it.

cpp/inc/bond/core/config.h
```diff
@@ -40,3 +40,9 @@
 #define BOND_CONSTEXPR          BOOST_CONSTEXPR
 #define BOND_CONSTEXPR_OR_CONST BOOST_CONSTEXPR_OR_CONST
 #define BOND_STATIC_CONSTEXPR   BOOST_STATIC_CONSTEXPR
+
+#ifdef BOND_HEADER_ONLY
+#define BOND_HEADER_ONLY_INLINE inline
```
chwarr: `BOND_HEADER_ONLY` is intended to be a public-facing macro, but `BOND_HEADER_ONLY_INLINE` is a helper for us. Consider `BOND_DETAIL_HEADER_ONLY_INLINE`?

cpp/CMakeLists.txt
```diff
@@ -140,6 +144,7 @@ add_library (bond
     STATIC
     ${schemas}
     ${generated_files_types}
+    ${precompiled_sources}
```
chwarr: The new precompiled sources will also need to be installed when BOND_LIBRARIES_INSTALL_CPP is defined. (This is used when we produce C++ packages for some odd build systems used internally at Microsoft.)

The description of the CMake variable `BOND_LIBRARIES_INSTALL_CPP` in Config.cmake also needs to be updated:

> If TRUE, the .cpp files for the Bond libraries will be installed under src/ as part of the INSTALL target.

cpp/inc/bond/core/config.h
```diff
@@ -40,3 +40,9 @@
 #define BOND_CONSTEXPR          BOOST_CONSTEXPR
 #define BOND_CONSTEXPR_OR_CONST BOOST_CONSTEXPR_OR_CONST
 #define BOND_STATIC_CONSTEXPR   BOOST_STATIC_CONSTEXPR
+
+#ifdef BOND_HEADER_ONLY
```
chwarr: We should augment our documentation about [optimizing C++ build time](https://microsoft.github.io/bond/manual/bond_cpp.html#optimizing-build-time) (or similar) to mention this new macro.

cpp/inc/bond/core/config.h
ara-ayvazyan: Agree, but none of the rest is for public use either I suppose? :)

chwarr: Opened https://github.com/Microsoft/bond/issues/721 for this. I found one other public one, `BOND_DEFINE_BUFFER_MAGIC` and I could be convinced that `BOND_VERSION` and `BOND_MIN_CODEGEN_VERSION` are public.

cpp/inc/bond/core/config.h
chwarr: The default for this needs to be header-only, to avoid breaking people who aren't linking with libbond. Also, consider something along the lines of

```
#define BOND_LIB_TYPE_HEADER 1
#define BOND_LIB_TYPE_STATIC 2
// #define BOND_LIB_TYPE_DYNAMIC 3 // not yet implemented
// #define BOND_LIB_TYPE_DYNAMIC 4 // not yet implemented

#ifndef BOND_LIB_TYPE
#define BOND_LIB_TYPE BOND_LIB_TYPE_HEADER
#endif
```

cpp/inc/bond/protocol/detail/rapidjson_utils_impl.h
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#pragma once
+
+#include <bond/protocol/detail/rapidjson_utils.h>
```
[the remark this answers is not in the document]
ara-ayvazyan: Moved.

cpp/CMakeLists.txt
ara-ayvazyan: Updated.

cpp/inc/bond/core/config.h
ara-ayvazyan: Updated.

cpp/inc/bond/core/config.h
```diff
@@ -41,8 +41,23 @@
...
-#ifdef BOND_HEADER_ONLY
+
+#define BOND_LIB_TYPE_HEADER    1
+#define BOND_LIB_TYPE_STATIC    2
+#define BOND_LIB_TYPE_DYNAMIC   3   // Not implemented
+
+#ifndef BOND_LIB_TYPE
+#define BOND_LIB_TYPE BOND_LIB_TYPE_HEADER
+#elif (BOND_LIB_TYPE != BOND_LIB_TYPE_HEADER) && (BOND_LIB_TYPE != BOND_LIB_TYPE_STATIC)
+#error Unknown library type is defined for BOND_LIB_TYPE
```
chwarr: "Unsupported" instead of "unknown"? `BOND_LIB_TYPE_DYNAMIC` isn't yet supported, but it isn't unknown.

cpp/inc/bond/protocol/detail/rapidjson_utils.h
```diff
@@ -22,6 +22,10 @@ bool try_lexical_convert(const char* str, uint16_t& result);
 } // namespace bond
 
 
-#ifdef BOND_HEADER_ONLY
+#ifdef BOND_LIB_TYPE
+#if BOND_LIB_TYPE == BOND_LIB_TYPE_HEADER
 #include "rapidjson_utils_impl.h"
 #endif
+#else
+#error BOND_LIB_TYPE is undefined
```
chwarr: Do we need this explicit error handling? Don't all our supported compilers emit an error if you redefine a macro or access one that isn't defined? Similar comment for `BOND_DETAIL_HEADER_ONLY_INLINE`.

doc/src/bond_cpp.md
```diff
@@ -1775,6 +1775,10 @@ to reuse the pre-instantiated code. The one exception are applications using
 custom protocols - by definition templates pre-instantiated at the time Bond
 library was built can't support custom protocols.
 
+If header-only consumption of the library is not required, then defining `BOND_LIB_TYPE`
```
chwarr: Minor: "then defining the `BOND_LIB_TYPE` macro..."

doc/src/bond_cpp.md
```diff
@@ -1775,6 +1775,10 @@ to reuse the pre-instantiated code. The one exception are applications using
...
+macro to `BOND_LIB_TYPE_STATIC` will pre-compile more common code into `bond`
```
chwarr: Minor: "...into the `bond.lib`/`libbond.a` library..."

cpp/CMakeLists.txt
```diff
@@ -147,6 +147,8 @@ add_library (bond
     ${precompiled_sources}
     ${headers})
 
+target_compile_definitions (bond PUBLIC -DBOND_LIB_TYPE=BOND_LIB_TYPE_STATIC)
```
chwarr: We should be consistent with this across our compilation units, so I think we want to set this in Config.cmake like we do with BOOST_ALL_NO_LIB.

cpp/inc/bond/protocol/detail/rapidjson_utils.h
ara-ayvazyan: Redefinition is usually a warning. Accessing undefined one in pre-processor considers it as empty. We want to be sure that `BOND_LIB_TYPE` is explicitly defined. This will catch the cases where `config.h` is not included for example.

cpp/inc/bond/core/config.h
ara-ayvazyan: What if user really provides unknown one?

chwarr: An unknown one is also unsupported. ;-)

cpp/inc/bond/protocol/detail/rapidjson_utils.h
chwarr: OK. I thought they were always errors in both cases. Carry on.

cpp/CMakeLists.txt
chwarr: Wait. It's `PUBLIC`, so it will be propagated automatically. This is fine.

doc/src/bond_cpp.md
ara-ayvazyan: Updated.

doc/src/bond_cpp.md
ara-ayvazyan: Updated.

cpp/inc/bond/core/config.h
ara-ayvazyan: OK.

PR #718, opened 2017-11-15
[c++] Remove unused PrimitiveTypes

ara-ayvazyan:


chwarr:
The only harm I see in keeping this around is that we'd be prevented from removing `boost::mpl::list` when we finish the types transition. Do you expect that we'll have a macro like `BOND_PROVIDE_MPL_LIST` that we can put this code behind?

ara-ayvazyan:
I'm trying to eliminate some heavy headers earlier in the places where it is possible to do even without reconsidering `Schema` changes (`boost/mpl/for_each.hpp` in this case). We can do that way if we decide that this needs to stay.

PR #725, opened 2017-11-18
WIP: testing automated docker image building

chwarr:
This is a substantive change to Dockerfile to test automated building. It should build and publish the image.


chwarr:
Doesn't work from forks. Looks like it may have gotten branch names confused...

PR #726, opened 2017-11-18
WIP: testing docker image building - no op

chwarr:
This doesn't touch anything interesting, so it should skip the image build.


chwarr:
Doesn't work from forks.

PR #728, opened 2017-11-20
Split CI into flavor scripts.

tstein:
WIP while debugging.


tstein:
LGTM. Please review.

tools/ci-scripts/linux/build.zsh
```diff
@@ -7,35 +7,39 @@ FLAVOR=$2
...
-mkdir -p $SYMLINKED_HOME $BUILD_PATH
+mkdir -p $SYMLINKED_HOME $BUILD_ROOT
+ln -s /root/.ccache $SYMLINKED_HOME/.ccache
 ln -s /root/.stack $SYMLINKED_HOME/.stack
 
-cd $BUILD_PATH
+cd $BUILD_ROOT
 
 case "$FLAVOR" in
     cpp-*)
```
ara-ayvazyan: Why is this not moved to separate script as well?

tstein: Just to reduce the complexity of the change. The only parameters to `hs`, `cs`, and `java` were already passed as envvars, meaning the scripts themselves require no arguments;`cpp-*` is a little more complex. I'm imagining the next change passing most or all of this script's argv to the flavor-specific script, but didn't want to tackle that in this PR.

ara-ayvazyan: OK.

PR #729, opened 2017-11-20
[c++] Clean up includes

chwarr:
* Use #pragma once everywhere.
* Every header now include <bond/core/config.h> before anything else.
* Includes are done in sections, from least general to most general.
* Includes within each section are lexicographicaly sorted.
* CODING_GUIDELINES_CPP.md added to capture these guidelines.

### DIFF
--- a/cpp/CODING_GUIDELINES_CPP.md
+++ b/cpp/CODING_GUIDELINES_CPP.md
@@ -1,0 +1,86 @@
+# Bond C++ Coding Guidelines
+
+## Background
+
+A consistent coding style benefits authors and readers. Authors have fewer,
+trivial decisions to make, and readers' assumptions about overall structure
+are more likely to be correct.
+
+Here are the Bond C++ coding guidelines to help have a consistent style.
+
+Guidelines are not hard and fast rules.
+
+This is a work in progress.
+
+### History
+
+Bond has been developed by many people of many years. It is *not* internally
+consistent, so you will very easily find places where these guidelines are
+not followed.
+
+### Tooling
+
+Ideally, as many of these would be enforced via tooling to help with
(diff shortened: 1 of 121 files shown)


chwarr:
I've build much of this with G++ and Clang++ already, but we should wait until CI passes to make sure I didn't miss any corners.

cpp/CODING_GUIDELINES_CPP.md
```diff
@@ -0,0 +1,86 @@
+# Bond C++ Coding Guidelines
```
ara-ayvazyan: I think we should also mention about which headers are required to be included and in which cases it might be OK to skip. For example, if `std::vector` is used, then it is required to have `#include <vector>`. Also something about headers that only have forward declarations (e.g. use `_fwd.h`)?

chwarr: I'm having trouble articulating this beyond "Each header includes headers for what is uses." How would you say this?

ara-ayvazyan: Avoid implicit dependencies on indirectly included headers?

chwarr:
Based on [this build failure](https://travis-ci.org/Microsoft/bond/jobs/305016671) it looks like rapidjson_helper.h does use RapidJsonException, so I added that header back. @ara-ayvazyan had removed it in commit edf6bdc62c009c859b6bcd0e9fd44cf340c3444e.

cpp/CODING_GUIDELINES_CPP.md
chwarr: Done.

ara-ayvazyan:
Oh, wow. I did not realize that `RAPIDJSON_PARSE_ERROR` is such a hack..

PR #736, opened 2017-11-28
[c++] Faster Serializer::Field

ara-ayvazyan:
The change adds overloads of `Serializer::Field` function that avoids generating calls to `detail::omit_field` in the cases where the result is known at compilation. This makes the compilation a bit faster (it depends on actual types being used, but the expected estimate would be 3-5%) and also likely to generate faster, smaller and more efficient runtime code when compiler fails to optimize away the above calls.

### DIFF
--- a/cpp/inc/bond/core/transforms.h
+++ b/cpp/inc/bond/core/transforms.h
@@ -92,7 +92,8 @@ public:
     }
 
     template <typename T>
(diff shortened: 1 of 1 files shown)


chwarr:
Is this faster at runtime or compile time? Approximately how much faster?

ara-ayvazyan:
@chwarr, added a description.

PR #737, opened 2017-11-28
[c++] Remove duplicated code in DynamicParser

ara-ayvazyan:


cpp/inc/bond/core/parser.h
```diff
@@ -522,6 +491,13 @@ class DynamicParser
         else
             return detail::BasicTypeField(id, schema<Unknown>::type::metadata, type, BindUnknownField(transform), _input);
     }
+
+
+    void ReadFieldEndBegin(BondDataType& type, uint16_t& id)
```
chwarr: ReadSubsequentField?

cpp/inc/bond/core/parser.h
```diff
@@ -468,37 +455,19 @@ class DynamicParser
...
-            // the end of the struct, it means that:
-            //
-            // 1) Actual data in the payload had deeper hierarchy than payload schema.
-            //
-            // or
-            //
-            // 2) We parsed only part of the hierarchy because that was what
-            //    the transform "expected".
-            //
-            // In both cases we emit remaining fields as unknown
```
chwarr: There's a bug here that we probably should fix before refactoring. If we're parsing and we detect that we're in an unexpected hierarchy depth, that should cause serialization--and perhaps other transforms--to fail.

ara-ayvazyan: Why do you think that needs to be fixed before this change? Also won't the fix likely be in `detail::ParserInheritance`?

chwarr: Take a look at issue https://github.com/Microsoft/bond/issues/742 for the details of the problem.

Issue #735, opened 2017-11-28
undefined reference in inheritance.h in debug mode


FDUdannychen:
Learning from the example, my code looks like:
```
void DoSomething(
            bond::ext::gRPC::unary_call<
                bond::bonded<MyRequest>,
                bond::Void> call) override
        {
            auto request = call.request().Deserialize();
            //...
            call.Finish(bond::Void());
        }
```
When my project is compiled in Debug mode:

> cmake -DCMAKE_BUILD_TYPE=Debug

I will get error messages when `make`:
```
CMakeFiles/server.dir/host/server.cpp.o: In function `_ZN4bond6detail17ParserInheritanceIRNS_19CompactBinaryReaderINS_11InputBufferEEENS_13DynamicParserIS5_EEE4ReadINS_4Void6SchemaENS_10SerializerINS_19CompactBinaryWriterINS_20CompactBinaryCounter4typeEEENS_16BuiltInProtocolsEEEEEN5boost12disable_if_cIXgtsr15hierarchy_depthIT_EE5valuesr14expected_depthIT0_EE5valueEbE4typeERKSL_RKSM_':
/usr/local/include/bond/core/detail/inheritance.h:99: undefined reference to `bond::Void::Schema::metadata'
```
Switching between Clang3.8 and GNU5.4 will not help. But everything is working fine in Release mode:

> cmake -DCMAKE_BUILD_TYPE=Release          //or simply cmake without the build type arg

Am I missing something?

ara-ayvazyan:
From the error message it seems like you're missing a dependency on `bond` static library?

FDUdannychen:
@ara-ayvazyan But it's working in release mode. I don't see any dependency difference between debug and release.

chwarr:
It looks like the generated _grpc.h header may be missing a `#include <bond/core/bond_reflection.h>` when events or no-argument functions are being used.

Potential work around: add this before your include of the relevant generated _grpc.h headers:

```
#include <bond/core/bond_reflection.h>
```

After including this header your executable will have a dependency on libbond.a/bond.lib. That's where the metadata for types like `bond::Void` is compiled.

[...]

PR #738, opened 2017-11-29
[c++ grpc] Fix includes when bond::Void is used

chwarr:
When a service contains events or parameterless methods, the generated
code uses bond::Void as the request/result type. To serialize
bond::Void, bond_reflection.h needs to be included.

The generated service code now includes bond_reflection.h if any service
contains an event or a parameterless method.

Fixes https://github.com/Microsoft/bond/issues/735

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,8 @@ different versioning scheme, following the Haskell community's
   has required Boost 1.58 or later since version 5.2.0, but this was not
   enforced.
 * gRPC v1.7.1 is now required to use Bond-over-gRPC.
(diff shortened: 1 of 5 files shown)


ara-ayvazyan:
chwarr, is it possible to use gRPC with `bond_apply`?

chwarr:
It should work with bond_apply.lib; however, you'd have to make sure to `#include <bond/core/bond_apply.h>` before any of the generated `_grpc.h` headers. We have an [example that does this](https://github.com/Microsoft/bond/tree/master/examples/cpp/grpc/grpc_static_library) for the user-generated type.

Issue #617, opened 2017-09-18
Attributes for services


acgull:
In the [idl manual](https://microsoft.github.io/bond/manual/compiler.html#custom-attributes) it shows how to add attributes to services and methods.

When compiling the following bond file [test.txt](https://github.com/Microsoft/bond/files/1309816/test.txt) (renamed to txt so I can attach it)
I don't see any attributes for the service or the service methods (although I do see it for the struct).
Is there a special flag that needs to be used?
I run it using the following command
`gbc c# --grpc test.bond`

chadwalters:
Thanks for the report. Looks like this was missed in the implementation and needs to be added. I am working on this now.

chwarr:
@acgull, for my information, what are you planning to do with the attributes when you have them?

lalo:
This is being worked on https://github.com/Microsoft/bond/pull/618

acgull:
I want to add some metadata to the different methods. E.g. description

chadwalters:
@acgull: https://github.com/Microsoft/bond/pull/618 is complete as far as I can tell. Take a look there and verify that it meets your needs.

acgull:
Using bond 7.0.1 it generates the attributes in C# but when compiling I get the following error
```
: Attribute 'global::Bond.Attribute' is not valid on this declaration type. It is only valid on 'class, struct, enum, property, indexer, field, interface' declarations
```

This is the `AttributeAttributes.cs` file

```
#region Assembly Bond.Attributes, Version=7.0.1.100, Culture=neutral, PublicKeyToken=87e9ead25a117286
// MyPath\packages\Bond.Core.CSharp.7.0.1\lib\net45\Bond.Attributes.dll
#endregion

using System;

namespace Bond
{
    //
    // Summary:
    //     Specifies user defined schema attribute
    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Enum | AttributeTargets.Property | AttributeTargets.Field | AttributeTargets.Interface, AllowMultiple = true, Inherited = false)]
    public sealed class AttributeAttribute : Attribute
    {
        public AttributeAttribute(string name, string value);

        public string Name { get; }
        public string Value { get; }
    }
}
```

Looks like it doesn't allow `AttributeTargets.Method`

chwarr:
Good catch. Thanks.

For those also following along at home, the golden files output by the gbc-test aren't compiled by their respective language compilers...

[...]

PR #743, opened 2017-12-06
[c# grpc]  Allow [Bond.Attribute] on methods

chwarr:
* This fixes broken C# codegen when a service method has attributes.
* Add both C++ and C# unit test to make sure that service and method
  attributes are properly generated and accessible.

Fixes https://github.com/Microsoft/bond/issues/617


chwarr:
Build errors understood. Working on a fix.

cs/src/attributes/Attributes.cs
```diff
@@ -142,8 +142,14 @@ public sealed class RequiredOptionalAttribute : Attribute
...
     /// Specifies user defined schema attribute
     /// </summary>
-    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Enum | 
-        AttributeTargets.Interface | AttributeTargets.Field | AttributeTargets.Property, 
+    [AttributeUsage(
+        AttributeTargets.Class |
+        AttributeTargets.Enum |
+        AttributeTargets.Field |
+        AttributeTargets.Interface |
+        AttributeTargets.Method |
```
lalo: I'm guessing this was the missing one that caused the reported failure right?

chwarr: Yup.

Issue #742, opened 2017-12-05
[c++] BT_STOP during a base deserialization confuses deserializer: can read unrelated struct's field as its own


chwarr:
During deserialization, BT_STOP and BT_STOP_BASE are treated identically when handling what is expected to be a base. This can cause deserialization to succeed with incorrect data when a deserialization failure is expected. I've adjusted [one of the examples to demonstrate this failure](https://github.com/chwarr/bond/commit/a14486f0d3fd70cb0dd0de3922a492db98fa32cc) in a branch on my fork.

PR #745, opened 2017-12-08
[c++] Fix base to derived deserialization

ara-ayvazyan:
With this change deserialization will now fail by throwing `bond::CoreException` when `BT_STOP` is encountered while deserializing a base struct.

Fixes https://github.com/Microsoft/bond/issues/742

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,8 +48,8 @@ different versioning scheme, following the Haskell community's
 * gRPC v1.7.1 is now required to use Bond-over-gRPC.
 * Fixed includes for gRPC services with events or parameterless methods.
   [Issue #735](https://github.com/Microsoft/bond/issues/735)
-* Fixed a bug which would read unrelated struct's field(s) when deserializing base struct
-  [Issue #742](https://github.com/Microsoft/bond/issues/742).
+* Fixed a bug which would read an unrelated struct's field(s) when deserializing a
+  base struct. [Issue #742](https://github.com/Microsoft/bond/issues/742)
 
 ### C# ###
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,8 @@ different versioning scheme, following the Haskell community's
 * gRPC v1.7.1 is now required to use Bond-over-gRPC.
 * Fixed includes for gRPC services with events or parameterless methods.
   [Issue #735](https://github.com/Microsoft/bond/issues/735)
+* Fixed a bug which would read unrelated struct's field(s) when deserializing base struct
+  [Issue #742](https://github.com/Microsoft/bond/issues/742).
 
 ### C# ###
 
--- a/cpp/inc/bond/core/transforms.h
+++ b/cpp/inc/bond/core/transforms.h
@@ -388,6 +388,12 @@ protected:
         _required = next_required_field<typename schema<T>::type::fields>::value;
     }
 
+    void Base(bool done) const
+    {
+        if (done)
+            UnexpectedStructStopException();
+    }
+
     template <typename Head>
     typename boost::enable_if<std::is_same<typename Head::field_modifier,
                                            reflection::required_field_modifier> >::type
@@ -426,6 +432,8 @@ protected:
 private:
     BOND_NORETURN void MissingFieldException() const;
 
+    BOND_NORETURN void UnexpectedStructStopException() const;
+
     mutable uint16_t _required;
 };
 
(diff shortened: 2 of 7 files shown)


cpp/inc/bond/core/transforms.h
```diff
@@ -388,6 +388,12 @@ class RequiredFieldValiadator
         _required = next_required_field<typename schema<T>::type::fields>::value;
     }
 
+    void Base(bool done) const
```
ara-ayvazyan: Technically, this is a breaking change. Not sure if this needs to be mentioned in changelog.

chwarr: Yes, this should be mentioned in the change log.

CHANGELOG.md
```diff
@@ -48,6 +48,8 @@ different versioning scheme, following the Haskell community's
 * gRPC v1.7.1 is now required to use Bond-over-gRPC.
 * Fixed includes for gRPC services with events or parameterless methods.
   [Issue #735](https://github.com/Microsoft/bond/issues/735)
+* Fixed a bug which would read unrelated struct's field(s) when deserializing base struct
```
chwarr: Minor: a few words missing "...would read an unrelated struct's... deserializing a base struct." and move the period to the end of this sentence have have the "Issue #742" float at the end.

cpp/inc/bond/core/transforms.h
```diff
@@ -388,6 +388,12 @@ class RequiredFieldValiadator
...
+    {
+        if (done)
```
chwarr: Minor: slight preference for curly braces always.

cpp/inc/bond/core/transforms.h
```diff
@@ -426,17 +432,31 @@ class RequiredFieldValiadator
...
 
+    BOND_NORETURN void UnexpectedStructStopException() const;
+
     mutable uint16_t _required;
 };
 
 template <typename T>
 void RequiredFieldValiadator<T>::MissingFieldException() const
 {
+    (void)typename schema<T>::type();
```
chwarr: Why do we need to construct an instance of the field?

cpp/inc/bond/core/transforms.h
```diff
@@ -426,17 +432,31 @@ class RequiredFieldValiadator
...
           " is missing from " << schema<T>::type::metadata.qualified_name);
 }
 
+template <typename T>
+void RequiredFieldValiadator<T>::UnexpectedStructStopException() const
+{
+    (void)typename schema<T>::type();
+
+    BOND_THROW(CoreException,
+        "De-serialization failed: unexpected struct stop is encountered for "
```
chwarr: Minor: "unexpected struct stop encountered for..."

cpp/inc/bond/core/transforms.h
chwarr: It feels confusing to be doing this in something called "RequiredFieldValidator", as this doesn't have to do with required fields.

Should/can this be in `To` directly?

cpp/inc/bond/core/transforms.h
ara-ayvazyan: Was trying to be consistent here.

cpp/inc/bond/core/transforms.h
ara-ayvazyan: To force-instantiate a possibly template static data member `metadata` (see [this link error](https://ci.appveyor.com/project/MicrosoftBond/bond/build/1.0.1048/job/8o6t55ekn6lqoq0s)).

cpp/inc/bond/core/transforms.h
ara-ayvazyan: In that case the user will not be able to customize that behavior. If we treat this as an incorrect payload detection, then I guess should be OK to move it to `To<T>` transform.

cpp/inc/bond/core/transforms.h
chwarr: Can we comment the line?

cpp/inc/bond/core/transforms.h
chwarr: I'm fine having this apply to all cases of deserialization and not be customizable. This is about detecting an invalid payload/bad downcast.

To keep it customizable, we could have a different validator that composes with `RFV`. `PayloadIsntBeingDowncastIncorrectlyValidator<T> : T` or something.

cpp/inc/bond/core/transforms.h
```diff
@@ -550,9 +533,12 @@ class To
...
     {
-        bool done = AssignToBase(_var, value);
-        Validator::Base(done);
-        return done;
+        if (bool done = AssignToBase(_var, value))
+        {
+            UnexpectedStructStopException();
+        }
+
+        return false;
```
chwarr: Just wanted to check that it's structured this way on purpose (e.g., to give a name to the return value of `AssignToBase`) and that it shouldn't read `return done`.

PR #748, opened 2017-12-12
[c++] Improve bonded<void, Reader>::_Apply

ara-ayvazyan:
Simplifies the `bonded<void, Reader>::_Apply` function when used with protocols that specify `uses_marshaled_bonded<Reader>` to be `false_type` (currently all except `Simple`) by excluding some run-time code paths.

### DIFF
--- a/cpp/inc/bond/core/detail/protocol_visitors.h
+++ b/cpp/inc/bond/core/detail/protocol_visitors.h
@@ -207,10 +207,10 @@ inline bool Parse(const Transform& transform, Reader& reader, const Schema& sche
     return Parser<T, Schema, Transform>::Apply(transform, reader, schema, base);
 }
 
-template <typename Protocols, typename Transform, typename Reader, typename Schema>
-inline bool Parse(const Transform& transform, Reader& reader, const RuntimeSchema& schema, bool base)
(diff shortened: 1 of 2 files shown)


cpp/inc/bond/core/detail/protocol_visitors.h
```diff
@@ -207,42 +207,51 @@ inline bool Parse(const Transform& transform, Reader& reader, const Schema& sche
     return Parser<T, Schema, Transform>::Apply(transform, reader, schema, base);
 }
 
+template <typename T, typename Protocols, typename Transform, typename Reader, typename Schema>
+inline bool Parse(const Transform& transform, Reader& reader, const Schema& schema, std::nullptr_t, bool base)
```
chwarr: Don't forget to add a `#include <cstddef>` for `nullptr_t`.

chwarr:
Did the code before actually result in runtime conditional checks (when O2 is enable)? Color me surprised if the optimizer didn't remove that check.

ara-ayvazyan:
@chwarr, I've experimented with

ara-ayvazyan:
@chwarr, it does with `/GL`, the `/O2` is not enough.

PR #754, opened 2017-12-14
[c++] Improve StaticParser<Input>::ReadFields

ara-ayvazyan:
Simplifies the `StaticParser<Input>::ReadFields` function when used with protocols that do not implement field omitting by excluding some run-time code paths.


lalo:
We reviewed this IRL

PR #757, opened 2017-12-18
[c++] Fix missing type param. in MapTo<T>::Field

ara-ayvazyan:


cpp/inc/bond/core/transforms.h
```diff
@@ -830,7 +830,7 @@ class MapTo
         if (it != _mappings.end())
         {
             if (!it->second.fields.empty())
-                return Apply(MapTo(_var, it->second.fields), value);
+                return Apply<Protocols>(MapTo(_var, it->second.fields), value);
```
lalo: Could we search/grep for other similar bugs where we are using Apply without specializing with Protocols?

ara-ayvazyan: These are the ones that we missed somehow (~2nd so far).

PR #765, opened 2017-12-29
[c++ grpc] Fix race in unary_call destruction

ara-ayvazyan:
Fixes a race condition in `bond::ext::gRPC::detail::unary_call_impl` when one thread tries to send an error by calling `FinishWithError` when reference count becomes 1 during destruction, and the other thread that tries to perform `delete this` since the reference count becomes 0 due to a prior call to `Finish` by the user.

### DIFF
--- a/cpp/inc/bond/ext/grpc/detail/unary_call_impl.h
+++ b/cpp/inc/bond/ext/grpc/detail/unary_call_impl.h
@@ -113,58 +113,50 @@ namespace bond { namespace ext { namespace gRPC { namespace detail {
             // The response has been sent, so we no longer need to keep
             // ourselves alive: release the implicit initial refcount that
             // this instance was constructed with.
-            Release(ReleaseTrySend::No);
+            Release();
         }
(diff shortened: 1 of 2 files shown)


cpp/inc/bond/ext/grpc/detail/unary_call_impl.h
```diff
@@ -113,58 +113,50 @@ namespace bond { namespace ext { namespace gRPC { namespace detail {
...
+                // the last user reference has just gone away, but Finish was
+                // not called. In this case, we are responsible for sending
+                // an error response and decrementing the final ref
+                // count. FinishWithError will schedule the send of the
+                // error response, and notification of completion of the
+                // send via invoke() will decrement the final ref count.
+                // Since we hold the ref count ourselves, we will not
+                // get deleted until we're done sending.
+                //
+                // Case 2b:
```
chwarr: This comment doesn't apply here after the refactoring.

PR #773, opened 2018-01-09
[c++] Improve compact/fast binary protocols' Skip

ara-ayvazyan:
The change makes more compile-time dispatch on `BondDataType` in protocols' `Skip` functionality instead of relying on compiler's optimization.

### DIFF
--- a/cpp/inc/bond/protocol/compact_binary.h
+++ b/cpp/inc/bond/protocol/compact_binary.h
@@ -396,14 +396,21 @@ public:
     }
 
 protected:
-    template <BondDataType T>
+#if defined(_MSC_VER) && (_MSC_VER < 1900)
+    // Using BondDataType directly in non-trivial boolean template checks fails on VC12.
+    using BT = std::underlying_type<BondDataType>::type;
+#else
+    using BT = BondDataType;
+#endif
+
+    template <BT T>
     typename boost::enable_if_c<(T == BT_BOOL || T == BT_UINT8 || T == BT_INT8)>::type
     SkipType(uint32_t size = 1)
     {
         _input.Skip(size * sizeof(uint8_t));
     }
 
-    template <BondDataType T>
+    template <BT T>
     typename boost::enable_if_c<(T == BT_UINT16 || T == BT_UINT32 || T == BT_UINT64
                                 || T == BT_INT16 || T == BT_INT32 || T == BT_INT64)>::type
     SkipType()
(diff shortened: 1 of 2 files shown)


cpp/inc/bond/protocol/compact_binary.h
```diff
@@ -381,129 +381,231 @@ class CompactBinaryReader
...
+
+    template <BondDataType T>
+    typename boost::enable_if_c<(T == BT_SET || T == BT_LIST)>::type
+    SkipType()
+    {
+        BondDataType element_type;
+        uint32_t     size;
+
+        ReadContainerBegin(size, element_type);
+
```
chwarr: Minor: could collapse these into one code paragraph

cpp/inc/bond/protocol/compact_binary.h
```diff
@@ -381,129 +381,231 @@ class CompactBinaryReader
...
-        switch (type)
+        SkipType(type);
+    }
+
+protected:
+    template <BondDataType T>
+    typename boost::enable_if_c<(T == BT_BOOL || T == BT_UINT8 || T == BT_INT8)>::type
+    SkipType(uint32_t size = 1)
+    {
+        _input.Skip(size * sizeof(uint8_t));
```
chwarr: We need to make sure that the multiplication doesn't overflow.

If this `size`--or any of the other `size` argument in the other overloads/specializations--come from the payload, we'll need to do a runtime check. (See https://github.com/Microsoft/bond/blob/master/cpp/inc/bond/core/detail/checked_add.h for helpers that we have for doing this on addition.)

If these are just other constants in the code, we can perform a case-by-case analysis and add a (static?) assert.

cpp/inc/bond/protocol/fast_binary.h
```diff
@@ -235,141 +235,215 @@ class FastBinaryReader
...
+    }
+
+    template <BondDataType T>
+    typename boost::enable_if_c<(T == BT_SET || T == BT_LIST)>::type
+    SkipType()
+    {
+        BondDataType element_type;
+        uint32_t size;
+
+        ReadContainerBegin(size, element_type);
```
chwarr: Minor: can collapse into one code paragraph.

cpp/inc/bond/protocol/compact_binary.h
```diff
@@ -381,129 +381,231 @@ class CompactBinaryReader
...
+    SkipType(uint32_t size)
+    {
+        for (uint32_t i = 0; i < size; ++i)
+        {
+            SkipType<T>();
+        }
+    }
+
+    template <typename... Args>
+    void SkipType(BondDataType type, Args&&... args)
```
chwarr: Does this ever get called with non-empty `args`?

ara-ayvazyan: Yeah, this is just a trick to share the same code for `SkipType(type)` and `SkipType(type, size)`.

chwarr: 👌

cpp/inc/bond/protocol/compact_binary.h
ara-ayvazyan: Added #774.

PR #775, opened 2018-01-10
[c++] Fix loop overflows

ara-ayvazyan:


chwarr:
Don't worry about the "VSTS: Travis Ubuntu image - forks" check. That's me playing with more automation for the Travis CI images. But it's not working right now...

cpp/inc/bond/core/schema.h
```diff
@@ -261,16 +265,26 @@ class InitSchemaDef
     template <typename T>
     uint16_t GetStructDef() const
     {
-        size_t n;
+        const auto& structs = _schema.structs;
+
+        BOOST_ASSERT(structs.size() <= (std::numeric_limits<uint16_t>::max)());
+
+        auto it = std::find_if(
+            std::cbegin(structs),
```
chwarr: Apparently `cbegin` is [missing](https://travis-ci.org/Microsoft/bond/jobs/327429088) from the standard library we're using in Travis.

cpp/inc/bond/protocol/fast_binary.h
```diff
@@ -373,7 +373,7 @@ class FastBinaryReader
         uint32_t size;
 
         ReadContainerBegin(size, element_type);
-        for (uint32_t i = 0; i < size; ++i)
+        for (int64_t i = 0; i < size; ++i)
```
chwarr: Don't the `SkipType(size_t)` functions also need checking?

ara-ayvazyan: I was planning to fix that separately. This change does not actually add any runtime checks.

PR #777, opened 2018-01-16
[c++] Add missing static_casts in io_manager tests

chwarr:
grpc::Alarm takes a void*, but io_manager expects io_manager_tag*. Thus,
we should have be casting alarm_completion_tag* to io_manager_tag*.


chwarr:
Moved to #778.

PR #781, opened 2018-01-25
[c# grpc] Add support for streaming RPCs

chwarr:
This is rough at the moment, and my TODO list is pretty big. You're welcome to start reviewing, but I'll ping you, @lalo, again when I've finished up the TODO items as well.

TODO:

* [x] Make streaming/MethodType a non-breaking change in JSON AST
* [x] Augment golden tests
* [x] Scrub Haskell code for consistency of naming and order of Void,
  Unary, Streaming
* [x] Update documentation
* [x] Change to stream keyword
* [x] Improve error messages when parsing stream/var in method
* [x] Clean up streaming event checking in parser
* [x] Ensure no regressions in unary cases
* [x] Add tests for generic service + streaming


examples/cs/grpc/scalar/svc.cs
```diff
@@ -0,0 +1,162 @@
...
+// This code was generated by a tool.
+//
+//   Tool : Bond Compiler 0.11.0.0
+//   Input filename:  scalar.bond
+//   Output filename: scalar_grpc.cs
+//
+// Changes to this file may cause incorrect behavior and will be lost when
+// the code is regenerated.
+// <auto-generated />
+//------------------------------------------------------------------------------
```
lalo: Why are we including this codegen file? The csproj will generate scalar_grpc.cs, can't seem to decipher the need for this.

compiler/src/Language/Bond/Syntax/Types.hs
```diff
@@ -121,22 +123,30 @@ data TypeParam =
...
     |
     Event
         { methodAttributes :: [Attribute]   -- zero or more attributes
         , methodName :: String              -- method name
-        , methodInput :: Maybe Type         -- method parameter
+        , methodInput :: MethodType         -- method parameter
         }
     deriving (Eq, Show)
 
+methodTypeToOldStyle :: MethodType -> Maybe Type
```
lalo: A comment would be helpful here to state explicit that this was caused by adding stream and not 'old' as in 'old bond'.

examples/cs/grpc/scalar/svc.cs
chwarr: Ah. This is left-over from when I hand-wrote the initial streaming bindings to figure out what we needed to generate. I've reverted these changes in commit 6a29ad9b.

compiler/src/Language/Bond/Syntax/Types.hs
chwarr: Take a look at what I've done in commit 5fdc4d6.

chwarr:
@lalo: I've completed all the TODO items and squashed everything together. Can you review again?

If you want to see my working branch with all the incremental changes, it's at https://github.com/chwarr/bond/tree/chwarr/cs-stream-all

chwarr:
I considered making streaming methods no-ops in C++ (e.g., generating a comment). Instead of detecting that C++ doesn't support a streaming method at C++ development and compile time, I thought that an error message from gbc was a better development experience.

It should be a relatively short-lived situation as well.

chwarr:
Looks like AppVeyor failed because it couldn't talk to Chocolatey to install stack. I'm re-running those builds now.

PR #788, opened 2018-02-02
[c++] Protocol Buffers writer

ara-ayvazyan:
TODO:
* [x] Update documentation
* [x] Convert `wstring` to `string`
* [x] Add example
* [x] Transcoding


thirdparty/CMakeLists.txt
```diff
@@ -4,3 +4,13 @@ if (BOND_ENABLE_GRPC)
     cxx_add_compile_options (Clang -Wno-unused-value)
     add_subdirectory(grpc)
 endif()
+
+if (NOT BOND_SKIP_CORE_TESTS)
```
tstein: Does it make sense to tie this to `CORE_TESTS`?

cpp/inc/bond/protocol/detail/protobuf_utils.h
```diff
@@ -0,0 +1,299 @@
...
+                return WireType::VarInt;
+            }
+            break;
+
+        case BT_STRUCT:
+        case BT_STRING:
+            return WireType::LengthDelimited;
+
+        case BT_WSTRING:
+            NotSupportedException("Wide string");
```
tstein: Any thoughts on handling `WSTRING` in the same path as `STRING`? Unless you forsee a good way for protobufs to represent wstrings without losing the encoding, I think we should just do that and add "wstrings will be mapped to strings when writing to protobufs" to the docs.

cpp/inc/bond/protocol/detail/protobuf_utils.h
```diff
@@ -0,0 +1,299 @@
...
+    inline Encoding ReadEncoding(BondDataType type, const Metadata& metadata)
+    {
+        return ReadEncoding(type, metadata, AttributeName<>::encode);
+    }
+
+    inline Encoding ReadKeyEncoding(BondDataType type, const Metadata& metadata)
+    {
+        if (type == BT_FLOAT || type == BT_DOUBLE)
+        {
+            NotSupportedException("Float/double map key");
```
tstein: This is oddly specific. Why is this a problem?

cpp/inc/bond/protocol/detail/protobuf_utils.h
```diff
@@ -0,0 +1,299 @@
...
+        {
+        case BT_STRING:
+        case BT_STRUCT:
+        case BT_LIST:
+            return Packing::False;
+
+        case BT_WSTRING:
+        case BT_SET:
+        case BT_MAP:
+            NotSupportedException("Wide string or nested set/map");
```
tstein: Why are nested sets/maps a problem?

cpp/inc/bond/protocol/detail/protobuf_utils.h
```diff
@@ -0,0 +1,299 @@
...
+        }
+    }
+
+    inline uint32_t MakeTag(uint16_t id, WireType wireType)
+    {
+        BOOST_ASSERT(wireType != Unavailable<WireType>());
+
+        if (id == 0 || (id >= 19000 && id <= 19999))
+        {
+            NotSupportedException("Field ordinal with value 0 or in the range 19000-19999");
```
tstein: It's worth leaving a comment for us that protobuf fields count from 1. What's the deal with 19000-19999?

cpp/inc/bond/protocol/protobuf_binary_writer.h
```diff
@@ -0,0 +1,579 @@
...
+                detail::proto::NotSupportedException("Inheritance");
+            }
+
+            if (!_fields.empty())
+            {
+                const auto& info = _fields.top(std::nothrow);
+
+                if (info.level > 1)
+                {
+                    detail::proto::NotSupportedException("Container nesting");
```
tstein: Struct nesting? What's the issue there?

cpp/inc/bond/protocol/protobuf_binary_writer.h
```diff
@@ -0,0 +1,579 @@
...
+        }
+
+        void WriteFieldEnd()
+        {
+            _fields.pop(std::nothrow);
+        }
+
+        void WriteContainerBegin(uint32_t size, BondDataType type)
+        {
+            BOOST_VERIFY(size != 0);
```
tstein: We can't write empty containers?

thirdparty/CMakeLists.txt
tstein: Actually, I think it does. Nothing needed here.

cpp/inc/bond/protocol/detail/protobuf_utils.h
ara-ayvazyan: Restriction in protobuf: https://developers.google.com/protocol-buffers/docs/proto3#maps
> where the key_type can be any integral or string type (so, any scalar type except for floating point types and bytes)

cpp/inc/bond/protocol/detail/protobuf_utils.h
ara-ayvazyan: Cannot be expressed in protobuf (e.g. `vector<vector<bool>>` cannot be mapped to `repeated repeated bool`), unless we box it.

cpp/inc/bond/protocol/detail/protobuf_utils.h
ara-ayvazyan: A reserved range: https://developers.google.com/protocol-buffers/docs/proto3#assigning-tags
> The smallest tag number you can specify is 1, and the largest is 229 - 1, or 536,870,911. You also cannot use the numbers 19000 through 19999

cpp/inc/bond/protocol/protobuf_binary_writer.h
ara-ayvazyan: Not a struct nesting, but a container (with struct element in this case, e.g. `vector<vector<S>>`).

cpp/inc/bond/protocol/protobuf_binary_writer.h
ara-ayvazyan: We always skip them (yet to verify a case with `vector<blob>` with empty `blob`(s)).

CHANGELOG.md
```diff
@@ -55,6 +55,8 @@ different versioning scheme, following the Haskell community's
 * Fixed a race condition when `bond::ext::gRPC::io_manager::shutdown` and
   `bond::ext::gRPC::io_manager::wait` are called concurrently.
 * Fixed a race condition during `bond::ext::gRPC::unary_call` destruction.
+* Added `bond::ProtobufBinaryWriter<Buffer>` for serialization to Protocol
+  Buffers binary format.
```
lalo: Should we mention what version of PB are we writing to?

cpp/inc/bond/protocol/detail/protobuf_utils.h
lalo: All NotSupportedExceptions mean that ProtoBuf does not support it? Could it be confusing since it could be the case that Bond does not support it currently or that Protobuf does not support it? Meaning that we would need something that explains further, a better exception message or a comment explaining why, etc.

doc/src/bond_cpp.md
```diff
@@ -1271,6 +1271,37 @@ See examples:
...
+  attributes for numeric key/value types with `Fixed` and `ZigZag` encoding.
+- Since Bond uses a list representation on the wire for `nullable<T>`, it can
+  also be used to represent a scalar field (`optional` for `proto2`) or a
+  `repeated` one which is expected to have at most one element. Note, that in
+  case of a numeric scalar field or an unpacked `repeated` numeric type, the
+  `ProtoPack("False")` must be used.
+- The smaller 8 and 16 bit numeric Bond types can also be mapped to 32 or 64 bit
+  ones with corresponding encoding.
+
+
```
lalo: We could link to an example/test here like in the rest of documentation.

CHANGELOG.md
ara-ayvazyan: If you mean `proto2` vs `proto3` then the encoding is compatible (same?). The difference is mostly about IDL and default behavior.

cpp/inc/bond/protocol/detail/protobuf_utils.h
tstein: This is a good idea. Let's draw a distinction between "we haven't done it yet" and "we don't think there's a good way to do it".

doc/src/bond_cpp.md
```diff
@@ -1271,6 +1271,37 @@ See examples:
 - `examples/cpp/core/simple_json`
 
 
+Protocol Buffers Interop
+========================
+
+Bond provides a `ProtobufBinaryWriter` that can be used as any other built-in
+protocol writer in order to serialize objects to Protocol Buffers binary format.
```
tstein: `the Protocol Buffers` or `Protocol Buffers'`

doc/src/bond_cpp.md
```diff
@@ -1271,6 +1271,37 @@ See examples:
...
+It requires a compatible .bond schema definition that will map to corresponding
+.proto schema:
+
+- Field ordinals must match.
+- The `bool`, `enum`, `uint32`, `uint64`, `int32`, `int64`, `float`,
+  `double`, and `string` types are mapped to corresponding ones with same name.
+- The `bytes` is represented as a `blob`.
```
tstein: "`bytes` type"?

cpp/inc/bond/protocol/detail/protobuf_utils.h
ara-ayvazyan: From writer's perspective it is not supported by protobuf (except of `oneof` which is obvious). Please suggest a better phrasing for exception/message to make it clear.

doc/src/bond_cpp.md
ara-ayvazyan: Will add an example as well. Looks like both examples and tests are tied to `BOND_SKIP_CORE_TESTS` (was trying to avoid unnecessarily build protobuf).

doc/src/bond_cpp.md
ara-ayvazyan: Yep (hence `marked`)

cpp/inc/bond/protocol/detail/protobuf_utils.h
tstein: What do you think about `NotSupportedException` vs. `NotImplementedException`?

To be clear, I don't think we need the second one if the final version of this change only has exceptions of the first time. The wstring support is the only thing that jumps out at me as belonging to the second type.

ara-ayvazyan: Well, the `wstring` is more in the same group with nested containers for me (supported on bond, not on protobuf), but I'll be implementing a conversion for it as you suggested. that sounds reasonable.

cpp/inc/bond/protocol/detail/protobuf_utils.h
```diff
@@ -0,0 +1,296 @@
...
+        {
+            return Encoding::Fixed;
+        }
+
+        if (std::strcmp(it->second.c_str(), "ZigZag") == 0)
+        {
+            return Encoding::ZigZag;
+        }
+
+        InvalidEnumValueException(name.c_str(), "Encoding");
```
ara-ayvazyan: Should be `it->second.c_str()`.

cpp/inc/bond/protocol/detail/protobuf_utils.h
```diff
@@ -0,0 +1,296 @@
...
+        {
+            return Unavailable<Packing>();
+        }
+
+        if (std::strcmp(it->second.c_str(), "False") == 0)
+        {
+            return Packing::False;
+        }
+
+        InvalidEnumValueException(name.c_str(), "Packing");
```
ara-ayvazyan: Same here: `it->second.c_str()`.

doc/src/bond_cpp.md
```diff
@@ -1271,6 +1271,42 @@ See examples:
...
+  will be packed for scalar numeric types. For unpacked encoding, an additional
+  `ProtoPack("False")` attribute must be used.
+- The `map` requires different `ProtoEncodeKey` and `ProtoEncodeValue`
+  attributes for numeric key/value types with `Fixed` and `ZigZag` encoding.
+- Since Bond uses a list representation on the wire for `nullable<T>`, it can
+  also be used to represent a scalar field (`optional` for `proto2`) or a
+  `repeated` one which is expected to have at most one element. Note, that in
+  case of a numeric scalar field or an unpacked `repeated` numeric type, the
+  `ProtoPack("False")` must be used.
+- The smaller 8 and 16 bit numeric Bond types can also be mapped to 32 or 64 bit
```
ara-ayvazyan: Can be mapped to 32 bit ones only.

cpp/inc/bond/protocol/protobuf_binary_writer.h
```diff
@@ -0,0 +1,681 @@
...
+            info.field.metadata = &metadata;
+            info.has_element = false;
+            info.is_list = (type == BT_LIST);
+
+            _fields.push(info);
+        }
+
+        void WriteFieldBegin(BondDataType type, uint16_t id)
+        {
+            WriteFieldBegin(type, id, schema<Unknown>::type::metadata);
```
ara-ayvazyan: Need to skip the unknown fields during transcoding instead of writing them with default encoding which might be wrong.

cpp/inc/bond/protocol/protobuf_binary_writer.h
```diff
@@ -0,0 +1,692 @@
...
+        void Write(const blob& value)
+        {
+            WriteElement(value, _fields.top(std::nothrow));
+        }
+
+        void WriteContainerEnd()
+        {
+            auto& info = _fields.top(std::nothrow);
+
+            if (info.has_element)
```
ara-ayvazyan: Can be an assert.

cpp/test/core/protobuf_writer_test.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+    CheckBinaryFormat<
+        unittest::proto::UnpackedIntegersContainer, unittest::UnpackedIntegersSetContainer>();
+}
+
+using IntegerSetContainerWrongEncodingTests_Types = expand<integer_types, int8_t>;
+BOOST_AUTO_TEST_CASE_TEMPLATE(IntegerSetContainerWrongEncodingTests, T,
+    IntegerSetContainerWrongEncodingTests_Types)
+{
+    CheckUnsupportedType<unittest::BoxWrongEncoding<set<T> > >();
```
ara-ayvazyan: Missing `std::` for a `set`.

cpp/test/core/to_proto.h
```diff
@@ -0,0 +1,328 @@
...
+    is_blob_container
+        : std::false_type {};
+
+    template <typename T> struct
+    is_blob_container<T, typename boost::enable_if<is_list_container<T> >::type>
+        : std::is_same<typename element_type<T>::type, blob::value_type> {};
+
+    template <typename T> struct
+    is_blob
+        : std::integral_constant<bool, is_blob_container<T>::value || std::is_same<T, blob>::value> {};
```
ara-ayvazyan: Checking for `blob` explicitly is not needed.

ara-ayvazyan:
@chwarr, ready for review :)

cpp/inc/bond/protocol/detail/simple_array.h
```diff
@@ -20,20 +22,29 @@ namespace detail
...
+
+    SimpleArray(const SimpleArray& other) = delete;
+    SimpleArray& operator=(const SimpleArray& other) = delete;
 
     ~SimpleArray()
     {
         memfree();
     }
 
+    bool empty() const
```
ara-ayvazyan: Most of such functions can be marked as `BOND_NOEXCEPT`.

cpp/inc/bond/protocol/protobuf_binary_writer.h
```diff
@@ -0,0 +1,742 @@
...
+        template <typename T>
+        void LengthEnd(T&)
+        {}
+
+
+        Buffer& _output;
+        const bool _skip_unknowns;
+        const uint32_t* _it;
+        detail::SimpleArray<uint32_t> _stack;
+        detail::SimpleArray<uint32_t> _lengths;
```
ara-ayvazyan: The `_stack` and `_lengths` data members are only used by `Pass0`.

.gitmodules
```diff
@@ -5,3 +5,7 @@
 [submodule "thirdparty/rapidjson"]
     path = thirdparty/rapidjson
     url = https://github.com/miloyip/rapidjson.git
+
+[submodule "thirdparty/protobuf"]
+    path = thirdparty/protobuf
+    url = https://github.com/google/protobuf.git
```
ara-ayvazyan: Alternatively, we could use the nested `grpc/third_party/protobuf` submodule.

PR #789, opened 2018-02-04
[doc] Syntax highlighting

ara-ayvazyan:


doc/src/bond_java.md
```diff
@@ -24,25 +24,30 @@ Basic example
 In Bond data schemas are defined using idl-like
 [syntax](compiler.html#idl-syntax):
 
+```
     namespace examples
```
chwarr: Using both \`\`\` and four spaces causes the code blocks all be indented an extra four spaces when the generated HTML is produced by Pandoc.

ara-ayvazyan:
Yep, sending.

doc/src/compiler.md
```diff
@@ -875,7 +875,7 @@ Other complex types are:
 
     ```javascript
     {
-      "type": "constant,
+      "type": constant,
```
chwarr: It's actually `"type": "constant"` (both are strings).

ara-ayvazyan: Ok.

PR #790, opened 2018-02-06
Minor fixes (proposals)

kant:


chwarr:
Thank you for your pull request, but we prefer the formatting the way it currently is. We like showing the links directly in these sort of lists so people know where they're going without needing to hover (mismatched links notwithstanding).

(See earlier PR where this was also discussed: https://github.com/Microsoft/bond/pull/654)

PR #793, opened 2018-02-10
[travis] Include gbc dependencies in CI image

chwarr:
With this commit, the package dependencies that gbc and gbc-tests use
will be built once with Stack when creating the CI image. Subsequent CI
builds will then just build the gbc and gbc-tests parts and link with
the already built dependencies.

Should a commit modify the dependencies of gbc/gbc-tests, Stack is smart
enough to figure this out, and it will build the missing dependencies,
until a newer CI image is used that contains the updated dependencies.

Stack and cabal don't allow us to just redirect their intermediate
output directories, so the build_init_srack.zsh and build.zsh scripts
now collaborate to symlink .stack-work so that it's shared between image
building and image execution.

This increases the image size by about 2 GiB.

Minor changes:

* Added BUILD_SOURCE_COMMIT Docker build argument. This needs to be set
  to a commit in the https://github.com/Microsoft/bond repository that
  is used to determine the dependencies to build.
* The STACK_ROOT environment variable is defined in the Dockerfile so
  that all invocations of stack--both during image build and during
  image execution--will share the same location.
* Moved the BOND_ROOT environment variable to Dockerfile so it's shared
  across all scripts and across image build and image execution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,7 @@ before_install:
     - docker images # Dump the image ID
 
 script:
+    # TODO: when switching to consuming an image that has a saved .stack-work, remove the .stack caching and mounting.
     - docker run -v $HOME/.ccache:/root/.ccache -v $HOME/.stack:/root/.stack -v `pwd`:/root/bond $CI_BUILD_CONTAINER $HOME $FLAVOR $BOOST $COMPILER
     # docker runs as root and may leave files in the following directories that are not readable by the travis user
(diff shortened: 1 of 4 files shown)


chwarr:
Missing `BOND_ROOT` should be fixed in commit 62f6f01932be39accf5764. I've also merge with the recent Dockerfile changes in master (commit 230b3415650ac03cba17dd).

chwarr:
Let's try that fix again with commit b1b6b186.

chwarr:
Recalling for now so I can debug the missing gradle-plugin problem...

PR #803, opened 2018-02-24
[c++] Fix value::Deserialize of non-matching types

ara-ayvazyan:
The following code does not compile on clang. The issue does not occur on MSVC and GCC due to active bugs.
```cpp
bond::InputBuffer input;
bond::CompactBinaryReader<bond::InputBuffer> reader{ input };
bond::value<uint32_t, decltype(reader)&> value{ reader };
uint8_t var;
value.Deserialize(var); // Deserialize into a non-matching type
```
The issue is that all the SFINAE-enabled overloads of `Deserialize` inside of the base `bond::value_common` and derived `bond::value` have the same signature (e.g. all use the return type for SFINAE check and the rest of the signature is identical), so the one in the base class cannot be found because it will be shadowed by the ones in derived class. The fix is to change the signature of the base `Deserialize` by using a default parameter for a SFINAE check rather than a return type so that the function will not be shadowed by derived ones.

### DIFF
--- a/cpp/inc/bond/core/value.h
+++ b/cpp/inc/bond/core/value.h
@@ -209,8 +209,7 @@ public:
 
     // skip value of non-matching type
     template <typename Protocols = BuiltInProtocols, typename X>
-    typename boost::disable_if_c<is_matching<T, X>::value>::type
-    Deserialize(X& /*var*/) const
+    void Deserialize(X& /*var*/, typename boost::disable_if<is_matching<T, X> >::type* = nullptr) const
     {
         Skip();
     }


cpp/inc/bond/core/value.h
```diff
@@ -209,8 +209,7 @@ class value_common
 
     // skip value of non-matching type
     template <typename Protocols = BuiltInProtocols, typename X>
-    typename boost::disable_if_c<is_matching<T, X>::value>::type
-    Deserialize(X& /*var*/) const
+    void Deserialize(X& /*var*/, typename boost::disable_if<is_matching<T, X> >::type* = nullptr) const
```
ara-ayvazyan: This changes the function signature so that the similar overloads in derived types won't hide this one.

ara-ayvazyan:
The java build is failing again.

chwarr:
Java build issue is known, and we've added logging to help debug that the next time it occurs.

chwarr:
Can you add a copy/pasteable commit description to this PR? The first comment is close. It describes the problem, but doesn't explain the fix.

ara-ayvazyan:
Updated the description.

PR #805, opened 2018-02-26
[c++] Remove is_blob/is_nullable traits

ara-ayvazyan:
The `blob` and `nullable` types are not customizable. Also `is_blob` and `is_nullable` traits are not part of container concept.


chwarr:
This change deserves mention in the changelog, as something in a non-detail namespace is being removed.

Issue #636, opened 2017-10-06
Investigate whether a ProjectSchemaDefinition can be used to provide automatic codegen for .bond file in the new .NET Core project system


chwarr:
As reported in #634, installing the Bond.CSharp NuGet package doesn't provide the BondCodegen build action. (Reloading the solution sometimes makes this available.) There's [an open feature request in the .NET project system](https://github.com/dotnet/project-system/issues/113) to make `AvailableItemName` work again.

The new project system appears to use [`ProjectSchemaDefinitions`](https://github.com/dotnet/project-system/issues/1715#issuecomment-285490242) to do things [like automatically cause .cs files to be added to the `Compile` item](https://github.com/dotnet/project-system/blob/2c7322d35b7c37f1cbfd181130f00f694ca65235/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/CSharp.ProjectItemsSchema.xaml).

We should investigate whether this is something that can be included in the NuGet package for the project system to pick up. If it can, it would make .bond files in the project directory automatically run codegen, which would be better than what happens today (nothing).

PR #806, opened 2018-02-27
[c#] Add automatic codegen for .NET Core projects

chwarr:
* Enabled automatic codegen when implicit inclusion is turned on via the
  MSBuild property EnableDefaultCompileItems
* Added Visual Studio CPS UI integration so that .bond files appear in
  the Visual Studio 2017+ UI.

Fixes https://github.com/Microsoft/bond/issues/636


lalo:
Had to manually merge it after screwing it up with the github ui. Merged with a8821da8082fe90d1fee5982aeaa2a7d82ec465b

PR #807, opened 2018-02-27
ci: Add debugging info for travis gradle install failures.

tstein:


tstein:
I can't repro the failures we've been seeing locally, so I'd like to commit this to try to catch it on travis.

PR #808, opened 2018-03-02
[c++] Add detail::Field helper function

ara-ayvazyan:
The change adds a helper `detail::Field` that deals with calling into the correct version of `Field` of the provided Transform.
This also fixes the broken move constructor of `bond::bonded<T, Reader&>`.

### DIFF
--- a/cpp/inc/bond/core/bonded.h
+++ b/cpp/inc/bond/core/bonded.h
@@ -26,6 +26,8 @@ template <typename Protocols, typename Transform, typename T, typename Reader>
 typename boost::enable_if<need_double_pass<Transform>, bool>::type inline
 ApplyTransform(const Transform& transform, const bonded<T, Reader>& bonded);
 
+
+// Helper function move_data for dealing with [not] moving a Reader& in bonded<T, Reader&>
 template <typename T, typename U>
 typename boost::enable_if<std::is_reference<T>, T>::type
 inline move_data(U& data)
--- a/cpp/inc/bond/core/bonded.h
+++ b/cpp/inc/bond/core/bonded.h
@@ -26,6 +26,22 @@ template <typename Protocols, typename Transform, typename T, typename Reader>
 typename boost::enable_if<need_double_pass<Transform>, bool>::type inline
 ApplyTransform(const Transform& transform, const bonded<T, Reader>& bonded);
 
+template <typename T, typename U>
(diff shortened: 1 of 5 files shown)


cpp/inc/bond/core/parser.h
```diff
@@ -155,38 +142,18 @@ class StaticParser
...
-    NextField(const T&, const Transform& transform)
-    {
-        return transform.Field(T::id, T::metadata, value<typename T::field_type, Input>(_input));
-    }
-
-
-    template <typename T, typename Transform>
-    typename boost::enable_if_c<is_reader<Input>::value && !is_nested_field<T>::value
-                             && is_fast_path_field<T, Transform>::value, bool>::type
+    typename boost::enable_if_c<is_reader<Input>::value && !is_nested_field<T>::value, bool>::type
```
ara-ayvazyan: Thinking to move the check for `is_nested_field` to utility function as well. Seems the same everywhere.

cpp/inc/bond/core/bonded.h
```diff
@@ -26,6 +26,22 @@ template <typename Protocols, typename Transform, typename T, typename Reader>
 typename boost::enable_if<need_double_pass<Transform>, bool>::type inline
 ApplyTransform(const Transform& transform, const bonded<T, Reader>& bonded);
 
+template <typename T, typename U>
+typename boost::enable_if<std::is_reference<T>, T>::type
+inline move_data(U& data)
```
biocomp: Please add comments for why this is needed ) Other readers could not understand the issue and the solution )

cpp/inc/bond/core/detail/parser_utils.h
```diff
@@ -0,0 +1,91 @@
...
+    template <typename T, typename Transform, typename X>
+    typename boost::disable_if_c<is_reader<X>::value && is_fast_path_field<T, Transform>::value, bool>::type
+    inline Field(const T&, const Transform& transform, X&& value)
+    {
+        return transform.Field(T::id, T::metadata, GetFieldValue<T>(std::forward<X>(value)));
+    }
+
+
+    template <typename Reader, typename Transform>
+    inline bool Field(const FieldDef& field, const RuntimeSchema& schema, const Transform& transform, Reader& input)
```
chwarr: Consider `NonBasicTypeField` as an analogue to `BasicTypeField`.

PR #811, opened 2018-03-06
Upgraded grpc to version 1.10.0

grisharav:
Upgraded grpc to latest released version

### DIFF
--- a/compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
+++ b/compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
@@ -263,11 +263,11 @@ inline #{className}::#{proxyName}<TThreadPool>::#{proxyName}(
             Async#{methodName}(::std::make_shared< ::grpc::ClientContext>(), #{bonded methodInput}{request});
         }|]
 
-        privateProxyMethodDecl Function{..} = [lt|const ::grpc::RpcMethod rpcmethod_#{methodName}_;|]
-        privateProxyMethodDecl Event{..} = [lt|const ::grpc::RpcMethod rpcmethod_#{methodName}_;|]
+        privateProxyMethodDecl Function{..} = [lt|const ::grpc::internal::RpcMethod rpcmethod_#{methodName}_;|]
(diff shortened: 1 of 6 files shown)


ara-ayvazyan:
Please run `$ stack test --test-arguments --accept` from the `compiler/tests` folder in order to update the `gbc` test files under `compiler/tests/generated`.

compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
```diff
@@ -263,11 +263,11 @@ inline #{className}::#{proxyName}<TThreadPool>::#{proxyName}(
             Async#{methodName}(::std::make_shared< ::grpc::ClientContext>(), #{bonded methodInput}{request});
         }|]
 
-        privateProxyMethodDecl Function{..} = [lt|const ::grpc::RpcMethod rpcmethod_#{methodName}_;|]
-        privateProxyMethodDecl Event{..} = [lt|const ::grpc::RpcMethod rpcmethod_#{methodName}_;|]
+        privateProxyMethodDecl Function{..} = [lt|const ::grpc::internal::RpcMethod rpcmethod_#{methodName}_;|]
```
ara-ayvazyan: Please run `$ stack test --test-arguments --accept` from the `compiler/tests` folder in order to update the `gbc` test files under `compiler/tests/generated`.

grisharav: Thanks, just did it, waiting for CI

chwarr: The addition of the `internal` namespace makes sense here. It's an artifact of [gRPC proposal L8, "Separate internal-only and public sections of the C++ wrapping"](https://github.com/grpc/proposal/blob/master/L8-c%2B%2B-internalization.md). The "internal" parts of gRPC++ are not intended to be used by code using clients or implementing services. The internal parts are OK to be used by things like codegen, which gbc is doing.

chwarr:
The Doc build that's run as part of AppVeyor has failed. It looks like the server hosting Doxygen installer binaries is down at the moment. You don't need to worry about this.

Issue #813, opened 2018-03-06
[c++ grpc] Switch to grpcpp include path


chwarr:
After [gRPC proposal L22](https://github.com/grpc/proposal/blob/master/L22-change-grpc%2B%2B-dir-name.md), the include directory for gRPC++ [was changed from "grpc++" to "grpcpp"](0e00c430827e81d61e1e7164ef04ca21ccbfaa77). This is part of the v1.10.0 release, which #811 is upgrading us to.

After this upgrade, Bond code that uses gRPC++ should be changed to use "grpcpp" over "grpc++" to track the upstream project's changes.

PR #815, opened 2018-03-06
[c++ grpc] Switch to grpcpp include path

chwarr:
The gRPC++ headers were moved from grpc++ to grpcpp as part of [gRPC
proposal L22, "gRPC C++ Public Header Directory Change"][1].

Fixes https://github.com/Microsoft/bond/issues/813

[1]: https://github.com/grpc/proposal/blob/master/L22-change-grpc%2B%2B-dir-name.md

### DIFF
--- a/compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
+++ b/compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
@@ -49,11 +49,11 @@ grpc_h export_attribute cpp file imports declarations = ("_grpc.h", [lt|
 #pragma warning (disable: 4100 4267)
 #endif
 
-#include <grpc++/impl/codegen/channel_interface.h>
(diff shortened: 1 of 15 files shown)


compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
```diff
@@ -49,11 +49,11 @@ grpc_h export_attribute cpp file imports declarations = ("_grpc.h", [lt|
 #pragma warning (disable: 4100 4267)
 #endif
 
-#include <grpc++/impl/codegen/channel_interface.h>
```
ara-ayvazyan: Missing updated golden test files for `gbc`.

chwarr: Thanks. Fixed in 0544f7a7.

Issue #812, opened 2018-03-06
[c# grpc] Upgrade to gRPC 1.10.0


chwarr:
In #811, C++ is being upgraded to consume gRPC v1.10.0, as there are some memory leak fixes. We should also upgrade the C# version as well, following our policy of keeping the two in sync.

PR #817, opened 2018-03-06
[c# grpc] Upgrade to gRPC v1.10.0

chwarr:
Fixes https://github.com/Microsoft/bond/issues/812

### DIFF
--- a/cs/dnc/src/grpc/project.json
+++ b/cs/dnc/src/grpc/project.json
@@ -36,7 +36,7 @@
         "attributes": "1.0.0-*",
         "core": "1.0.0-*",
         "Grpc.Core": {
-            "version": "1.1.0"
+            "version": "1.10.0"
         },
         "io": "1.0.0-*",
         "reflection": "1.0.0-*"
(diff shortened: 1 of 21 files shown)


ara-ayvazyan:
Looks like this requires an upgrade to NETStandard.Library 1.6.**1**.

chwarr:
Upgrading shouldn't be an issue. NETStandardLibrary 1.6.1 is bug fixes for 1.6, not new APIs or anything.
https://github.com/dotnet/standard/issues/171

cs/dnc/src/grpc/project.json
```diff
@@ -44,7 +44,7 @@
     "frameworks": {
         "netstandard1.6": {
             "dependencies": {
-                "NETStandard.Library": "1.6.0"
+                "NETStandard.Library": "1.6.1"
```
ara-ayvazyan: I think I saw more places with this dependency. Shouldn't we upgrade them all?

chwarr: No, that shouldn't be needed. Only Bond.Grpc needs this bumped dependency. The rest are fine with 1.6.0 until we get a bug report about behavior that's fixed in 1.6.1. The dependency solver should pick 1.6.1 in a project consuming Bond.Grpc, even if all the rest of the assemblies say they need 1.6.0.

PR #818, opened 2018-03-06
[c++] Fix bond::value's move ctor

ara-ayvazyan:
The change fixes the `bond::value`'s move constructor to actually move the underlying reader, unless it is a reference type. Also, fixes the `noexcept` condition for `bond::bonded`'s move constructor that was missed in [5131a950](https://github.com/Microsoft/bond/commit/5131a95042ade4881c5fa700b5cf683b4925142d).


CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# Bond Changelog #
```
ara-ayvazyan: Oops.

PR #819, opened 2018-03-07
travis: Pull down git tags before starting the build.

tstein:
Without this, Java will automatically configure a version string equal
to the checked-out hash prefix, which has a 6/16 chance of starting with
a letter instead of a number. Gradle doesn't seem to resolve
dependencies properly when this happens.

### DIFF
--- a/tools/ci-scripts/linux/build_java.zsh
+++ b/tools/ci-scripts/linux/build_java.zsh
@@ -2,6 +2,13 @@
 
 set -eux
 
+# Java needs git tags to detect the version it's building. If we haven't
+# released in a while, the last tag might be many commits back, so we'll fetch
+# the entire history here.
+pushd $BOND_ROOT
+git fetch --unshallow --tags
+popd
+
 # Install gbc.
 cmake \
     -DBOND_ENABLE_JAVA=TRUE \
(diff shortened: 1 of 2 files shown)


chwarr:
The merge presence of the tags doesn't seem to have changed the version generated:
https://travis-ci.org/Microsoft/bond/jobs/350096239

tools/ci-scripts/linux/build_java.zsh
```diff
@@ -2,6 +2,11 @@
 
 set -eux
 
+# Java needs git tags to detect the version it's building. If we haven't
+# released in a while, the last tag might be many commits back, so we'll fetch
+# the entire history here.
+git fetch --unshallow --tags
```
chwarr: Do we need to `cd "$BOND_ROOT"` first? IIRC, we'll be in `$BUILD_ROOT` here. We'll see what happens with the build. :-)

tools/ci-scripts/linux/build_java.zsh
```diff
@@ -2,6 +2,13 @@
 
 set -eux
 
+# Java needs git tags to detect the version it's building. If we haven't
+# released in a while, the last tag might be many commits back, so we'll fetch
+# the entire history here.
+pushd ~/bond
```
chwarr: `pushd "$BOND_ROOT"` ?

PR #822, opened 2018-03-08
[c#] Handle gbc from path on macOS

chwarr:
On macOS (and likely Linux), when gbc was not found under
BOND_COMPILER_DIRECTORY--typically when it was found on $PATH--the
output directory for the generated C# files ended up having inconsistent
directory separator translation applied to it.

gbc was invoked with an output directory like
`--output-dir="obj/Debug/net45/\."`, which placed the generated files in
a directory *under* `net45` named `\.`. However, when csc was invoked,
it was given generated file paths like `obj/Debug/net45//foo_types.cs`,
which did not exist, causing compilation to fail.

This inconsistency occurred because when the command line for gbc from
$PATH was constructed, none of the elements look like a path, so MSBuild
didn't translate the trailing `\.` added to the output-dir argument:

    <_BondCommand>&quot;$(_BondExe)&quot; $(BondCodegenMode) $(_BondImportDirs) --jobs=-2 --namespace=bond=Bond --output-dir="$(BondOutputDirectory)\."</_BondCommand>

The `\.` was added to guard against $(BondOutputDirectory) ending with a
slash and then having CMD treat a string like
`--output-dir="obj\Debug\net45\"` as having an escaped double-quote in
it.

When gbc came from BOND_COMPILER_PATH, it looked like a path, so the
added `\.` got translated to `/.`

As a fix, the codegen targets now use $(_BondOutputDirectoryNormalized),
which always has a trailing slash and always uses forward slashes.
Windows, macOS, and Linux can all handle forward slashes for paths.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,9 @@ different versioning scheme, following the Haskell community's
     * This version include a number of memory leak fixes that users of Bond-over-gRPC were encountering. [Issue #810](https://github.com/Microsoft/bond/issues/810)
 * `BondCodegen` items will now appear in the Visual Studio 2017+ UI in .NET
   Core projects.
+* Fixed a bug in the codegen targets when using `gbc` from $PATH on macOS
+  and Linux that prevented the C# compiler from finding the generated C#
+  files.
 
 ## 7.0.2: 2017-10-30 ##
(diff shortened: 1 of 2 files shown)


lalo:
Merged with eca53e2278c6ef350bbf687fd4a9a248e450d9df

PR #823, opened 2018-03-09
[c#] Automatically find and run C# examples

chwarr:
* Comm examples are excluded, as Comm will be deleted soon.
* GRPC shared-type-assembly examples are excluded, as they require
  starting two processing in a specific order and console input.


chwarr:
To see what gets output when one of the examples fails, take a look [at this AppVeyor build](https://ci.appveyor.com/project/MicrosoftBond/bond/build/1.0.1271).

chwarr:
@lalo  @tstein : ready for review
This should either be rebased or merged. It should not be squashed.

PR #825, opened 2018-03-10
[c#] Fully strong-name sign .NET Core assemblies

chwarr:
Previously, we were just [public signing][1] the .NET Core assemblies.
We intended to fully strong-name sign them, so this commit removes the
"publicSign": true entry from the project files.

[1]: https://github.com/dotnet/roslyn/commit/ca61bad288c1d2e216281680fe92ff55d7122e7f


chwarr:
Someone else could do either. Both the public and private keys are in the bond.snk file that's in this repository. To get the old publicSign behavior, they'd have to re-introduce the option to the project file. This could be useful if, say, one wanted to build a strong-name signed Bond assembly using .NET Core on macOS or Linux. Since we build these on Windows, we can perform full strong-name signing.

Why were we using publicSign=true in the past? I misread the documentation and though it was needed to do any signing at all.

PR #826, opened 2018-03-14
[gbc] Add debug/release Haskell configurations

chwarr:
We were always building gbc with GHC optimizations turned on. We can get
faster build times for things like CI if we don't build with optimizations.

To do that, we now pass different stack & gbc options for the various
CMake build types. The MSBuild Compiler.csproj project translates
MSBuild configurations into CMake configurations.


chwarr:
Looks like there's not much affect on build times. The Haskell build may be faster with this, but its pretty noisy.

chwarr:
Yes: I need it so that I can automate the MinSizeRel build for NuGet package releases.

PR #832, opened 2018-03-22
[c++ travis] Split C++ build into its own scripts

chwarr:
Splits the C++ build into its own set of scripts, like the other
flavors. This is in preperation for adding another C++ flavor: one that
builds with the gRPC master code instead of the current submodule
version.

### DIFF
--- a/tools/ci-scripts/linux/build_cpp-common.zsh
+++ b/tools/ci-scripts/linux/build_cpp-common.zsh
@@ -1,0 +1,15 @@
+#!/bin/zsh
+
+set -eux
+
+if [ ! $BOOST ]; then echo "BOOST not specified"; exit 1; fi
+
+export BOOST_ROOT=/opt/boosts/boost_`echo $BOOST | tr . _`
+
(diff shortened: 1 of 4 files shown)


chwarr:
I'm using this PR to test the Travis CI integration. Please do not review yet.

chwarr:
@ara-ayvazyan, @tstein: ready for review

tools/ci-scripts/linux/build_cpp-common.zsh
```diff
@@ -0,0 +1,15 @@
+#!/bin/zsh
+
+set -eux
+
+if [ ! $BOOST ]; then echo "BOOST not specified"; exit 1; fi
+
+export BOOST_ROOT=/opt/boosts/boost_`echo $BOOST | tr . _`
```
ara-ayvazyan: Should we set the default value of `$BOOST` to be `1_63_0` and move this to `build.zsh`, instead of having a default value for `$BOOST_ROOT`?

chwarr: ✔️

PR #834, opened 2018-03-22
Build gRPC++ using latest gRPC master

chwarr:
This build flavor builds Bond with the latest gRPC master branch so we
can detect potential breaks faster than when we upgrade our submodule.

This build flavor isn't needed on every commit or pull request, so it's
only enabled for Travis's cron builds.

### DIFF
--- a/tools/ci-scripts/linux/build_cpp-grpc-master.zsh
+++ b/tools/ci-scripts/linux/build_cpp-grpc-master.zsh
@@ -3,13 +3,12 @@
 set -eux
 
 # Get gRPC's current master
-cd "$BOND_ROOT/thirdparty/grpc"
+pushd "$BOND_ROOT/thirdparty/grpc"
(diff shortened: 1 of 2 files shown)


chwarr:
FYI: @vjpai & @jtattermusch

tools/ci-scripts/linux/build_cpp-grpc-master.zsh
```diff
@@ -0,0 +1,15 @@
+#!/bin/zsh
+
+set -eux
+
+# Get gRPC's current master
+cd "$BOND_ROOT/thirdparty/grpc"
```
ara-ayvazyan: `pushd`?

chwarr: Yeah, that's more maintainable.

PR #836, opened 2018-03-22
[travis] Refresh Travis CI image

chwarr:
This uses an image built on 2018-03-22.


chwarr:
Looks like we didn't merge #814 soon enough and the image it was using (build-10674409) got purged.

10674409

Issue #771, opened 2018-01-09
[c++] Dropping support for versions of Boost more than 2 years old


chwarr:
In preparation for adding CI testing for [Boost 1.66](http://www.boost.org/users/history/version_1_66_0.html) (release December 18, 2017), we're thinking about dropping support for versions of Boost older than 2 years. This would mean that we only test with Boost 1.61 and above.

Over time, as functionality only available in later versions of Boost is used, Bond would stop building with older versions of Boost.

Please add a comment to this issue with your thoughts and concerns.

jasonzio:
Significant changes to Boost Test in 1.56 make Boost 1.55 really sticky.

chwarr:
Good point, @jasonzio. Thanks.

(And, we've had to do the Boost Test migration ourselves. :-))

One thing we're seeing in the opposite direction is the removal of things like `std::auto_ptr` from the standard library is pulling people up to Boost 1.65, the earliest version that excised its use of `std::auto_ptr`.

PR #838, opened 2018-03-23
[travis] Test Boost 1.61+ with clang & g++

chwarr:
Drops support for versions of Boost older than two years. Boost 1.61+ or
newer is now required. (Bond will likely continue to work with older
versions of Boost, but we'll no longer actively test against them.)

Using cron jobs, add coverage for all supported versions of Boost with
both Clang and G++.

This is the Travis-side fix for:
* https://github.com/Microsoft/bond/issues/666
* https://github.com/Microsoft/bond/issues/771

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Bond C++ library requires some C++11 features (currently limited to those
 supported bv Visual C++ 2013); a C++11 compiler is required. Additionally,
 to build Bond you will need CMake (3.1+),
 [Haskell Stack](https://docs.haskellstack.org/en/stable/README/#how-to-install)
-(1.5.1+) and Boost (1.58+).
+(1.5.1+) and Boost (1.61+).
 
 Following are specific instructions for building on various platforms.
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -40,7 +40,7 @@ Bond C++ library requires some C++11 features (currently limited to those
 supported bv Visual C++ 2013); a C++11 compiler is required. Additionally,
 to build Bond you will need CMake (3.1+),
 [Haskell Stack](https://docs.haskellstack.org/en/stable/README/#how-to-install)
-(1.5.1+) and Boost (1.58+).
+(1.5.1+) and Boost (1.61+).
```
ara-ayvazyan: Update the minimum version check in [cmake/config.cmake](https://github.com/Microsoft/bond/blob/master/cmake/Config.cmake#L126)?

ara-ayvazyan:
The #666 has a checkbox for Appveyor as well.

chwarr:
The CMake change will need to happen after AppVeyor has been updated. To unblock this PR, I've opened [an issue to bump the Boost version in CMake](https://github.com/Microsoft/bond/issues/845) and backed out the CMake change in this PR.

PR #839, opened 2018-03-23
WIP: [appveyor] Test Boost 1.62+ only

chwarr:
This is part 1 of 2 in the process of dropping support for versions of
Boost older than two years. Boost 1.61+ or newer is now required. (Bond
will likely continue to work with older versions of Boost, but we'll no
longer actively test against them.)

We'd like to test with the current lowest version of Boost that Bond
supports, Boost 1.61. However, there is no Boost 1.61 available on
AppVeyor, so we start with the lowest version that is available in
AppVeyor and supported by Bond: 1.62.

This AppVeyor change needs to be made before the CMake minimum version
bump that we're planning to make with the Travis CI and README changes.

This is the AppVeyor-side change for
https://github.com/Microsoft/bond/issues/771

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -43,14 +43,14 @@
               BOND_VS: "Visual Studio 12 2013"
               BOND_VS_NUM: 12
               BOND_ARCH: 32
-              BOND_BOOST: 58
+              BOND_BOOST: 1.62.0
               BOND_CMAKE_FLAGS: "-DBOND_ENABLE_GRPC=FALSE;-DBOND_SKIP_GBC_TESTS=TRUE"
               # C++ Core build and tests
             - BOND_BUILD: C++
               BOND_VS: "Visual Studio 14 2015"
               BOND_VS_NUM: 14
               BOND_ARCH: 64
-              BOND_BOOST: 60
+              BOND_BOOST: 1.62.0
               BOND_CMAKE_FLAGS: "-DBOND_ENABLE_GRPC=FALSE"
             - BOND_BUILD: C#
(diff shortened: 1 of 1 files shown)


ara-ayvazyan:
Should we remove the check at [appveyor.yml#L97](https://github.com/Microsoft/bond/blob/master/appveyor.yml#L97)?

appveyor.yml
```diff
@@ -43,14 +43,14 @@
               BOND_VS: "Visual Studio 12 2013"
               BOND_VS_NUM: 12
               BOND_ARCH: 32
-              BOND_BOOST: 58
+              BOND_BOOST: 61
```
ara-ayvazyan: Appveyor does not seem to have 1.61 according to https://www.appveyor.com/docs/build-environment/#boost.

ara-ayvazyan: Appveyor does not seem to have 1.61 according to https://www.appveyor.com/docs/build-environment/#boost.

chwarr:
I did a force push, as I needed to update the commit message.

Changes since most recent reviews:

* Rewrote Boost selection to be more table-driven.
* Switched to 1.62 as there is no 1.61.

chwarr:
I need to so some debugging of this...

chwarr:
AppVeyor only has Boost 1.56 and 1.58 binaries for MSVC12...

ara-ayvazyan:
https://www.nuget.org/packages/boost-vc120

chwarr:
Closing this PR. I'll need to rework how we build with AppVeyor and MSVC 2013 and 2017...

PR #840, opened 2018-03-23
[travis] Skip install step to speed up build

chwarr:
Since we bring our own build dependencies in the Docker images we use,
we don't need Travis to install anything other than Docker. It is
currently install a Ruby toolchain; this can be skipped.


chwarr:
This doesn't actually skip the Ruby setup step like I though. There's no reason to make this change, so I'm going to recall this PR.

PR #842, opened 2018-03-30
[c++ grpc] Avoid data coping during serialization

ara-ayvazyan:


cpp/inc/bond/core/blob.h
```diff
@@ -248,6 +248,41 @@ class blob
         return _content + _length;
     }
 
+    /// @brief Returns a blob with same data that owns the memory
+    blob own() const
```
ara-ayvazyan: Changing this to a free function will be better.

CHANGELOG.md
```diff
@@ -69,6 +69,9 @@ different versioning scheme, following the Haskell community's
 * Fixed a race condition during `bond::ext::gRPC::unary_call` destruction.
 * Fixed the broken move constructor of `bond::bonded<T, Reader&>`.
 * Fixed the move constructor of `bond::value` to actually move the underlying reader.
+* Added `bond::blob_own` helper function that will return a `bond::blob` with
+  a copied data if the original one does own the memory.
```
ara-ayvazyan: `does not`

cpp/inc/bond/core/blob.h
```diff
@@ -397,5 +400,25 @@ inline T blob_cast(const blob& from)
     }
 }
 
+/// @brief Returns a blob with same data that owns the memory
```
chwarr: Consider talking about when this will and will not make a copy.

cpp/inc/bond/core/blob.h
```diff
@@ -252,6 +252,9 @@ class blob
     template <typename T>
     friend T blob_cast(const blob& from);
 
+    template <typename A>
+    friend blob blob_own(blob src, const A& allocator);
```
chwarr: `blob_prolong`?

cpp/inc/bond/ext/grpc/bond_utils.h
```diff
@@ -14,6 +14,9 @@
 
 #include <boost/assert.hpp>
 #include <boost/make_shared.hpp>
+#include <boost/container/small_vector.hpp>
+#include <boost/smart_ptr/intrusive_ref_counter.hpp>
+#include <boost/intrusive_ptr.hpp>
```
chwarr: Minor: alpha sort headers

cpp/inc/bond/ext/grpc/bond_utils.h
```diff
@@ -22,6 +25,35 @@
...
+
+        boost::container::small_vector<grpc::Slice, 8> slices;
+        slices.reserve(buffers->size());
+
+        for (blob& data : *buffers)
+        {
+            data = blob_own(std::move(data));
+
+            slices.emplace_back(
+                const_cast<void*>(data.data()),
```
chwarr: Can there be a comment for why the const cast? (E.g., to adhere to the slice API?)

cpp/inc/bond/stream/output_buffer.h
```diff
@@ -144,9 +144,9 @@ class OutputMemoryStream
     }
 
 
-    /// @brief Get content of the stream as a vector of memory blobs
+    /// @brief Get content of the stream as a collection of memory blobs
```
chwarr: Consider saying that the container needs to support `emplace_back` of `blob`.

cpp/inc/bond/ext/grpc/bond_utils.h
ara-ayvazyan: The C++ API does not need `const_cast` actually. Left over code from older C API use.

cpp/inc/bond/core/blob.h
```diff
@@ -400,9 +400,10 @@ inline T blob_cast(const blob& from)
     }
 }
 
-/// @brief Returns a blob with same data that owns the memory
+/// @brief Returns a blob with a copied data if the original one does not own the memory
```
chwarr: Minor: "with a copy of the data" or "with copied data"

cpp/inc/bond/core/blob.h
```diff
@@ -400,9 +400,10 @@ inline T blob_cast(const blob& from)
...
+/// (i.e. constructed using raw memory), and the same one otherwise.
```
chwarr: Minor: "and the same blob otherwise"

cpp/inc/bond/ext/grpc/bond_utils.h
```diff
@@ -40,10 +40,10 @@ namespace bond { namespace ext { namespace gRPC { namespace detail
 
         for (blob& data : *buffers)
         {
-            data = blob_own(std::move(data));
+            data = blob_prolong(std::move(data));
 
             slices.emplace_back(
-                const_cast<void*>(data.data()),
+                const_cast<void*>(data.data()), // The buffer is not expected to be modified.
```
chwarr: Minor: "modified, but we have to const_cast<void*> because the grpc::Slice ctor only takes void*"
This way we explain why the cast is needed too without requiring the reading to have memoryized the grpc::Slice ctor overloads.

PR #844, opened 2018-04-05
[c++ c#] Remove lingering references to Bond Comm

chwarr:
There were a few lingering references to Bond Comm in the C++ and C#
manuals. They've now been deleted.


ara-ayvazyan:
You may want to update the [bond_over_grpc.md](https://github.com/Microsoft/bond/blob/master/doc/src/bond_over_grpc.md), it mentions about comm.

PR #848, opened 2018-04-07
[c++] Use a "variant" for SimpleJsonReader state

chwarr:
A SimpleJsonReader either has an input buffer and an associated stream
or it is a child of some other SimpleJsonReader that will use its
parent's already parsed document. This logic is now more explicit:

* A SimpleJsonReader now either has an input stream XOR a parent
  pointer. This logic is implemented in the helper StreamHolder class.
* detail::RapidJsonInputStream now owns its input buffer instead of
  having a reference to its input buffer.
    * Before, when child SimpleJsonReader instances had their own
      detail::RapidJsonInputStream instances, we needed to be able to
      make cheap copies of detail::RapidJsonInputStream. But, the
      children didn't actually use their stream except to sometimes make
      a deep copy. Since they now just have a pointer to their parent
      which can be used to make a deep copy,
      detail::RapidJsonInputStream can own its buffer: we only copy
      detail::RapidJsonInputStream instances when we need a deep copy.

Also, minor refactoring like the removal of boost::call_traits in favor
of using rvalue references where appropriate.

### DIFF
--- a/cpp/inc/bond/protocol/detail/rapidjson_helper.h
+++ b/cpp/inc/bond/protocol/detail/rapidjson_helper.h
@@ -62,73 +62,72 @@ class RapidJsonInputStream
 public:
     typedef char Ch;
 
-    explicit RapidJsonInputStream(const Buffer& buf)
-        : input(buf),
-          current(0),
-          count(0)
+    explicit RapidJsonInputStream(const Buffer& input)
+        : _input(input),
+          _current(0),
+          _count(0)
     {
-        input.Read(current);
+        _input.Read(_current);
     }
 
+    #if defined(_MSC_VER) && _MSC_VER < 1900
+    // since we explicitly implement a move ctor, we need to explicitly
+    // default these
     RapidJsonInputStream(const RapidJsonInputStream&) = default;
     RapidJsonInputStream& operator=(const RapidJsonInputStream&) = default;
 
-    #if defined(_MSC_VER) && _MSC_VER < 1900
     // MSVC cannot = default rvalue ctor or move-assign operators
     RapidJsonInputStream(RapidJsonInputStream&& other)
-        : input(std::move(other.input)),
-          current(std::move(other.current)),
-          count(std::move(other.count))
+        : _input(std::move(other._input)),
+          _current(std::move(other._current)),
+          _count(std::move(other._count))
     { }
 
     RapidJsonInputStream& operator=(RapidJsonInputStream&& other)
     {
-        input = std::move(other.input);
-        current = std::move(other.current);
-        count = std::move(other.count);
+        _input = std::move(other._input);
+        _current = std::move(other._current);
+        _count = std::move(other._count);
         return *this;
     }
-    #else
-    RapidJsonInputStream(RapidJsonInputStream&&) = default;
-    RapidJsonInputStream& operator=(RapidJsonInputStream&&) = default;
     #endif
 
     const Buffer& GetBuffer() const
     {
-        return input;
+        return _input;
     }
 
     Buffer& GetBuffer()
     {
-        return input;
+        return _input;
     }
 
     char Peek()
     {
-        if (!current)
+        if (!_current)
         {
-            input.Read(current);
+            _input.Read(_current);
         }
 
-        return current;
+        return _current;
     }
 
     size_t Tell() const
     {
-        return count;
+        return _count;
     }
 
     char Take()
     {
-        char c = current;
-        current = '\0';
+        char c = _current;
+        _current = '\0';
 
         if (!c)
         {
-            input.Read(c);
+            _input.Read(c);
         }
 
-        ++count;
+        ++_count;
         return c;
     }
 
@@ -138,9 +137,9 @@ public:
     size_t PutEnd(char*) { BOOST_ASSERT(false); return 0; }
 
 private:
-    Buffer input;
-    uint8_t current;
-    size_t count;
+    Buffer _input;
+    uint8_t _current;
+    size_t _count;
 };
 
 
@@ -149,8 +148,8 @@ template <typename Buffer>
 class RapidJsonOutputStream
 {
 public:
-    RapidJsonOutputStream(Buffer& output)
-        : output(output)
+    explicit RapidJsonOutputStream(Buffer& output)
+        : _output(output)
     {
     }
 
@@ -165,7 +164,7 @@ public:
 
     void Put(char c)
     {
-        output.Write(c);
+        _output.Write(c);
     }
 
     void Flush()
@@ -173,7 +172,7 @@ public:
     }
 
 private:
-    Buffer& output;
+    Buffer& _output;
 };
 
 
@@ -181,7 +180,7 @@ private:
 template <>
 struct RapidJsonInputStream<const rapidjson::UTF8<>::Ch*> : rapidjson::StringStream
 {
-    RapidJsonInputStream(const char* buffer)
+    explicit RapidJsonInputStream(const char* buffer)
         : rapidjson::StringStream(buffer)
     {}
 
--- a/cpp/inc/bond/protocol/detail/rapidjson_helper.h
+++ b/cpp/inc/bond/protocol/detail/rapidjson_helper.h
@@ -45,7 +45,6 @@
 
 #include "rapidjson/writer.h"
 
-#include <boost/call_traits.hpp>
 #include <boost/noncopyable.hpp>
 
 #include <algorithm>
@@ -63,24 +62,53 @@ class RapidJsonInputStream
 public:
     typedef char Ch;
 
-    RapidJsonInputStream(typename boost::call_traits<Buffer>::reference input)
-        : input(&input),
+    explicit RapidJsonInputStream(const Buffer& buf)
+        : input(buf),
           current(0),
           count(0)
     {
         input.Read(current);
     }
 
-    RapidJsonInputStream(const RapidJsonInputStream& that, typename boost::call_traits<Buffer>::reference input)
-        : input(&input),
-          current(that.current),
-          count(that.count)
-    {}
+    RapidJsonInputStream(const RapidJsonInputStream&) = default;
+    RapidJsonInputStream& operator=(const RapidJsonInputStream&) = default;
+
+    #if defined(_MSC_VER) && _MSC_VER < 1900
+    // MSVC cannot = default rvalue ctor or move-assign operators
+    RapidJsonInputStream(RapidJsonInputStream&& other)
+        : input(std::move(other.input)),
+          current(std::move(other.current)),
+          count(std::move(other.count))
+    { }
+
+    RapidJsonInputStream& operator=(RapidJsonInputStream&& other)
+    {
+        input = std::move(other.input);
+        current = std::move(other.current);
+        count = std::move(other.count);
+        return *this;
+    }
+    #else
+    RapidJsonInputStream(RapidJsonInputStream&&) = default;
+    RapidJsonInputStream& operator=(RapidJsonInputStream&&) = default;
+    #endif
+
+    const Buffer& GetBuffer() const
+    {
+        return input;
+    }
+
+    Buffer& GetBuffer()
+    {
+        return input;
+    }
 
     char Peek()
     {
         if (!current)
-            input->Read(current);
+        {
+            input.Read(current);
+        }
 
         return current;
     }
@@ -96,7 +124,9 @@ public:
         current = '\0';
 
         if (!c)
-            input->Read(c);
+        {
+            input.Read(c);
+        }
 
         ++count;
         return c;
@@ -107,18 +137,8 @@ public:
     void Put(char) { BOOST_ASSERT(false); }
     size_t PutEnd(char*) { BOOST_ASSERT(false); return 0; }
 
-    RapidJsonInputStream& operator=(const RapidJsonInputStream& that)
-    {
-        // rapidjson reader makes a local copy of stream within some functions
-        // and assigns it back to its main stream variable before function exit.
-        BOOST_ASSERT(input == that.input);
-        current = that.current;
-        count = that.count;
-        return *this;
-    }
-
 private:
-    typename std::remove_reference<Buffer>::type* input;
+    Buffer input;
     uint8_t current;
     size_t count;
 };
@@ -129,7 +149,7 @@ template <typename Buffer>
 class RapidJsonOutputStream
 {
 public:
-    RapidJsonOutputStream(typename boost::call_traits<Buffer>::reference output)
+    RapidJsonOutputStream(Buffer& output)
         : output(output)
     {
     }
--- a/cpp/inc/bond/protocol/simple_json_reader.h
+++ b/cpp/inc/bond/protocol/simple_json_reader.h
@@ -227,7 +227,7 @@ private:
         // MSVC cannot = default rvalue ctor or move-assign operators
         StreamHolder(StreamHolder&& other)
             : _stream(std::move(other._stream)),
-             _parent(std::move(other._parent))
+              _parent(std::move(other._parent))
         { }
 
         StreamHolder& operator=(StreamHolder&& other)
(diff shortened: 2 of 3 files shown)


cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -39,11 +39,12 @@ class SimpleJsonReader
...
           _stream(_input),
           _document(boost::make_shared<rapidjson::Document>()),
-          _value(NULL)
+          _value(nullptr)
     {}
 
     SimpleJsonReader(const SimpleJsonReader& that, const Field& value)
-        : _stream(that._stream),
+        : _input(that._input),
+          _stream(that._stream),
```
ara-ayvazyan: So in this case `_stream.input` will point to a buffer the lifetime of which we do not control?

chwarr: The caller is already required to keep the input buffer alive as long as the stream is alive.
Simple JSON doesn't give you back a reference to a blob: it has to copy and decode the bytes, so the lifetime issues shouldn't be a problem there.

Perhaps Simple JSON just shouldn't expose its underlying buffer at all... Thoughts?

chwarr: Let me try switching this to have a "parent" SimpleJsonReader instead of copying the input and stream.

ara-ayvazyan: Yeah, that sounds better.

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -179,15 +195,190 @@ class SimpleJsonReader
...
+                BOOST_ASSERT(!IsParent());
+                BOOST_ASSERT(IsChild());
+            }
+
+            return *this;
+        }
+
+        bool IsParent() const
+        {
+            return static_cast<bool>(_input) && static_cast<bool>(_stream);
```
ara-ayvazyan: Explicit `static_cast` is not needed I think.

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -179,15 +195,190 @@ class SimpleJsonReader
...
+        }
+
+        bool IsChild() const
```
ara-ayvazyan: It looks weird that the code for `IsChild` does not look like `!IsParent`. Maybe just keep a single `IsParent` and check the invariant only there?

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -179,15 +195,190 @@ class SimpleJsonReader
...
+        }
+
+        StreamHolder* GetParent()
+        {
+            if (IsParent())
+            {
+                return this;
+            }
+
+            BOOST_ASSERT(IsChild());
```
ara-ayvazyan: This also won't be needed if you just keep `IsParent`. Also make this a single-line function. Larger code is harder to read, even if it is simple enough :)

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -179,15 +195,190 @@ class SimpleJsonReader
...
+            else
+            {
+                return _parent->GetStream();
+            }
+        }
+
+    private:
+        // NOTE: the ctors rely on _input being initialized before _stream
+        boost::optional<Buffer> _input;
+        boost::optional<detail::RapidJsonInputStream<Buffer>> _stream;
```
ara-ayvazyan: I think you should keep a single `optional` member. Also, maybe changing the `detail::RapidJsonInputStream` by making it own the `Buffer` instead of referring to it may simplify. I don't see it ever referring anything else but the neighboring `_input`.

chwarr: Hmmm... Let me see.

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -179,15 +195,190 @@ class SimpleJsonReader
...
+            }
+        }
+
+        StreamHolder& operator=(const StreamHolder& that)
+        {
+            _parent = that._parent;
+
+            if (that.IsParent())
+            {
+                _input.emplace(that._input.value());
```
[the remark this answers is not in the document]
chwarr: Not with G++ when building bf, which has a copyable, but not assignable, `InputFile` buffer type.

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -122,39 +133,44 @@ class SimpleJsonReader
...
 
     rapidjson::Value::ConstValueIterator ArrayEnd() const
     {
+        BOOST_ASSERT(_value);
         return _value->IsArray() ? _value->End() : rapidjson::Value::ConstValueIterator();
     }
 
     uint32_t ArraySize() const
     {
+        BOOST_ASSERT(_value);
```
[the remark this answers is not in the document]
chwarr: Yeah, that's cleaner.

ara-ayvazyan:
I think it would be better to update the PR title.

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -179,15 +195,190 @@ class SimpleJsonReader
...
+            : _input(),
+              _stream(),
+              _parent(that._parent)
+        {
+            if (that.IsParent())
+            {
+                _input = std::move(that._input);
+                _stream.emplace(that._stream.value(), _input.value());
+
+                BOOST_ASSERT(IsParent());
```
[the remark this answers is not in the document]
chwarr: This is `if (other was parent) then assert(this is a parent)`
Let me rename `that` to `other`.

cpp/inc/bond/protocol/simple_json_reader.h
chwarr: Good suggestion.

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -13,6 +13,8 @@
 
 #include <boost/call_traits.hpp>
 #include <boost/make_shared.hpp>
+#include <boost/none.hpp>
+#include <boost/optional/optional.hpp>
```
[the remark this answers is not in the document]
chwarr: `boost/optional.hpp` is [an alias](https://www.boost.org/doc/libs/1_61_0/libs/optional/doc/html/boost_optional/reference/header__boost_optional_hpp_.html) for `boost/optional/optional.hpp`, so I skipped the alias and just included the (current) implementation.

cpp/inc/bond/protocol/simple_json_reader.h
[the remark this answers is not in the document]
ara-ayvazyan: Can you check how that was working before? The `operator=` for this reader existed even before and was holding the `Buffer`.

cpp/inc/bond/protocol/simple_json_reader.h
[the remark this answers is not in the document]
ara-ayvazyan: Well, that is why there are aliases, so that you can simply include `<boost/optional.hpp>` :)

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -179,15 +195,190 @@ class SimpleJsonReader
...
+            BOOST_ASSERT(!IsChild());
+        }
+
+        explicit StreamHolder(SimpleJsonReader& parent)
+            : _input(),
+              _stream(),
+              // If we're creating a child of a child, use the "root" parent
+              // as our parent instead of having a bunch of intermediate
+              // children as parent.
+              _parent(parent._streamHolder.GetParent())
```
ara-ayvazyan: Just add `// Resolve the parent` or something like that next to variable. This does not deserve 3 lines of commenting.

chwarr:
I've addressed all the comments save the one about moving the input buffer into the stream. I need to look more deeply at that one yet.

chwarr:
In the latest commit RapidJsonInputStream now owns its input buffer.

cpp/inc/bond/protocol/detail/rapidjson_helper.h
```diff
@@ -63,24 +62,53 @@ class RapidJsonInputStream
 public:
     typedef char Ch;
 
-    RapidJsonInputStream(typename boost::call_traits<Buffer>::reference input)
-        : input(&input),
+    explicit RapidJsonInputStream(const Buffer& buf)
+        : input(buf),
```
ara-ayvazyan: Minor: rename the variables to have `_` prefix since you are updating this file anyways?

cpp/inc/bond/protocol/detail/rapidjson_helper.h
```diff
@@ -63,24 +62,53 @@ class RapidJsonInputStream
...
+        current = std::move(other.current);
+        count = std::move(other.count);
+        return *this;
+    }
+    #else
+    RapidJsonInputStream(RapidJsonInputStream&&) = default;
+    RapidJsonInputStream& operator=(RapidJsonInputStream&&) = default;
+    #endif
+
+    const Buffer& GetBuffer() const
```
ara-ayvazyan: I think you can skip the `const` version, it is not used I suppose(?).

cpp/inc/bond/protocol/detail/rapidjson_helper.h
```diff
@@ -129,7 +149,7 @@ template <typename Buffer>
 class RapidJsonOutputStream
 {
 public:
-    RapidJsonOutputStream(typename boost::call_traits<Buffer>::reference output)
+    RapidJsonOutputStream(Buffer& output)
```
ara-ayvazyan: Missing `explicit`?

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -179,15 +188,88 @@ class SimpleJsonReader
...
+              _parent()
+        {
+            BOOST_ASSERT(IsParent());
+        }
+
+        #if defined(_MSC_VER) && _MSC_VER < 1900
+        // MSVC cannot = default rvalue ctor or move-assign operators
+        StreamHolder(StreamHolder&& other)
+            : _stream(std::move(other._stream)),
+             _parent(std::move(other._parent))
```
ara-ayvazyan: Minor: misaligned.

cpp/inc/bond/protocol/detail/rapidjson_helper.h
```diff
@@ -63,24 +62,53 @@ class RapidJsonInputStream
...
+    RapidJsonInputStream& operator=(RapidJsonInputStream&& other)
+    {
+        input = std::move(other.input);
+        current = std::move(other.current);
+        count = std::move(other.count);
+        return *this;
+    }
+    #else
+    RapidJsonInputStream(RapidJsonInputStream&&) = default;
+    RapidJsonInputStream& operator=(RapidJsonInputStream&&) = default;
```
ara-ayvazyan: I think you can simply skip all 4 `=default` copy/move constructors for non VC12 case.

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -179,15 +188,88 @@ class SimpleJsonReader
...
+            _stream = std::move(other._stream);
+            _parent = std::move(other._parent);
+            return *this;
+        }
+        #else
+        StreamHolder(StreamHolder&&) = default;
+        StreamHolder& operator=(StreamHolder&&) = default ;
+        #endif
+
+        StreamHolder& operator=(const StreamHolder&) = default;
```
ara-ayvazyan: I think the copy assignment should behave same way as copy construction.

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -179,15 +188,88 @@ class SimpleJsonReader
...
+
+        StreamHolder& operator=(StreamHolder&& other)
+        {
+            _stream = std::move(other._stream);
+            _parent = std::move(other._parent);
+            return *this;
+        }
+        #else
+        StreamHolder(StreamHolder&&) = default;
+        StreamHolder& operator=(StreamHolder&&) = default ;
```
ara-ayvazyan: Minor: extra space.

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -179,15 +188,88 @@ class SimpleJsonReader
...
+        #endif
+
+        StreamHolder& operator=(const StreamHolder&) = default;
+
+        const detail::RapidJsonInputStream<Buffer>& GetStream() const
```
ara-ayvazyan: Minor: you can rename to `Get`, the `Stream` part is redundant.

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -179,15 +188,88 @@ class SimpleJsonReader
...
+
+    private:
+        bool IsParent() const
+        {
+            const bool isParent = _parent == nullptr;
+            BOOST_ASSERT(isParent == static_cast<bool>(_stream)); // parents must have a stream
+            return isParent;
+        }
+
+        StreamHolder* GetParent()
```
ara-ayvazyan: Return reference to exclude questions about `nullptr`.

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -35,26 +36,27 @@ class SimpleJsonReader
...
+    }
 
     /// @brief Copy constructor
-    SimpleJsonReader(const SimpleJsonReader& that)
-        : _input(that._input),
-          _stream(that._stream, _input),
-          _document(that._document),
-          _value(that._value)
-    {}
+    SimpleJsonReader(const SimpleJsonReader&) = default;
```
ara-ayvazyan: I think this is not needed.

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -179,15 +188,88 @@ class SimpleJsonReader
...
-          _document(that._document),
-          _value(&(*that._value)[name])
-    {}
+    /// @brief Holds either an input stream XOR a pointer to some parent
+    /// StreamHolder.
+    class StreamHolder
+    {
+    public:
+        explicit StreamHolder(const Buffer& input)
+            : _stream(),
```
ara-ayvazyan: Skip both? ;)

cpp/inc/bond/protocol/simple_json_reader.h
```diff
@@ -179,15 +188,88 @@ class SimpleJsonReader
...
+        }
 
+        boost::optional<detail::RapidJsonInputStream<Buffer>> _stream;
+        StreamHolder* _parent;
+    } _streamHolder;
 
-    Buffer _input;
-    detail::RapidJsonInputStream<Buffer> _stream;
     boost::shared_ptr<rapidjson::Document> _document;
     const rapidjson::Value* _value;
```
ara-ayvazyan: Make the `_value` first member. It is accessed more frequently.

cpp/inc/bond/protocol/simple_json_reader.h
chwarr: No, that makes that static analysis tool that reported the original bugs (in an internal bug tracking system) unhappy.

cpp/inc/bond/protocol/simple_json_reader.h
chwarr: It was just there for Doxygen. The library is inconsistent in documenting what is and isn't copyable.
I've, instead, encoded that a `SimpleJsonReader` is copyable (for `InputBuffer` at least) in the json_tests unit test with `static_assert` tests.

cpp/inc/bond/protocol/simple_json_reader.h
chwarr: ✔️

cpp/inc/bond/protocol/simple_json_reader.h
chwarr: Good catch.
✔️

cpp/inc/bond/protocol/simple_json_reader.h
chwarr: ✔️

cpp/inc/bond/protocol/simple_json_reader.h
chwarr: ✔️

PR #849, opened 2018-04-07
[c++] Overhaul bond::maybe

chwarr:
bond::maybe now has better support for rvalue-references and allocator
propagation.

This implementation also avoids default initializing an instance of T
when creating an maybe that holds nothing

bond::maybe is now implemented in terms of boost::optional. We should be
able to std::optional with minimal effort when Bond requires a minimum
of C++17.

Additional changes:

* Removed a dangerous implicit conversion operator from `bond::maybe<T>`
  to `const T&`. To access a `bond::maybe<T>` value, use
* Added noexcept variants of `bond::maybe<T>::value`.
* Added `bond::maybe<T>::emplace` to construct a maybe's value in place.
* Added various rvalue-reference and allocator-aware constructors and
  assignment operators.
* Added `bond::maybe<T>::operator==(const T&)` to comapre directly to
  instances of `T`.
* Improved test coverage of `set_value()`: it was only effectively
  constructing a new T the first time.
* Added tests for moved-from bond::maybe instances.
* Added tests for allocator propagation.


chwarr:
I'm broken something to do with allocator propagation. Let me figure that out...

chwarr:
The problem is coming from the [`To` transform](https://github.com/Microsoft/bond/blob/31d7393b14d5de2da21412e8a2c7ec070b0e4400/cpp/inc/bond/core/transforms.h#L464-L474) when it's filling in a `maybe` (triggered by `InitRandom` in the `capped_allocator_tests`)

```
    template <typename Protocols, typename V, typename X>
    void AssignToVar(V& var, const X& value) const
    {
        value.template Deserialize<Protocols>(var);
    }


    template <typename Protocols, typename V, typename X>
    void AssignToVar(maybe<V>& var, const X& value) const
    {
        AssignToVar<Protocols>(var.set_value(), value);
    }
```

With this change to delay constructing the value until it is needed, the new code in `set_value()` attempts to call the default ctor:

```
    T& set_value()
    {
        if (is_nothing())
        {
            _impl.emplace();
        }

        return _impl.get();
    }
```

When the T being constructed is using an allocator that cannot be default constructed, we don't have an instance to forward along when we get here.

I need to think more about this...

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
-
-    /// @brief Constructor from a value T
-    explicit
-    maybe(const T& value)
-        : _value(value),
-          _nothing(false)
+    /// @brief Construct in place
+    ///
+    /// @since 8.0.0
+    template<class... Args>
```
ara-ayvazyan: Prefer `typename`.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
+    void emplace(Args&&... args)
```
ara-ayvazyan: Return `T&` (`emplace` should return that)?

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
-        swap(_nothing, that._nothing);
-        swap(_value, that._value);
+        return _value.get();
     }
 
-    /// @brief Check if this object contains nothing
-    bool is_nothing() const
+    /// @brief Convert to constant reference to value
+    /// @throw NothingException if the object contains nothing
+    operator const T&() const
```
ara-ayvazyan: How about making a break change and making this an explicit one? I don't think the implicit one was intentional.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
+    { }
+
+    maybe_common(maybe_common&& that) BOND_NOEXCEPT_IF(std::is_nothrow_move_constructible<boost::optional<T>>::value)
+        : _value(that._value)
     {
-        _nothing = false;
-        return _value;
+        // unlike std::optional/boost::optional, moved-from bond::maybe
+        // instances are guaranteed to be nothing.
+        that._value = boost::none;
```
ara-ayvazyan: `that._value = {};`

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
+        : _value(value)
+    { }
+
+    template <typename Alloc>
+    maybe_common(const T& value, const Alloc& alloc)
+        : _value(boost::in_place_init, value, alloc)
+    { }
+
+    maybe_common(maybe_common&& that) BOND_NOEXCEPT_IF(std::is_nothrow_move_constructible<boost::optional<T>>::value)
+        : _value(that._value)
```
ara-ayvazyan: Missing `std::move`.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
-        _nothing = false;
-        return _value;
+        // unlike std::optional/boost::optional, moved-from bond::maybe
+        // instances are guaranteed to be nothing.
+        that._value = boost::none;
     }
 
-    /// @brief Set to nothing
-    void set_nothing()
+    boost::optional<T> _value = {};
```
ara-ayvazyan: I think the explicit initialization is redundant, or at least the `=` part :).

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
+};
+
+} // namespace detail
+
+template <typename T,
+    typename enabled = void>
```
ara-ayvazyan: `Enabled` and on same line?

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
+template <typename T>
+class maybe<T, typename boost::disable_if<detail::has_allocator<T>>::type>
+    : public detail::maybe_common<T>
+{
+public:
+    maybe() = default;
+    maybe(const maybe& that) = default;
+
+    maybe(maybe&& that) BOND_NOEXCEPT_IF(std::is_nothrow_move_constructible<typename detail::maybe_common<T>>::value)
+        : detail::maybe_common<T>(that)
```
ara-ayvazyan: Missing `std::move`.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
+{
+public:
+    using allocator_type = typename detail::allocator_type<T>::type;
+
+    maybe() = default;
+    maybe(const maybe&) = default;
+
+// TODO: specialize std::uses_allocator with conditional on has_allocator && then is_convertable
+
+    maybe(maybe&& that) BOND_NOEXCEPT_IF(std::is_nothrow_move_constructible<typename detail::maybe_common<T>>::value
```
ara-ayvazyan: I think you should not need the `typename`.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
+                                         && std::is_nothrow_move_constructible<allocator_type>::value)
+        : detail::maybe_common<T>(static_cast<detail::maybe_common<T>&&>(that)),
+          allocator_type(static_cast<allocator_type&&>(that))
```
ara-ayvazyan: Maybe better to provide private `base()` and `allocator()` functions and use `std::move`? Explicit `static_cast` to `&&` looks error-prone.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
+    { }
+
+    /// @brief Move constructor
+    maybe(
+        maybe&& that,
```
ara-ayvazyan: Keep args on same line?

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
+        const allocator_type& alloc)
+        BOND_NOEXCEPT_IF(std::is_nothrow_move_constructible<typename detail::maybe_common<T>>::value
+                         && std::is_nothrow_copy_constructible<allocator_type>::value)
+        : detail::maybe_common<T>(static_cast<detail::maybe_common<T>&&>(that)),
+          allocator_type(alloc)
+    { }
+
+    explicit maybe(const T& value)
+        : detail::maybe_common<T>(value),
+          allocator_type(allocator_type { })
```
ara-ayvazyan: Mm, just `allocator_type()` (or `{}`) please :)

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
+    { }
+
+    explicit maybe(const allocator_type& alloc)
+        : detail::maybe_common<T>(),
+          allocator_type(alloc)
+    { }
+
+    maybe(const T& value, const allocator_type& alloc)
```
ara-ayvazyan: Missing similar move ctors with allocator?

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
+    bool operator!=(const maybe& that) const
+    {
+        return !(*this == that);
+    }
+};
 
-template<typename T>
-inline void swap(maybe<T>& x, maybe<T>& y)
+template<typename T, typename enabled>
+inline void swap(maybe<T, enabled>& x, maybe<T, enabled>& y)
```
ara-ayvazyan: I think there is no need to mention the `enabled` here. We are not going to use it with anything else but the default (`void`) one.

cpp/test/core/maybe_tests.cpp
```diff
@@ -69,20 +73,27 @@ TEST_CASE_BEGIN(MaybeInterface)
...
     UT_AssertIsFalse(map.is_nothing());
-    UT_AssertIsFalse(map.value().empty());
+    UT_AssertIsTrue(map.value().size() == 2);
 
     std::map<double, std::string> map2;
 
     std::swap(map2, map.value());
     std::swap(map.value(), map2);
 
+    const int numbers[] = { 1, 2, 3, 4, 5 };
```
ara-ayvazyan: I think you do not need the `[]`, just include the `<initialization_list>`.

cpp/test/core/maybe_tests.cpp
```diff
@@ -107,6 +118,18 @@ TEST_CASE_BEGIN(MaybeInterface)
 }
 TEST_CASE_END
 
+TEST_CASE_BEGIN(MovedFromIsNothing)
+{
+    bond::maybe<bool> m(false);
+    UT_AssertIsFalse(m.is_nothing());
+
+    bond::maybe<bool> sink(std::move(m));
```
ara-ayvazyan: Test with something non-copyable?

cpp/inc/bond/core/maybe.h
chwarr: I mistakenly thought `move` wasn't needed because `that` is `maybe_common&&`. However, `that` is  an lvalue (of an rvalue ref). Will switch.

cpp/inc/bond/core/maybe.h
chwarr: Line got too long.

cpp/inc/bond/core/maybe.h
chwarr: boost::optional does not return `T&`, so `return _value.value()` it is.

ara-ayvazyan: Prefer `operator*` or `value()` instead of `get`. The `std::optional` does not have `get`.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
 
-    template <typename Allocator>
-    maybe(maybe&& that, const Allocator& alloc)
-        : _value(std::move(that._value), alloc),
-          _nothing(that._nothing)
+    /// @brief Set to nothing
+    void set_nothing()
     {
-        that._nothing = true;
+        _value = boost::none;
```
ara-ayvazyan: `_value = {};` or `_value.reset();`

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
+    : public detail::maybe_common<T>,
+      private detail::allocator_type<T>::type
+{
+public:
+    using allocator_type = typename detail::allocator_type<T>::type;
+
+    maybe() = default;
+    maybe(const maybe&) = default;
+
+// TODO: specialize std::uses_allocator with conditional on has_allocator && then is_convertable
```
ara-ayvazyan: Looks like you do need to specialize [`std::uses_allocator`](http://en.cppreference.com/w/cpp/utility/tuple/uses_allocator) if you define `allocator_type`.

chwarr: Excellent!

chwarr:
The `boost::in_place_init` I've been using was added in Boost 1.63. Need to find something that works back to 1.61 as well. https://www.boost.org/doc/libs/1_61_0/libs/optional/doc/html/boost_optional/tutorial/in_place_factories.html is promising.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
     }
 
     bool operator!=(const maybe& that) const
     {
         return !(*this == that);
     }
+};
+
+template <typename T>
+class maybe<T, typename boost::enable_if<detail::has_allocator<T>>::type>
```
chwarr: `has_allocator` or `std::uses_allocator`?

ara-ayvazyan: For `std::uses_allocator` you need to know the allocator type.

ara-ayvazyan:
Yeah, either `boost::in_place`, or default-construct and `emplace(...)`.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
-        : _nothing(true)
-    {}
+    using value_type = T;
 
-    /// @brief Move constructor
-    maybe(maybe&& that) BOND_NOEXCEPT_IF(std::is_nothrow_move_constructible<T>::value)
-        : _value(std::move(that._value)),
-          _nothing(std::move(that._nothing))
+    /// @brief Check if this object contains nothing
+    bool is_nothing() const
```
ara-ayvazyan: Add
```cpp
explicit operator bool() const BOND_NOEXCEPT { return !is_nothing(); }
```

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
+                                         && std::is_nothrow_move_constructible<allocator_type>::value)
+        : detail::maybe_common<T>(static_cast<detail::maybe_common<T>&&>(that)),
+          allocator_type(static_cast<allocator_type&&>(that))
+    { }
+
+    /// @brief Move constructor
+    maybe(
+        maybe&& that,
+        const allocator_type& alloc)
+        BOND_NOEXCEPT_IF(std::is_nothrow_move_constructible<typename detail::maybe_common<T>>::value
```
ara-ayvazyan: Consider the `BOND_NOEXCEPT_EXPR`, might be useful in some cases.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
-        _value = value;
-        _nothing = false;
+        // TODO: yes/no on passing allocator?
+        this->_value.emplace(value, static_cast<const allocator_type&>(*this));
         return *this;
     }
 
+    /// @brief If this object contains nothing, set it to a default instance
+    /// of T; otherwise, preserve the existing value. Then, return a
+    /// reference to the value.
```
ara-ayvazyan: Put the comment about return under `@return`?

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +18,249 @@ namespace bond
...
-private:
-    T       _value;
-    bool    _nothing;
-};
+    /// @brief Compare for equality
+    bool operator==(const maybe& that) const
+    {
+        return this->_value == that._value
+            &&
+        static_cast<const allocator_type&>(*this)
```
ara-ayvazyan: Do we really care about allocator equality?

chwarr: I'm not sure. Looking at MSVC and libstdc++'s vector and map, they don't care, so I think we can drop the allocator comparison too.

ara-ayvazyan: Yeah, the `operator==` in this case is about value equality.

chwarr:
@ara-ayvazyan: I've addressed a bunch of your comments.
I'm still working on adding more Doxygen comments, but the code is ready for reviewing.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +20,395 @@ namespace bond
...
+protected:
+    maybe_common() = default;
+    maybe_common(const maybe_common&) = default;
+
+    explicit maybe_common(const T& value)
+        : _value(value)
+    { }
+
+    template <typename Alloc>
+    maybe_common(const T& value, const Alloc& alloc)
```
ara-ayvazyan: Maybe generalize to `maybe_common(U&& value, Args&&.. args)`?

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +20,395 @@ namespace bond
...
+        : _value(boost::in_place<T>(value, alloc))
+    { }
+
+    maybe_common(maybe_common&& that) BOND_NOEXCEPT_IF(std::is_nothrow_move_constructible<boost::optional<T>>::value)
```
ara-ayvazyan: The ctor's body should be `noexcept` as well.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +20,395 @@ namespace bond
...
 
-    /// @brief Compare for equality
-    bool operator==(const maybe& that) const
+    /// @brief Compares a maybe and a value for equality.
+    ///
+    /// @return false if the maybe holds nothing; otherwise, calls
+    /// operator== with maybe's value and the provided value.
+    ///
+    /// @since 8.0.0
+    bool operator==(const T& that) const
```
ara-ayvazyan: Call it `value` when it is `T` and not `maybe`?

ara-ayvazyan: Consider making a free function rather than a member.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,143 +20,395 @@ namespace bond
...
 
+    const allocator_type& allocator() const BOND_NOEXCEPT
+    {
+        return static_cast<const allocator_type&>(*this);
+    }
+};
 
 template<typename T>
-inline void swap(maybe<T>& x, maybe<T>& y)
+inline void swap(maybe<T, void>& x, maybe<T, void>& y)
```
ara-ayvazyan: Skip the `, void` :)

cpp/inc/bond/core/maybe.h
```diff
@@ -14,140 +20,418 @@ namespace bond
...
+    maybe_common(const maybe_common&) = default;
+
+    template <typename... Args>
+    explicit maybe_common(const T& value, Args&&... args)
+        : _value(boost::in_place<T>(value, std::forward<Args>(args)...))
+    { }
+
+    template <typename... Args>
+    explicit maybe_common(T&& value, Args&&... args)
+        : _value(boost::in_place<T>(value, std::forward<Args>(args)...))
```
ara-ayvazyan: Missing `std::move`.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,140 +20,418 @@ namespace bond
...
+    /// @since 8.0.0
+    allocator_type get_allocator() const BOND_NOEXCEPT
+    {
+        return allocator();
+    }
+
+private:
+    detail::maybe_common<T>& base() BOND_NOEXCEPT
+    {
+        return static_cast<detail::maybe_common<T>&>(*this);
```
ara-ayvazyan: You can skip explicit cast I think.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,140 +20,418 @@ namespace bond
...
+    /// @brief Create a maybe that holds a copy of the value.
+    explicit maybe(const T& value)
+        : detail::maybe_common<T>(value),
+          allocator_type()
+    { }
+
+    /// @since 8.0.0
+    maybe(const T& value, const allocator_type& alloc)
+        : detail::maybe_common<T>(value, alloc),
+        allocator_type(alloc)
```
ara-ayvazyan: Indentation (few more paces).

cpp/inc/bond/core/maybe.h
```diff
@@ -14,140 +20,418 @@ namespace bond
...
+
+    /// @brief Create a  maybe that holds nothing.
+    maybe() = default;
+    maybe(const maybe&) = default;
+
+    maybe(const maybe& that, const allocator_type& alloc)
+        : detail::maybe_common<T>(),
+          allocator_type(alloc)
+    {
+        if (that._value)
```
ara-ayvazyan: Use `is_nothing` for consistency?

cpp/test/core/maybe_tests.cpp
```diff
@@ -1,20 +1,27 @@
 #include "precompiled.h"
 #include "maybe_tests.h"
 
+#include "allocators.h"
+
+#include <boost/none.hpp>
```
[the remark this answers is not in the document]
chwarr: `boost::none` is back, as it gets us `noexcept` in a few places.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,140 +20,418 @@ namespace bond
...
+          allocator_type(alloc)
+    {
+        this->emplace(std::move(value), alloc);
+    }
+
+    maybe& operator=(const maybe&) = default;
+    maybe& operator=(maybe&&) = default;
+    using detail::maybe_common<T>::operator=;
+
+    bool operator==(const allocator_type&) = delete;
```
ara-ayvazyan: Explanatory comment?

cpp/inc/bond/core/maybe.h
```diff
@@ -14,140 +20,418 @@ namespace bond
...
+    /// @return A reference to the value.
+    T& set_value()
     {
-        return value();
+        if (this->is_nothing())
+        {
+            this->emplace();
+        }
+
+        return *this->_value;
```
ara-ayvazyan:
```cpp
return this->is_nothing() ? this->emplace() : this->value(std::nothrow);
```

cpp/inc/bond/core/maybe.h
```diff
@@ -14,140 +20,418 @@ namespace bond
...
+    ///
+    /// @since 8.0.0
+    explicit
+    maybe(T&& value)
+        : detail::maybe_common<T>(std::move(value))
+    { }
+
+    maybe& operator=(const maybe&) = default;
+    maybe& operator=(maybe&&) = default;
+    using detail::maybe_common<T>::operator=;
```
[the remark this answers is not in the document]
chwarr: Good catch!

cpp/inc/bond/core/maybe.h
```diff
@@ -14,141 +19,556 @@ namespace bond
...
+    maybe_common(const maybe_common& that)
+        : _value(that._value)
+    { }
+    #else
+    maybe_common(const maybe_common&) = default;
+    #endif
+
+    template <typename... Args>
+    explicit maybe_common(const T& value, Args&&... args)
+        : _value()
```
ara-ayvazyan: The `: _value()` is redundant (will not be a POD for any `T`).

cpp/inc/bond/core/maybe.h
```diff
@@ -14,141 +19,556 @@ namespace bond
...
+    //
+    // Manually implementing this function works just fine, however.
+    maybe_common& operator=(const maybe_common& that)
+    {
+        _value = that._value;
+        return *this;
+    }
+
+    // MSVC 2013 cannot = default rvalue asignment operators
+    maybe_common& operator=(maybe_common&& that) BOND_NOEXCEPT_IF(BOND_NOEXCEPT_EXPR(_value = std::move(that._value)))
```
ara-ayvazyan: The `BOND_NOEXCEPT_EXPR` doesn't help much here, just use `std::is_nothrow_move_assignable`?

cpp/inc/bond/core/maybe.h
```diff
@@ -14,141 +19,556 @@ namespace bond
...
+    maybe& operator=(const maybe& that)
+    {
+        detail::maybe_common<T>::operator=(that);
+        return *this;
     }
 
-    /// @brief Check if this object contains nothing
-    bool is_nothing() const
+    // MSVC 2013 cannot = default rvalue asignment operators
+    maybe& operator=(maybe&& that) BOND_NOEXCEPT_IF(BOND_NOEXCEPT_EXPR(detail::maybe_common<T>::operator=(std::move(that))))
```
ara-ayvazyan: Same here, `std::is_nothrow_move_assignable<detail::maybe_common<T>>`?

cpp/inc/bond/core/maybe.h
```diff
@@ -14,141 +19,556 @@ namespace bond
...
+    { }
+
+    /// @brief Create a maybe that holds a copy of the value.
+    explicit maybe(const T& value)
+        : detail::maybe_common<T>(value),
+          allocator_type()
+    { }
+
+    /// @since 8.0.0
+    maybe(const T& value, const allocator_type& alloc)
```
ara-ayvazyan: Why is this constructor needed?

cpp/inc/bond/core/maybe.h
```diff
@@ -14,141 +19,556 @@ namespace bond
...
+    // to make extra copies, which makes this type fail when it's holding a
+    // move-only type.
+    //
+    // Manually implementing this function works just fine, however.
+    maybe& operator=(const maybe& that)
     {
-        return _nothing == that._nothing
-            && (_nothing || _value == that._value);
+        detail::maybe_common<T>::operator=(that);
+        allocator_type::operator=(that);
```
ara-ayvazyan: Use `base()` and `allocator()` helpers?

CHANGELOG.md
```diff
@@ -58,6 +58,9 @@ different versioning scheme, following the Haskell community's
   have been removed. The `blob` and `nullable` types are not customizable,
   so these where never needed or used. The related functionality provided by
   `bond::get_list_sub_type_id` remains.
+* **Breaking change*** Removed a dangerous implicit conversion operator from
```
ara-ayvazyan: There is an extra `*`.

CHANGELOG.md
```diff
@@ -74,6 +77,14 @@ different versioning scheme, following the Haskell community's
 * Added `bond::blob_prolong` helper function that will return a `bond::blob` with
   a copied data if the original one does not own the memory.
 * The `bond::OutputBuffer::GetBuffers` now can accept arbitrary STL-like containers.
+* `bond::maybe<T>` has been overhauled.
+    * Fixed a bug that default initialized an instance of `T` even when a maybe held nothing.
+    * Added `noexcept` variants of `bond::maybe<T>::value`.
+    * Added `bond::maybe<T>::emplace` to construct a maybe's value in place.
+    * Added various rvalue-reference and allocator-aware constructors and
+      assignment operators.
+    * Added `bond::maybe<T>::operator==(const T&)` to comapre directly to
```
ara-ayvazyan: The `operator==` is not a member anymore.

cpp/inc/bond/core/maybe.h
chwarr: ✔️

CHANGELOG.md
chwarr: ✔️

CHANGELOG.md
chwarr: ✔️

cpp/inc/bond/core/maybe.h
chwarr: ✔️

cpp/inc/bond/core/maybe.h
chwarr: ✔️

cpp/inc/bond/core/maybe.h
chwarr: ✔️ 
I tried to add tests for these, but the MSVC 2013 implementation is missing a bunch of the type traits. I could have used to Boost wrappers, but then the 2013 STL is missing `noexcept` specifiers on a lot of its types. So, for 2013, full `noexcept` support doesn't seem worth it...

cpp/inc/bond/core/maybe.h
chwarr: ✔️

cpp/inc/bond/core/maybe.h
```diff
@@ -14,140 +22,541 @@ namespace bond
...
+protected:
+    maybe_common() = default;
+
+    #if defined(_MSC_VER) && _MSC_VER < 1900
+    // Using = default with MSVC 2013 on this function causes the compiler
+    // to make extra copies, which makes this type fail when it's holding a
+    // move-only type.
+    //
+    // Manually implementing this function works just fine, however.
+    maybe_common(const maybe_common& that)
```
ara-ayvazyan: Make standard copy/move constructors public?

cpp/inc/bond/core/maybe.h
```diff
@@ -14,140 +22,541 @@ namespace bond
...
+
+    // Using = default with MSVC 2013 on this function causes the compiler
+    // to make extra copies, which makes this type fail when it's holding a
+    // move-only type.
+    //
+    // Manually implementing this function works just fine, however.
+    maybe(const maybe& that)
+        : detail::maybe_common<T>(that),
+          holder(that)
+        { }
```
ara-ayvazyan: Indentation

cpp/inc/bond/core/maybe.h
```diff
@@ -14,140 +22,541 @@ namespace bond
...
     {
-        return _nothing == that._nothing
-            && (_nothing || _value == that._value);
+        _value.emplace(value, std::forward<Args>(args)...);
     }
 
-    bool operator!=(const maybe& that) const
+    template <typename... Args>
+    explicit maybe_common(T&& value, Args&&... args)
+        : _value()
```
ara-ayvazyan: Can you check all the redundant constructor calls? :)

cpp/test/core/maybe_tests.cpp
```diff
@@ -64,25 +71,36 @@ TEST_CASE_BEGIN(MaybeInterface)
     UT_AssertIsFalse(z == x);
     UT_AssertIsTrue(z == y);
     UT_AssertIsTrue(x.is_nothing());
+    UT_AssertIsFalse(x);
```
ara-ayvazyan: Also check for `!x` to make sure that `!` works as expected.

cpp/test/core/maybe_tests.cpp
```diff
@@ -107,6 +126,259 @@ TEST_CASE_BEGIN(MaybeInterface)
 }
 TEST_CASE_END
 
+TEST_CASE_BEGIN(CopyAndMoveFromValues)
+{
+    { // simple type
```
ara-ayvazyan: Use `BOOST_TEST_CONTEXT`?

cpp/test/core/maybe_tests.cpp
```diff
@@ -107,6 +126,259 @@ TEST_CASE_BEGIN(MaybeInterface)
...
+        source2.emplace_back(new int{ 2 });
+
+        movedValue = std::move(source2);
+        BOOST_CHECK(source2.empty());
+        BOOST_REQUIRE(movedValue);
+        BOOST_CHECK(movedValue.value().size() == 2);
+    }
+}
+TEST_CASE_END
+
```
ara-ayvazyan: Test for non-default-constructible types?

cpp/test/core/maybe_tests.cpp
```diff
@@ -107,6 +126,259 @@ TEST_CASE_BEGIN(MaybeInterface)
...
+        BOOST_CHECK(sink.value().size() == 2);
+    }
+}
+TEST_CASE_END
+
+struct UsesAllocator
+{
+    using allocator_type = allocator_with_state<>;
+
+    UsesAllocator() = default;
```
ara-ayvazyan: Also a test with non-default-constructible allocator (perhaps skipping VC12).

chwarr: `allocator_with_state` is not default constructible, so I'll just remove this.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,140 +22,541 @@ namespace bond
...
+    #if defined(_MSC_VER) && _MSC_VER < 1900
+    // Using = default with MSVC 2013 on this function causes the compiler
+    // to make extra copies, which makes this type fail when it's holding a
+    // move-only type.
+    //
+    // Manually implementing this function works just fine, however.
+    maybe& operator=(const maybe& that)
+    {
+        base_common() = that.base_common();
+        holder::operator=(that);
```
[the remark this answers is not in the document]
chwarr: Whoops.

cpp/test/core/maybe_tests.cpp
chwarr: The `AllocatorPropagated` test below uses a non-default-constructible type (with an allocator). Are you saying we need a test for a non-default-constructible type without an allocator?

cpp/test/core/maybe_tests.cpp
chwarr: Added in the first two places, but not everywhere.

cpp/test/core/maybe_tests.cpp
chwarr: Oh. Neat!

cpp/inc/bond/core/maybe.h
chwarr: Removed the handful that were left.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,140 +22,541 @@ namespace bond
...
+    { }
+    #else
+    /// @brief Create a  maybe that holds nothing.
+    maybe() = default;
+    maybe(const maybe&) = default;
+    maybe(maybe&&) = default;
+    #endif
+
+    maybe(const maybe& that, const allocator_type& alloc)
+        : detail::maybe_common<T>(),
```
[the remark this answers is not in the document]
chwarr: In cases when we're initializing one of the bases, I'd like to mention all of them. ("All or none" is the guideline I'm following.)

cpp/test/core/maybe_tests.cpp
ara-ayvazyan: No, that should be enough.

cpp/test/core/maybe_tests.cpp
```diff
@@ -1,28 +1,41 @@
...
+#include <initializer_list>
+#include <iterator>
+#include <map>
+#include <vector>
+
 void ConstInterface(const bond::maybe<int8_t>& x)
 {
-    bond::maybe<int8_t> z;
-    int8_t y;
+    BOOST_TEST_CONTEXT("Const interface")
```
ara-ayvazyan: Looks like VC12 is still using boost 1.58 which does not have `BOOST_TEST_CONTEXT`.

cpp/inc/bond/core/detail/alloc.h
```diff
@@ -0,0 +1,86 @@
...
+    >;
+
+template <typename Alloc>
+class allocator_holder<Alloc, typename boost::enable_if<empty_base_eligible<Alloc>>::type>
+    : private Alloc
+{
+public:
+    allocator_holder() = default;
+
+    allocator_holder(const Alloc& alloc)
```
ara-ayvazyan: Missing `explicit`? Same below.

cpp/inc/bond/core/maybe.h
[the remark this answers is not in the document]
ara-ayvazyan: Ok.

cpp/inc/bond/core/maybe.h
```diff
@@ -14,141 +22,542 @@ namespace bond
...
+        if (!that.is_nothing())
+        {
+            this->emplace(std::move(*that._value), alloc);
+        }
+
+        // unlike std::optional/boost::optional, moved-from bond::maybe
+        // instances are guaranteed to be nothing.
+        //
+        // asigning boost::none is noexcept, but assigning { } is not
+        that._value = boost::none;
```
ara-ayvazyan: Move this inside the above `if`?

cpp/inc/bond/core/detail/alloc.h
```diff
@@ -0,0 +1,86 @@
...
+///
+/// @remarks Applies empty-base-optimization when possible.
+template <typename Alloc, typename Enable = void>
+class allocator_holder;
+
+template <typename Alloc>
+using empty_base_eligible = std::integral_constant<bool,
+       std::is_empty<Alloc>::value
+    && std::is_copy_constructible<Alloc>::value
+#if __cplusplus >= 201402L
```
ara-ayvazyan: Make this the first condition to avoid having `>` on a single line (31)? :)

cpp/test/core/maybe_tests.cpp
chwarr: Sigh. I have a fix for that to productionalize...

cpp/inc/bond/core/detail/alloc.h
chwarr: I tried a few different ways to lay this out. The cleanest I found is adding a vacuous `true`.

cpp/inc/bond/core/maybe.h
chwarr: ✔️

cpp/inc/bond/core/detail/alloc.h
chwarr: ✔️

cpp/inc/bond/core/maybe.h
```diff
@@ -14,141 +22,584 @@ namespace bond
...
+    /// maybes holding nothing are handled.
+    ///
+    /// @since 8.0.0
+    friend bool operator!=(const T& lhs, const maybe_common& rhs)
+    {
+        return !(lhs == rhs);
+    }
+
+protected:
+    maybe_common& assign(const T& value)
```
ara-ayvazyan: Do we really need `assign` functions?

chwarr: Nope. Removed.

PR #854, opened 2018-04-18
[appveyor] Use Boost from NuGet packages if needed

chwarr:
The AppVeyor builder images contain some versions of Boost, but not all
the versions that Bond supports for all the compilers that it supports.

This commit:

* Switches to a minimum Boost of 1.61, the minimum supported version as
  of this commit.
* Adds a script to easily find the pre-installed Boost library.
* Adds a script to install a given version of Boost from NuGet.org
  packages.
* Updates the AppVeyor build to install missing versions of Boost from
  NuGet.org.
* Refactors the AppVeyor build to base the MSVC compiler off of the
  BOND_VS_VERSION or the APPVEYOR_BUILD_WORKER_IMAGE variable, which
  should set us up to add MSVC 2017 to the build later.

This is the AppVeyor side of https://github.com/Microsoft/bond/issues/771

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -66,7 +66,7 @@
             - BOND_BUILD: Python
               APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
               BOND_ARCH: 32
-              BOND_BOOST: "1.61.0"
+              BOND_BOOST: "1.62.0"
               BOND_CMAKE_FLAGS: "-DBOND_ENABLE_GRPC=FALSE"
             - BOND_BUILD: Python
               APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -40,17 +40,19 @@
         matrix:
               # C++ Core build and tests (Visual C++ 2013)
             - BOND_BUILD: C++
-              BOND_VS: "Visual Studio 12 2013"
-              BOND_VS_NUM: 12
+              BOND_VS_VERSION: Visual Studio 2013
+              # We need to use the Visual Studio 2015 image for the 2013 build
+              # as the C# compatibility tests need a C# 6 compiler, which is
+              # available starting with the 2015 image.
+              APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
               BOND_ARCH: 32
-              BOND_BOOST: 58
+              BOND_BOOST: "1.61.0"
               BOND_CMAKE_FLAGS: "-DBOND_ENABLE_GRPC=FALSE;-DBOND_SKIP_GBC_TESTS=TRUE"
               # C++ Core build and tests
             - BOND_BUILD: C++
-              BOND_VS: "Visual Studio 14 2015"
-              BOND_VS_NUM: 14
+              APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
               BOND_ARCH: 64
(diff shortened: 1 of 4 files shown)


tools/ci-scripts/windows/Install-Boost.ps1
```diff
@@ -0,0 +1,172 @@
...
+        'boost_unit_test_framework'),
+
+    [string]
+    $OutputDirectory = $null
+)
+
+Set-StrictMode -Version Latest
+$ErrorActionPreference = 'Stop'
+
+function Get-TempFolder
```
ara-ayvazyan: Why not `System.IO.Path.GetTempFileName`, or simply append a GUID?

tools/ci-scripts/windows/Install-Boost.ps1
```diff
@@ -0,0 +1,172 @@
...
+
+    if ($PSCmdlet.ShouldProcess('boost', 'Install NuGet package'))
+    {
+        Install-NuGetPackage `
+            -PackageId 'boost' `
+            -InstallDir $workDir `
+            -PackageVersion $Version
+
+        Move-Item `
+            -Path ([System.IO.Path]::Combine($workDir, 'boost', 'lib', 'native', 'include', 'boost')) `
```
ara-ayvazyan: Is there any benefit in using the sub-folder names separately instead of a simple `boost\lib\native\include\boost`?

appveyor.yml
```diff
@@ -62,26 +64,33 @@
...
+              BOND_BOOST: "1.63.0"
               BOND_CMAKE_FLAGS: "-DBOND_SKIP_CORE_TESTS=TRUE;-DBOND_SKIP_GBC_TESTS=TRUE"
     install:
         - ps: >-
+            if (-not $env:BOND_VS_VERSION)
+            {
+                $env:BOND_VS_VERSION = $env:APPVEYOR_BUILD_WORKER_IMAGE
+            }
+
+            if (-not $env:BOND_VS_VERSION)
```
ara-ayvazyan: Should this be moved inside the above `if`?

appveyor.yml
```diff
@@ -92,17 +101,38 @@
...
+                    'boost_system',
+                    'boost_unit_test_framework')
+
+                if ($env:BOND_BUILD -eq 'Python') {
+                    $boostComponents += 'boost_python'
+                }
 
-            if ($env:BOND_BOOST -eq 56) {
-                # Hard-coded Boost path per https://www.appveyor.com/docs/installed-software#languages-libraries-frameworks
+                $vsNum = tools\ci-scripts\windows\ConvertTo-MsvcVersion.ps1 -ImageName $env:BOND_VS_VERSION -Format VsNum
```
ara-ayvazyan: A minor thing, but I think this is `VcNum` not a `VsNum`.

chwarr: Hmmm... Let me check. I've pushed a fixup for some of your other comments, and will come back to this one when I have an answer.

tools/ci-scripts/windows/Install-Boost.ps1
chwarr: ✔️

chwarr:
I renamed it from "VsNum" to "VcToolSetVer".

chwarr:
I just found out that this won't work for Boost 1.66. The directory layout has changed a bit. I'd like to get this in and make adjustments for Boost 1.66 in a different PR if that's OK by you.

PR #857, opened 2018-04-18
[c# gbc] Add support to generate constructors with parameters

BertusGreeff:
This provides a codegen supported way to construct and initialize C#
classes generated with the –readonly-properties option.


compiler/tests/generated/constructor_parameters_types.cs
```diff
@@ -0,0 +1,158 @@
...
+        public Foo()
+        {
+            someText = "BaseText1";
+        }
+    }
+
+    [global::Bond.Schema]
+    [System.CodeDom.Compiler.GeneratedCode("gbc", "0.11.0.0")]
+    public partial class Bar
+        : Foo
```
sapek: Could you add a test with a class that derived from `Bar`?

compiler/src/Language/Bond/Syntax/Internal.hs
```diff
@@ -51,3 +52,13 @@ metaField Field {..} = Any $ isMetaName fieldType
 isBaseField :: String -> Maybe Type -> Bool
 isBaseField name = getAny . optional (foldMapFields (Any.(name==).fieldName))
 
+-- If a Declaration is a Struct then return the Fields of its Parent
```
sapek: There is a helper function `foldMapFields` you can use here:

    baseFields Struct{..} = foldMapFields return <$> structBase

The type of above is  `Maybe [Field]`. It will return `Nothing` if there is no base and list of fields, possibly empty, the whole base *hierarchy* (which is what I think you need). I think that `Maybe [Field]` is better than `[Field]` for the codegen (code for class with base that has no fields may look different than code for class with no base). But if you wanted to return `[]` in all cases then you could do:

    baseFields Struct{..} = concat $ foldMapFields return <$> structBase

compiler/src/Language/Bond/Codegen/Cs/Types_cs.hs
```diff
@@ -111,7 +115,61 @@ namespace #{csNamespace}
...
+                else [lt|
+            : base(fullName, name)|]
+
+        -- constructor: ConstructorParameters option
+        constructorWithParameters = if not noMetaFields
+            then error "bond_meta usage is incompatible with constructor_parameters"
+            else [lt|
+
+        public #{declName}(
+#{baseParameterBlock}            #{commaLineSep 3 thisParam structFields}
```
sapek: It would more readable if you aligned the `#(baseParameterBlock}` here and not have the space padding it its definition.

compiler/Main.hs
```diff
@@ -119,8 +119,11 @@ csCodegen options@Cs {..} = do
             else if fields
                  then PublicFields
                  else Properties
+    constructorMapping = if constructor_parameters
```
sapek: Nit: `mapping` is not a good term here. In case of fields and types we have them in the schema and we map them in some way to the target language. There are no constructors in the schema so there is nothing to map.

compiler/tests/generated/constructor_parameters_types.cs
BertusGreeff: Added.  Thanks for pointing out that my code was not working for more than one level of derivation.

compiler/src/Language/Bond/Syntax/Internal.hs
BertusGreeff: I found foldMapFields but I could not figure out the syntax to use it.  My Haskell is not good.  I've changed to using it.

compiler/tests/schema/constructor_parameters_fields.bond
```diff
@@ -0,0 +1,44 @@
+namespace Test
```
chwarr: Since the two .bond files are identical, take a look at what was done for collection-interfaces to put the expected files in a different directory.

compiler/Options.hs
```diff
@@ -110,6 +111,7 @@ cs = Cs
     , fields = def &= name "f" &= help "Generate public fields rather than properties"
     , structs_enabled = True &= explicit &= name "structs" &= help "Generate C# types for Bond structs and enums (true by default, use \"--structs=false\" to disable)"
     , grpc_enabled = False &= explicit &= name "grpc" &= help "Generate C# services and proxies for gRPC"
+    , constructor_parameters = def &= help "Generate a constructor that takes all the fields as parameters.  Typically used with readonly-properties."
```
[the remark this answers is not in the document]
BertusGreeff: I can but how do we want this to ultimately work?  Right now --constructor-parameters can be used by itself even if the user does not specify --readonly-properties.

I was thinking to introduce --readonly-collections afther this PR and then all three options together generates a true immutable type.  Maybe we can then add a --immutable option that implies all three options as a convenience.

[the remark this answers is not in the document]
chwarr: I'd like to retain the ability to rework the flags after adding something like `--readonly-collections`, so I'd like to communicate to consumers of gbc that this particular flag may change its name/behavior in the future.

compiler/tests/Tests/Codegen.hs
```diff
@@ -141,25 +147,25 @@ verifyFiles options baseName =
...
         ] ++
         [ testGroup "constructors with allocator argument" $
-            map (verify (cppExpandAliasesTypeMapping $ cppCustomAllocTypeMapping False "arena") "alloc_ctors")
+            map (verify (cppExpandAliasesTypeMapping $ cppCustomAllocTypeMapping False "arena") (variation </> "alloc_ctors"))
                 (templates $ options { allocator = Just "arena", alloc_ctors_enabled = True })
             | isNothing allocator
         ] ++
         [ testGroup "type aliases" $
-            map (verify (cppCustomAllocTypeMapping False "arena") "type_aliases")
+            map (verify (cppCustomAllocTypeMapping False "arena") ("type_aliases" </> variation))
```
chwarr: Was this supposed to be `variation </> "type_aliases"`?

compiler/src/Language/Bond/Codegen/Cs/Types_cs.hs
```diff
@@ -158,16 +160,18 @@ namespace #{csNamespace}
...
+        uniqueBaseFieldNames = uniqueNames (map fieldName baseFieldList) []
+        uniqueThisFieldNames = uniqueNames (map fieldName structFields) uniqueBaseFieldNames
 
-        baseParam (f, n) = [lt|#{csType $ fieldType f} #{n}|]
-        thisParam f = [lt|#{csType $ fieldType f} #{uniqueName (fieldName f) uniqueBaseFieldNames}|]
+        paramDecl f = [lt|#{csType $ fieldType f} #{fieldName f}|]
+        paramDeclWithUniqueName (f, n) = [lt|#{csType $ fieldType f} #{n}|]
 
-        paramBasedInitializer f = [lt|this.#{fieldName f} = #{uniqueName (fieldName f) uniqueBaseFieldNames};|]
+        paramBasedInitializer f = [lt|this.#{fieldName f} = #{fieldName f};|]
```
chwarr: Consider merging these two so we always pass a tuple: sometimes it's the normal name and sometimes its the unique name.

compiler/tests/generated/constructor-parameters/collection-interfaces/complex_inheritance_types.cs
```diff
@@ -0,0 +1,210 @@
...
+            int someInt,
+            string moreText,
+            IList<Simple> someList,
+            IDictionary<string, double> someMap,
+            ISet<string> someSet,
+
+            // This class parameters
+            string someText1,
+            string evenMoreText,
+            string someText10
```
chwarr: This is a good test that got added. Thanks.

compiler/tests/Tests/Codegen.hs
BertusGreeff: Yes, thanks for pointing out.

compiler/src/Language/Bond/Codegen/Cs/Types_cs.hs
BertusGreeff: Addressed in next iteration.

PR #860, opened 2018-04-20
[c++] Rename scoped_allocator example

chwarr:
Rename the scoped allocator example from "scoped_allocator" to
"scoped_alloc" so that it doesn't conflict with the "scoped_allocator"
header from the C++ standard library.

When --scoped-alloc is passed to gbc, the generated _types.h header
has a `#include <scoped_allocator>`.

When using CMake to generate Makefiles, the directory where the _types.h
generated file is placed is the same directory where the final
executable is placed.

When rebuilding the scoped_allocator example, the preprocessor tries to
#include the previous generated executable.

Implementations of the preprocessor differ in how they handle "file" vs.
\<file\>. Some will search in the current directory for \<file\>, while some
will not.

Renaming avoids this problem instead of relying on the behavior of
\<file\>.


examples/cpp/core/scoped_allocator/CMakeLists.txt
```diff
@@ -1,10 +1,10 @@
...
         --header=\\\"allocator.h\\\"
         --alloc-ctors
         --scoped-alloc)
 
-add_bond_test (scoped_allocator
-	${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/scoped_allocator_types.cpp
-	scoped_allocator.cpp)
+add_bond_test (scoped_alloc
+        ${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}/scoped_alloc_types.cpp
+        scoped_alloc.cpp)
```
chwarr: There's a tabs to spaced changes as well, hence some of the diff noise.

PR #866, opened 2018-04-25
[c#] Generate files under gbc-specific directory

chwarr:
In order to avoid conflcits with other tools that write files to
$(IntermediateOutputPath) with the same names as gbc, the codegen
targets now generate the files under $(IntermediateOutputPath)\gbc
unless overridden.


chwarr:
This will likely make all our ReSharper workarounds stop working.
Let me see whether I can code something up that obviates the need for a ReSharper workaround...

chwarr:
I was not able to figure out an elegant way to get rid of the need for the ReSharper workaround.
I'm switching teams within Microsoft and will not be able to finish this PR before my move.

Issue #869, opened 2018-05-04
gbc cannot handle import with mixed/other slashes


chwarr:
On Linux, gbc cannot handle import statements with mixed or backslashes.

Given ./inc/one/two/a.bond:

```
namespace a;
struct foo { }
```

And ./b.bond:

```
import "one/two\a.bond"
namespace b;
struct bar {
    0: a.foo myFoo;
}
```

Run codegen as:

```
gbc c# -i ./inc/ b.bond
```

Expected results:

b_types.cs is created with references to `a.foo`

Actual results:

> gbc: user error (Can't find import file one\two/a.bond)

Import statements with mixed or backslashes are going to be common if the .bond files were originally authored on Windows. (gbc on Windows can handle mixed slashes just fine.)

amrmahdi:
@chwarr when will the change be merged ? Also, are you going to create a release for this change ?

chwarr:
Should be merged by end-of-day, Pacific time, on Wednesday, May 9, 2018.
We're planning the Bond 8.0 release some time around May 15, 2018.

amrmahdi:
@chwarr what is the current target date for 8.0 release ?

chwarr:
Tuesday, May 29, 2018

PR #870, opened 2018-05-05
[gbc] Handle forward- and backslashes in paths

chwarr:
Since .bond files can be authored on platforms with different directory
separator characters, the paths in build scripts and Bond `import`
statements may use a forwardslash when a backslash is expected or vice
versa.

In order to handle this, gbc now normalizes anything that is a
path (file paths, response file items, import directories, import
statement paths, output directories, &c.) to use the platform's
preferred directory separator character.

Tests for C++, C#, and Java have been updated to use some imports with
mixed slashes.

gbc doesn't use a dedicate type for file paths. `FilePath` is just a
type synonym for `String`, so some places may have been inadvertently
missed.

This does now mean that gbc cannot process files, on, say, Linux, with
backslashes in their names. This is expected to be rare, and is deemed
an acceptable compromise for a cross-platform tool.

Fixes https://github.com/Microsoft/bond/issues/869


chwarr:
The clang build is broken. For the `#include` generated for `import` statements, we should force the slashes to always be forward.

chwarr:
I think I've resolved the Clang build errors in commit a8f41fbc.

CHANGELOG.md
```diff
@@ -42,7 +42,10 @@ different versioning scheme, following the Haskell community's
 * C++ codegen now applies the `--export-attribute` to the `ToString`,
   `FromString`, `ToEnum` and `FromEnum` functions.
 * `import` statements can now end with an optional semicolon.
-
+* File and directory paths on the command line, in response files, or in
```
lalo: what are response files?

chwarr: gbc response files are inputs that start with "@". If the input file is prefixed with "@", then gbc will treat each line of that file as an input file.

chwarr:
@lalo: can you merge this?

chwarr:
I just rebased this atop the recent changes to codegen and the C# project changes.

lalo:
I'll merge this after CI.

ara-ayvazyan:
Generated test files need to be regenerated.

chwarr:
Thanks. I've fixed the generated test files.

PR #871, opened 2018-05-08
[c++] Refactor nullable<T>

ara-ayvazyan:
- Remove the `useValue` type argument in order to avoid unintentional use.
- Remove redundant `Allocator` type argument.
- Remove unnecessary typedefs (`reference`, `pointer`, etc.).
- Use `boost::optional<T>`.
- Use `detail::allocator_holder<A>`.


cpp/inc/bond/core/nullable.h
```diff
@@ -37,308 +37,303 @@ use_value
...
         reset();
     }
 
-    /// @brief Assignment operator
-    nullable& operator=(const nullable& src)
+    bool hasvalue() const BOND_NOEXCEPT
     {
-        nullable(src).swap(*this);
-        return *this;
+        return !!_value;
```
chwarr: Is the `!!` idiom to handle custom `pointer` types that don't have an `explicit operator bool`?

cpp/inc/bond/core/nullable.h
```diff
@@ -37,308 +37,303 @@ use_value
...
+
+    /// @brief Checks if the object is null
+    bool empty() const BOND_NOEXCEPT
+    {
+        return !hasvalue();
+    }
+
+    explicit operator bool() const BOND_NOEXCEPT
+    {
+        return !empty();
```
chwarr: Minor: `return hasvalue()` to remove one level of negation?

cpp/inc/bond/core/nullable.h
```diff
@@ -499,58 +448,49 @@ class nullable<T, Allocator, false>
...
-        return new_value(base());
+        return new_value();
     }
 
-    real_pointer set_value(std::false_type)
+    template <typename U>
+    void set_value(U&& value)
     {
-        return new_value();
+        if (empty())
```
chwarr: Minor: since this has an else, it should have curly braces now too.

cpp/inc/bond/core/nullable.h
ara-ayvazyan: Old style to enforce 0/1 :) Will make an explicit cast.

cpp/inc/bond/core/nullable.h
```diff
@@ -37,308 +37,303 @@ use_value
...
-          _hasvalue(true)
-    {}
-
-    nullable(nullable&& src) BOND_NOEXCEPT_IF(
-            std::is_nothrow_move_constructible<Allocator>::value
-            && std::is_nothrow_move_constructible<value_type>::value)
-        : Allocator(std::move(src.base())),
-          _value(std::move(src._value)),
-          _hasvalue(std::move(src._hasvalue))
+    void set(const T& value) BOND_NOEXCEPT_IF(std::is_nothrow_copy_constructible<T>::value)
```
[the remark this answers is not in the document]
ara-ayvazyan: Changed to use `std::is_nothrow_copy_assignable` and `std::is_nothrow_move_assignable`.

From the [doc](https://www.boost.org/doc/libs/1_67_0/libs/optional/doc/html/boost_optional/reference/header__boost_optional_optional_hpp_/header_optional_optional_values.html):
> Throws: Whatever T::operator=( T const& ) or T::T(T const&) throws.
> Throws: Whatever T::operator=( T&& ) or T::T(T &&) throws.

ara-ayvazyan:
@chwarr, should we update the versions in `bond_version.h`, since this change requires that generated code should match the library?

chwarr:
We should add a note to the changelog that the versions need to be bumped, but hold off on bumping the version for now. (E.g., a bullet under the C++ version note that we have now.) We bump all the ones we need to when we prepare a release in one commit, usually.

PR #873, opened 2018-05-08
[c++ grpc] Add helper io_manager_tag::tag

ara-ayvazyan:


cpp/inc/bond/ext/grpc/detail/io_manager_tag.h
```diff
@@ -15,13 +15,18 @@ namespace bond { namespace ext { namespace gRPC { namespace detail {
...
         /// @brief Called when this instance is dequeued from a completion
         /// queue.
         ///
         /// @param ok whether or not the initial operation succeeded
         virtual void invoke(bool ok) = 0;
+
+        io_manager_tag* tag()
+        {
+            return this;
+        }
```
chwarr: Consider adding the reverse function to move the tag/untag logic all into `io_manager_tag`:

```cpp
/// Invokes a previously tagged completion queue operation.
///
/// @param tag A %tag value, as produced by the tag function, to resume.
/// @param ok whether or not the initial operation succeeded
static void invoke_tag(void* tag, bool ok)
{
    static_cast<io_manager_tag*>(tag)->invoke(ok);
}
```

Doing this also lets us make `tag()` return `void*`, making the tag opaquer.

cpp/inc/bond/ext/grpc/detail/io_manager_tag.h
```diff
@@ -15,13 +15,18 @@ namespace bond { namespace ext { namespace gRPC { namespace detail {
...
-        virtual ~io_manager_tag() { }
+        virtual ~io_manager_tag() = default;
 
         /// @brief Called when this instance is dequeued from a completion
         /// queue.
         ///
         /// @param ok whether or not the initial operation succeeded
         virtual void invoke(bool ok) = 0;
+
+        io_manager_tag* tag()
```
chwarr: Minor: add a brief doc comment? Since this is an interface--albeit an internal one--it's been documented to outline the contract it expects.

```cpp
/// @return Returns a %tag value suitable for passing to completion queue routines.
```

cpp/inc/bond/ext/grpc/detail/io_manager_tag.h
chwarr: Good point. Carry on without.

ara-ayvazyan:
The build failure is during `stack` setup for C#.

ara-ayvazyan:
Oops, accidentally closed.

chwarr:
Oh well. CI build status will be delayed a bit. No worries.

PR #874, opened 2018-05-08
[c#] Migrate to VS 2017 csproj

lalo:
- This change will consolidate the .Net Core into the csproj
- The new csproj is only supported by VS2017 and the .Net Core SDK tooling
- Added new test project coreNS10 to test the Net Standard 1.0 binaries


lalo:
Had to trigger manually the docker image build, I'll update the script on a fixup commit

cs/test/core/Core.csproj
```diff
@@ -29,53 +23,23 @@
     <BondOptions>--fields --collection-interfaces</BondOptions>
   </PropertyGroup>
   <ItemGroup>
-    <BondImportDirectory Include="import dir with spaces\" />
+    <BondImportDirectory Include="import dir with spaces" />
```
chwarr: You had mentioned that when this had the trailing slash you were getting a warning. I cannot repro that warning with MSBuild 15.5.180.51428 and .NET Core SDK 2.1.4.

Perhaps it was showing up as an artifact of something else that hadn't been fixed during the migration.
If we can leave the trailing slash, let's do that.

chwarr: Gotcha. It's from the Visual Studio project system. Try this patch:

```patch
diff --git a/cs/build/nuget/Common.targets b/cs/build/nuget/Common.targets
index 70f6d268..8ae0ed45 100644
--- a/cs/build/nuget/Common.targets
+++ b/cs/build/nuget/Common.targets
@@ -52,7 +52,7 @@
       <Context>File;BrowseObject;</Context>
     </PropertyPageSchema>
     <PropertyPageSchema Include="$(MSBuildThisFileDirectory)cps\BondImportDirectory.xaml">
-      <Context>File;BrowseObject;</Context>
+      <Context>Invisible</Context>
     </PropertyPageSchema>
   </ItemGroup>
```

lalo: That worked

lalo:
@chwarr do not merge yet, I'm waiting on a vsts issue.

lalo:
@chwarr issue is fixed on that last commit, everything looks good depending on CI

PR #875, opened 2018-05-08
[c++ grpc] Simplify service_unary_call_data

ara-ayvazyan:
The change avoids the explicit call to `service_unary_call_data::queue_receive` in the generated code.

### DIFF
--- a/cpp/inc/bond/ext/grpc/client_callback.h
+++ b/cpp/inc/bond/ext/grpc/client_callback.h
@@ -55,11 +55,6 @@ namespace bond { namespace ext { namespace gRPC {
               status(status),
               context(std::move(context))
         { }
-
-        unary_call_result(const unary_call_result&) = default;
-        unary_call_result(unary_call_result&&) = default;
-        unary_call_result& operator=(const unary_call_result&) = default;
-        unary_call_result& operator=(unary_call_result&&) = default;
     };
 
 } } } // namespace bond::ext::gRPC
--- a/cpp/inc/bond/ext/grpc/detail/service.h
+++ b/cpp/inc/bond/ext/grpc/detail/service.h
@@ -35,7 +35,8 @@ template <typename TThreadPool>
 class service : private grpc::Service
 {
 public:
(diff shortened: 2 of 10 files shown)


cpp/inc/bond/ext/grpc/client_callback.h
```diff
@@ -55,11 +55,6 @@ namespace bond { namespace ext { namespace gRPC {
               status(status),
               context(std::move(context))
         { }
-
```
chwarr: The (incorrect) intention behind these `= default` was to say that `unary_call_result`  is copyable and moveable. Can we add `static_assert` tests for these properties?

cpp/inc/bond/ext/grpc/detail/service.h
```diff
@@ -35,7 +35,8 @@ template <typename TThreadPool>
 class service : private grpc::Service
 {
 public:
-    virtual ~service() { }
```
chwarr: Since this is a detail type, the client isn't supported to mention it, so there's no way to have something to polymorphically destroy. Good catch.

cpp/inc/bond/ext/grpc/client_callback.h
ara-ayvazyan: Will add a test check. The `=default` is not for documenting the code, in my opinion.

cpp/inc/bond/ext/grpc/detail/service.h
ara-ayvazyan: The `grpc::Service` already has a `virtual` destructor, so this is useless.

chwarr: Ah. In that case, can we also add a `has_virtual_destructor` static_assert test against one of the generated `ServiceCore` classes?

The API that has been shipped today explicitly says "You can polymorphically delete things derived from the generated `ServiceCore`.", and such a test will guard against the unlikely case that `grpc::Service` has it's virtual dtor removed.

PR #876, opened 2018-05-09
[c++] Fix allocator propagation in nullable<T>

ara-ayvazyan:


cpp/test/core/nullable.cpp
```diff
@@ -227,30 +227,30 @@ TEST_CASE_END
...
-    allocator_test::NullableSwappable ns2(a2);
-#else
-    allocator_test::NullableFields ns1(a1);
-    allocator_test::NullableFields ns2(a2);
-#endif
+    NullableFields ns1(a1);
+    NullableFields ns2(a2);
 
     ns1.nullable_uint32.set();
     ns1.nullable_list.set();
```
ara-ayvazyan: The code on line 257 will fail without this fix.
> `ns1.nullable_struct.set();`

PR #877, opened 2018-05-10
[gbc] Fix some Haskell warnings

chwarr:


chwarr:
Looks like the CI build doesn't fail on Haskell warnings...

FYI: @BertusGreeff

PR #878, opened 2018-05-10
[gbc] Treat Haskell warnigns as errors

chwarr:


chwarr:
This needs to be merged after https://github.com/Microsoft/bond/pull/877, as it builds atop that change.
I've verified that without the change in https://github.com/Microsoft/bond/pull/877, the Haskell build would have failed with the addition of `-Werror`

PR #879, opened 2018-05-10
[gbc c++] Generate allocator_type

ara-ayvazyan:


compiler/tests/generated/alias_with_allocator_types.h
```diff
@@ -21,6 +21,8 @@ namespace test
     
     struct foo
     {
+        using allocator_type = arena;
+
         std::list<bool, typename std::allocator_traits<arena>::template rebind_alloc<bool> > l;
```
ara-ayvazyan: We can now use `allocator_type` in all such places instead of using the original name (`arena` in this case). Let me know what you think.

CHANGELOG.md
```diff
@@ -41,6 +41,8 @@ different versioning scheme, following the Haskell community's
   `GetValueToNameMap`.
 * C++ codegen now applies the `--export-attribute` to the `ToString`,
   `FromString`, `ToEnum` and `FromEnum` functions.
+* C++ codegen now generates `allocator_type` typedef for a struct when `--allocator`
```
chwarr: Minor: "*an* `allocator_type` typedef for a struct when *the* `--allocator` *flag* is passed to `gbc`..."

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
```diff
@@ -241,6 +224,10 @@ namespace std
             fieldInit Field {..} = optional (\x -> [lt|#{fieldName}(#{x})|])
                 $ initValue fieldType fieldDefault
 
+        allocatorType alloc = [lt|using allocator_type = #{alloc};
+
+        |]
```
chwarr: Minor: do we need the empty indentation of just the newlines? If just the newline, we can put the `|]` in the first column.

compiler/tests/generated/alias_with_allocator_types.h
chwarr: Seems cleaner to not repeat the name all over the place. It will churn the golden files more, but that's OK.

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
ara-ayvazyan: We need indentation for the last line only.

compiler/tests/generated/alias_with_allocator_types.h
ara-ayvazyan: Will send a separate PR for that.

ara-ayvazyan: This won't work, since we need to use the original name for a generic base type of string/container (e.g. `struct Derived : Base<string>`).

PR #880, opened 2018-05-10
[gbc c++] Use allocator_type in generated code

ara-ayvazyan:


ara-ayvazyan:
This is made on top of the #879, so this needs to be merged after that.

ara-ayvazyan:
This won't work, since we need to use the original name for a generic base type of string/container (e.g. `struct Derived : Base<string>`).

PR #881, opened 2018-05-10
[c++] Cleanup allocator related traits

ara-ayvazyan:


cpp/inc/bond/core/nullable.h
```diff
@@ -21,6 +21,19 @@ namespace bond
 namespace detail
 {
 
+template<typename T, typename Enable = void> struct
+allocator_type
```
ara-ayvazyan: This will be removed in the future change.

cpp/inc/bond/core/stl_containers.h
```diff
@@ -230,12 +230,9 @@ use_map_allocator_for_keys
...
 template <typename T> struct
-use_map_allocator_for_keys
-<T, typename boost::enable_if<
-    std::is_same<typename std::allocator_traits<typename T::allocator_type>::template rebind_alloc<int>,
-    typename std::allocator_traits<typename element_type<T>::type::first_type::allocator_type>::template rebind_alloc<int> > >::type>
-    : std::integral_constant<bool,
-        !detail::is_default_allocator<typename T::allocator_type>::value> {};
+use_map_allocator_for_keys<T, typename boost::enable_if<
+    std::uses_allocator<typename element_type<T>::type::first_type, typename T::allocator_type> >::type>
+    : std::true_type {};
```
chwarr: It looks like with this change, we'll go down allocator paths even if the default allocator is being used. This shouldn't have an observable effect for correctly implemented collections, right?

ara-ayvazyan: Yes, it should not matter if specifically `std::allocator` was used, unless there is a bug in that code.

cpp/inc/bond/core/maybe.h
```diff
@@ -277,7 +287,7 @@ class maybe;
 /// \li operator!=(const detail::maybe_common&,const T&)
 /// \li operator!=(const T&,const detail::maybe_common&)
 template <typename T>
-class maybe<T, typename boost::disable_if<detail::has_allocator<T>>::type>
+class maybe<T, typename boost::disable_if<detail::has_allocator<T> >::type>
```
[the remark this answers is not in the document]
ara-ayvazyan: Not really, this is just an unintentional change due to editing.

PR #882, opened 2018-05-10
[c++ grpc] Simplify generated code of ClientCore

ara-ayvazyan:
This change reduces twice the number of generated functions in the `Client` by making the `grpc::ClientContext` parameter last one.


compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
```diff
@@ -270,50 +246,44 @@ inline #{className}::#{proxyName}<TThreadPool>::#{proxyName}(
 
         methodDecl Function{..} = [lt|#{template}template <typename TThreadPool>
 inline void #{className}::#{proxyName}<TThreadPool>::Async#{methodName}(
-    ::std::shared_ptr< ::grpc::ClientContext> context,
     #{voidParam (methodTypeToMaybe methodInput)}
-    const std::function<void(std::shared_ptr< ::bond::ext::gRPC::unary_call_result< #{payload (methodTypeToMaybe methodResult)}>>)>& cb)
+    const std::function<void(std::shared_ptr< ::bond::ext::gRPC::unary_call_result< #{payload (methodTypeToMaybe methodResult)}>>)>& cb,
```
lalo: missing ::std at two places

ara-ayvazyan: Intentionally did not update to avoid more changes.

PR #884, opened 2018-05-11
[c++ grpc] Use plain unary_call_result

ara-ayvazyan:
This change removes the unnecessary `std::shared_ptr` wrapper when passing `unary_call_result<Response>` to a client callback.

Built on top of #882.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,8 @@ different versioning scheme, following the Haskell community's
 * **Breaking change** When using Bond-over-gRPC, the generated `ClientCore::Async*`
   functions now accept `std::shared_ptr<grpc::ClientContext>`
   as the last parameter instead of as the first.
+* **Breaking change** When using Bond-over-gRPC, the client callback now directly accepts
+  `bond::ext::gRPC::unary_call_result<Response>` (drops the `std::shared_ptr`).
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -78,6 +78,8 @@ different versioning scheme, following the Haskell community's
 * **Breaking change** When using Bond-over-gRPC, the generated `ClientCore::Async*`
   functions now accept `std::shared_ptr<grpc::ClientContext>`
   as the last parameter instead of as the first.
+* **Breaking change** When using Bond-over-gRPC, the client callback now directly accepts
+  `bond::ext::gRPC::unary_call_result<Response>` (drops the `std::shared_ptr`).
```
chwarr: The changelog should also capture the change to `unary_call_result` to now have accessor functions instead of fields.

cpp/inc/bond/ext/grpc/client_callback.h
```diff
@@ -11,50 +11,56 @@
...
+            return _status;
+        }
+
+        const std::shared_ptr<grpc::ClientContext>& context() const BOND_NOEXCEPT
+        {
+            return _context;
+        }
+
+    private:
+        /// @brief The response received from the service.
```
chwarr: These comments should be on the new accessor methods instead of the now-private members.

PR #885, opened 2018-05-11
[c++ grpc] Simplify thread pool concept

ara-ayvazyan:
This change removes the unnecessary need for `TThreadPool` type parameter and instead uses a more simple runtime `std::function` to express a scheduler concept.

Built on top #884.


cpp/inc/bond/ext/grpc/thread_pool.h
```diff
@@ -62,7 +44,7 @@ class thread_pool : private boost::asio::io_service
     /// will be created.
     explicit
     thread_pool(size_t numThreads = USE_HARDWARE_CONC)
-        : _work(*this)
+        : _service{ std::make_shared<service>() }
```
chwarr: Let's split this into two ctor and get rid of the `USE_HARDWARE_CONC` constant completely.

chwarr: As we discussed, so we want to throw if `std::thread::hardware_concurrency()` returns 0?

compiler/src/Language/Bond/Codegen/Cpp/Grpc_h.hs
```diff
@@ -95,14 +95,13 @@ grpc_h export_attribute cpp file imports declarations = ("_grpc.h", [lt|
...
 
-    template <typename TThreadPool>
     class #{proxyName}
     {
     public:
         #{proxyName}(
             const std::shared_ptr< ::grpc::ChannelInterface>& channel,
             std::shared_ptr< ::bond::ext::gRPC::io_manager> ioManager,
-            std::shared_ptr<TThreadPool> threadPool);
+            const ::bond::ext::gRPC::Scheduler& scheduler = {});
```
ara-ayvazyan: Note the `= {}` part.

cpp/inc/bond/ext/grpc/thread_pool.h
```diff
@@ -28,87 +30,69 @@
...
-                const size_t recourseNumThreads = 2;
-                numThreads = recourseNumThreads;
-            }
+            throw InvalidThreadCount{};
         }
 
-        // Spin working threads.
-        for (size_t i = 0; i < numThreads; ++i)
+        auto& service = *_service;
+        for (unsigned long i = 0; i < numThreads; ++i)
```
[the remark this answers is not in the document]
ara-ayvazyan: This meant to be [signed] `long` for the [unlikely] counter overflow.

PR #889, opened 2018-05-13
[c++] Move capped_allocator to ext

ara-ayvazyan:


cpp/inc/bond/ext/capped_allocator.h
```diff
@@ -5,16 +5,16 @@
 
 #include <bond/core/config.h>
 
+#include <bond/core/detail/alloc.h>
```
lalo: minor: should be grouped with the other bond/

chwarr: It go after the final `#include "single_threaded_counter.h" and before the boost include, according to [our guidelines](https://github.com/Microsoft/bond/blob/master/cpp/CODING_GUIDELINES_CPP.md)

PR #888, opened 2018-05-13
 [c++ grpc] Add helper detail::client

ara-ayvazyan:
- Add helper `detail::client` to simplify the generated code.
- Deprecate `wait_callback` and generate additional client functions that return `std::future`.
- Rename `client_callback.h` to `unary_call_result.h`.

PR #891, opened 2018-05-13
 [c++ grpc] Make io_manager_tag non-copyable

ara-ayvazyan:
Built on top of #888.


ara-ayvazyan:
@chwarr, I've reverted the rest and merged with master.

PR #892, opened 2018-05-13
[c++ grpc] Remove unary_call_result::context

ara-ayvazyan:
Built on top of #888.


cpp/inc/bond/ext/grpc/unary_call_result.h
```diff
@@ -24,14 +23,9 @@ namespace bond { namespace ext { namespace gRPC {
         ///
         /// @param response The response.
         /// @param status The status.
-        /// @param context the context under which the request is being executed.
-        unary_call_result(
-            bonded<Response> response,
-            const grpc::Status& status,
-            std::shared_ptr<grpc::ClientContext> context)
```
chwarr: If we remove the context from here, then clients won't be able to get access to the server's metadata unless they explicitly create a context themselves.

cpp/inc/bond/ext/grpc/wait_callback.h
```diff
@@ -33,13 +33,11 @@ template <typename Response>
 class wait_callback final
 {
 public:
-    using arg_type = unary_call_result<Response>;
```
chwarr: Removing this is fine. Breaking change for the changelog, though.

ara-ayvazyan: Sent #899 for that and closing this one.

PR #893, opened 2018-05-14
[c++ grpc] Make call data nested structs

ara-ayvazyan:
Built on top of #891.


cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -35,61 +28,36 @@ class shared_unary_call;
...
-template <typename TRequest, typename TResponse>
-class unary_call final : public detail::unary_call_base<TRequest, TResponse>
+template <typename Request, typename Response>
+class unary_call final : public detail::unary_call_base<Request, Response>
 {
-    using base_type = detail::unary_call_base<TRequest, TResponse>;
+    using base_type = typename unary_call::unary_call_base;
 
 public:
-    /// @brief The default constructor of unary_call is not callable: the
```
chwarr: Can we leave this documentation comment? It's become some what of a FAQ ("Why can't I default construct a `unary_call` instance?"), as move-only types aren't (yet) broadly familiar types. We can put it behind `BOND_DOXYGEN_ONLY`.

cpp/inc/bond/ext/grpc/unary_call.h
```diff
@@ -35,61 +28,36 @@ class shared_unary_call;
...
-    /// assignment.
     unary_call(unary_call&& other) = default;
-
-    /// @brief Assign to this unary_call by moving from another unary_call.
-    ///
-    /// The moved-from instance is valid only for destruction and
-    /// assignment.
     unary_call& operator=(unary_call&& rhs) = default;
 
+    using base_type::operator bool;
```
chwarr: I think this pattern was to get the documentation for the functions to render in the non-detail types instead of just in the base when running Doxygen. We should check if that does what we think it does. If not, `BOND_DOXYGEN_ONLY` or leave this line off.

cpp/inc/bond/ext/grpc/shared_unary_call.h
```diff
@@ -5,64 +5,52 @@
...
-/// the shared owners are destroyed, a generic internal server error is
-/// sent.
-template <typename Request, typename Response>
-class shared_unary_call final : public detail::unary_call_base<Request, Response>
+namespace bond { namespace ext { namespace gRPC
 {
-    using base_type = typename shared_unary_call::unary_call_base;
-
-public:
-    /// @brief Creates an empty shared_unary_call.
```
chwarr: `shared_unary_call` was intentionally designed to be default constructible so that is can be used in places where a default constructible object is wanted or needed. What do we need to do make it default constructible again?

Yes, something like `boost::optional<unary_call>` could be used here, but our users *explicitly asked* for default constructibility.

The motivating use case is as a data member of a struct that holds per-request state, but the data member needs to be created before the service dispatches to the business logic.

How about we make `unary_call` default constructible too?

PR #895, opened 2018-05-14
[c++ grpc] Use marshaling API

ara-ayvazyan:
Built on top of #888.


chwarr:
We'll need to figure out a way to make this change without affecting wire compatibility.

PR #898, opened 2018-05-15
[c++] Replace bond::check_method with decltype

### LINKED ISSUE #896: bond::is_writer does not work for derived types if they don't implement WriteStructBegin
Repro (I copied `check_method` and `is_writer` from bond's headers): https://godbolt.org/g/2F76N9

The way `is_writer` works - it tries to assign member function pointer from `T::WriteStructBegin`. When derived class relies on base's `WriteStructBegin`, this fails because it tries to assign base's pointer to member function to derived one.

I'm also guessing is_writer is not the only problem.

Copy of repro above:
```cpp
#include <type_traits>
#include <boost/type_traits.hpp>

struct Metadata
{};

struct Base
{
    void WriteStructBegin(const Metadata&, bool) {}
};

struct Derived : Base {};

template <typename T, T> struct 
check_method
    : std::true_type {};

template <typename T, typename Enable = void> struct
is_writer
    : std::false_type {};

template <typename T> struct
is_writer<T, typename boost::enable_if<check_method<void (T::*)(const Metadata&, bool), &T::WriteStructBegin>>::type>
     : std::true_type {};

int main()
{
    static_assert(is_writer<Base>::value, "!Base is writer");
    static_assert(is_writer<Derived>::value, "!Derived is writer"); // Fails
}
```

ara-ayvazyan:
Resolves #896.

### DIFF
--- a/cpp/inc/bond/core/detail/mpl.h
+++ b/cpp/inc/bond/core/detail/mpl.h
@@ -17,11 +17,18 @@
 #include <utility>
 
 
-namespace bond
+namespace bond { namespace detail { namespace mpl
 {
 
-namespace detail { namespace mpl
+template <typename...> struct
+make_void
 {
+    using type = void;
+};
+
+template <typename... T>
+using void_t = typename make_void<T...>::type;
+
 
 template <typename T> struct
 identity
@@ -125,6 +132,4 @@ inline auto try_apply(F&& f)
 }
 
 
-}} // namespace mpl { namespace detail
-
-} // namespace bond
+} } } // namespace bond::detail::mpl
--- a/cpp/inc/bond/core/traits.h
+++ b/cpp/inc/bond/core/traits.h
@@ -113,10 +113,11 @@ get_protocol_writer<Reader<Input, MarshaledBondedProtocols>, Output>
     typedef typename Reader<Output, MarshaledBondedProtocols>::Writer type;
 };
 
-
+#if defined(_MSC_VER) && (_MSC_VER < 1900)
 template <typename T, T> struct
 check_method
     : std::true_type {};
+#endif
 
 
 template <typename T> struct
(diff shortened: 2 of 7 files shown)


cpp/inc/bond/core/detail/mpl.h
```diff
@@ -17,11 +17,18 @@
...
 
-namespace detail { namespace mpl
+template <typename...> struct
+make_void
 {
+    using type = void;
+};
+
+template <typename... T>
+using void_t = typename make_void<T...>::type;
```
chwarr: Is `make_void` there to make `void_t` dependent on `T`?

Is the compiler allowed to "inline" `make_void`? If so, how do we prevent it from seeing that `make_void` doesn't depend on `T` and thus neither does `void_t`?

cpp/inc/bond/core/traits.h
```diff
@@ -114,9 +114,11 @@ get_protocol_writer<Reader<Input, MarshaledBondedProtocols>, Output>
 };
 
 
+#ifdef BOND_NO_SFINAE_EXPR
```
chwarr: `check_method` is not in a detail namespace, so putting it in an `ifdef` makes this a breaking change depending on compiler version. Consider leaving it and adding a comment saying that this is no longer used by Bond with newer compilers.

cpp/inc/bond/core/detail/mpl.h
ara-ayvazyan: Following the [`std::void_t`](http://en.cppreference.com/w/cpp/types/void_t)

> Until CWG 1558 (a C++14 defect), unused parameters in alias templates were not guaranteed to ensure SFINAE and could be ignored, so earlier compilers require a more complex definition of void_t, such as
> ```cpp
> template<typename... Ts> struct make_void { typedef void type;};
> template<typename... Ts> using void_t = typename make_void<Ts...>::type;
> ```

PR #899, opened 2018-05-16
[c++ grpc] Remove wait_callback::arg_type

ara-ayvazyan:


cpp/test/grpc/wait_callback.cpp
```diff
@@ -171,7 +171,7 @@ namespace wait_callback_tests
         UT_AssertIsTrue(wasInvoked);
     }
 
-    static void Initialize()
+    void Initialize()
```
chwarr: These functions don't need to be linked with outside of this compilation unit, hence the `static`.

ara-ayvazyan: Anonymous namespace is a better option in that case, but here it is anyway inside of a test namespace, so did not bother adding that.

chwarr: Anyway, when we switch to a Boost auto test, they'll no longer be static.

PR #902, opened 2018-05-16
[c++ grpc] Fix a bug causing thread self join

ara-ayvazyan:


cpp/inc/bond/ext/grpc/detail/client_call_data.h
```diff
@@ -47,20 +47,20 @@ class client_unary_call_data
 public:
     client_unary_call_data(
         std::shared_ptr<grpc::ChannelInterface> channel,
-        std::shared_ptr<io_manager> ioManager,
+        const std::shared_ptr<io_manager>& ioManager, // TODO: Accept std::shared_ptr<grpc::CompletionQueue>
```
chwarr: Should I wait for a change addressing this TODO before merging?

ara-ayvazyan: No, will address it in #888.

PR #903, opened 2018-05-16
[gbc] Remove empty stack.yaml entries

chwarr:


chwarr:
One of the Travis runs has failed, but it isn't a problem. That Travis run was started because I pushed the same commit to a branch on this repository accidentally and then deleted it. Some of those build legs couldn't find the commit after I deleted it.

lalo:
Don't know when to merge this, I don't want to introduce more work for CI.

chwarr:
Now is fine.

PR #904, opened 2018-05-18
[c++ grpc] Replace server_builder with server::Start

ara-ayvazyan:


cpp/inc/bond/ext/grpc/server.h
```diff
@@ -56,17 +56,51 @@
...
+            std::pair<std::string, std::unique_ptr<Services>>... namedServices)
+        {
+            std::vector<std::pair<std::string, std::unique_ptr<detail::service>>> all;
+            std::initializer_list<int>{ (all.emplace_back(std::move(namedServices)), 0)... };
+            return Start(builder, std::move(all));
+        }
+
+        static server Start(
+            grpc::ServerBuilder& builder,
+            std::vector<std::unique_ptr<detail::service>> services)
```
chwarr: If a client of `server::Start` wants to have a vector of heterogeneous services, they have to mention a detail type when creating that vector. Should we extract out the parts that `server` needs into an abstract, non-detail interface?

I think we just need:
* `start`
* `grpc_service`
* `SetCompletionQueue`

This is an odd, clunky interface. Any better suggestions?

cpp/inc/bond/ext/grpc/server.h
```diff
@@ -118,12 +150,63 @@ namespace bond { namespace ext { namespace gRPC
             BOOST_ASSERT(_server);
             BOOST_ASSERT(_ioManager);
 
+            start();
+        }
+
+        void start()
```
chwarr: Nothing else calls start that I can see: keep this look inside the ctor?

cpp/inc/bond/ext/grpc/server.h
```diff
@@ -56,17 +56,51 @@
...
 {
     /// @brief Models a gRPC server powered by Bond services.
     ///
-    /// Servers are configured and started via
-    /// bond::ext:gRPC::server_builder.
+    /// Servers are configured and started via bond::ext:gRPC::server::Start.
     class server final
     {
     public:
+        template <typename... Services>
```
chwarr: Minor: missing Doxygen comments for these functions.

CHANGELOG.md
```diff
@@ -85,24 +85,19 @@ different versioning scheme, following the Haskell community's
...
-  - The `bond::ext::gRPC::server_builder::RegisterService` now only accepts service
-    instances managed by `std::unique_ptr`. This properly models the lifetime requirements.
-  - The `bond::ext::gRPC::server_builder::SetThreadPool` has been removed. Instead, service
-    implementations must now pass a `Scheduler` to the generated `Service` base class which
-    is no longer default constructible.
   - The `client_callback.h` header file has been renamed to `unary_call_result.h`.
+  - The `bond::ext::gRPC::server_builder` has been replaced by `bond::ext::gRPC::server::Start`
+    factory function which now returns plain `bond::ext::gRPC::server` object and accepts
+    service instances managed by `std::unique_ptr`. This properly models the lifetime requirements.
+    Service implementations must now pass a `Scheduler` to the generated `Service` base class which
```
chwarr: Minor: missing bullet for the "Service implementation must now..." entry?

cpp/inc/bond/ext/grpc/server.h
ara-ayvazyan: Added an empty `bond::ext::gRPC::abstract_service`.

cpp/inc/bond/ext/grpc/server.h
ara-ayvazyan: This way looks cleaner.

cpp/inc/bond/ext/grpc/detail/service.h
```diff
@@ -34,8 +34,29 @@
...
 
-class server;
+    namespace detail
+    {
+        class service;
+
+    } // namespace detail
+
+
+    /// @brief Base public class that all Bond gRPC++ services inherit.
```
chwarr: Should this be moved to service.h or some other public header?

ara-ayvazyan: Will move to `abstract_service.h`.

cpp/inc/bond/ext/grpc/service_collection.h
```diff
@@ -0,0 +1,62 @@
...
+        {
+            return _services;
+        }
+
+        template <typename Service>
+        void Add(const boost::optional<std::string>& host, std::unique_ptr<Service> s)
+        {
+            BOOST_STATIC_ASSERT(std::is_base_of<abstract_service, Service>::value);
+            _names.push_back(host);
+            _services.emplace_back(static_cast<detail::service*>(s.release()));
```
ara-ayvazyan: The call to `release` must be after `emplace_back`.

PR #906, opened 2018-05-18
[gbc] Sync `stack build` and CMake Debug config

chwarr:
With this commit, bare `stack build` at the command line and the CMake
Debug configuration use the same settings. This minimizes spurious
rebuilds when developing gbc out of the compiler directory and building
Bond via CMake (C++, Java, Python) or MSBuild (C#).

### DIFF
--- a/compiler/CMakeLists.txt
+++ b/compiler/CMakeLists.txt
@@ -17,14 +17,8 @@ function (add_stack_build target)
         set (options "--test")
     endif()
 
-    # If unspecified, turn off optimizations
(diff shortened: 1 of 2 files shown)


chwarr:
I have a different, more maintainable way to solve this problem. Give me a few minutes...

chwarr:
OK. Ready for review.

chwarr:
Looking into the build failures. Look like we may be hitting https://ghc.haskell.org/trac/ghc/ticket/13866

chwarr:
I am moving to a different team at Microsoft and will not be able to complete this PR before that transition.

PR #907, opened 2018-05-18
[c++ c# grpc] Upgrade to gRPC 1.12.0

chwarr:


chwarr:
Note to merger: the commit message may be incorrect. Check that the version of "1.12.0"

cs/src/grpc/grpc.csproj
```diff
@@ -11,8 +11,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Grpc.Core" Version="1.10.0" />
-    <PackageReference Include="System.Interactive.Async" Version="3.1.1" />
```
lalo: I had this on the back of my mind for a while... thanks

PR #910, opened 2018-05-18
[c#] Convert NuGet test to SDK project

chwarr:


lalo:
NuGet.config file might not be needed, since everything is now all nuget packages are PackageReference and the rest of the References are to Bond only.

PR #922, opened 2018-05-30
Add NuGet publish script to repository

chwarr:


tools/release/Publish-BondNuGetPackages.ps1
```diff
@@ -0,0 +1,116 @@
...
+param
+(
+    [Parameter(Mandatory=$True)]
+    [string]
+    $PackageSource,
+    [Parameter(Mandatory=$True)]
+    [string]
+    $PackageFeed,
+    [switch]
+    $SkipSymbols = $false,
```
lalo: should this be false on default?

chwarr: Let me rename this to OmitSymbols

Yes: being false will include symbols. That's what we want.
PowerShell switches need to default to false, as they're really awkward to set to false. Ideally, this wouldn't be a double negative.

PR #924, opened 2018-06-01
[gbc] Fix ambiguous ocurrence of some

### LINKED ISSUE #925: [gbc] Hackage compiles with cabal, issues may arise
When publishing the bond library to Hackage, errors may arise even if it compiles locally using stack. Mostly since stack freezes dependencies and cabal may end up using different ones depending on configurations. This will cause failing librarys to be uploaded to Hackage like 0.11.0.0, 0.11.0.1 and 0.11.0.2. In theory one should verify it compiles correctly with cabal locally before uploading to hackage or trusting stack.

for example:
https://github.com/Microsoft/bond/pull/924

lalo:
Resolves #925.


lalo:
Do not merge, it is incomplete

lalo:
seems like it is fixed now https://hackage.haskell.org/package/bond-0.11.0.3/reports/

Issue #932, opened 2018-06-16
Linux non-root install failure: Create BOND_ENABLE_BASHCOMPLETION


SoyYoRafa:
Hi,
I would like to install Bond in an environment where I do not have root access. It fails because it tries to create /etc/bash_completion.d/gbc. The library archive happens to be installed afterwards so it is important to install the remaining files. An easy fix would be to set BOND_ENABLE_BASHCOMPLETION=FALSE in cmake, similar to gRPC. 

CMake Error at compiler/cmake_install.cmake:53 (file):
  file INSTALL cannot copy file
  "/home/me/bond_src/build/compiler/gbc.comp" to
  "/etc/bash_completion.d/gbc".
Call Stack (most recent call first):
  cmake_install.cmake:71 (include)

lalo:
Sounds like you have some knowledge, care to submit a PR?

Also before working on a PR, play around with this cmake var https://cmake.org/cmake/help/v3.0/variable/CMAKE_INSTALL_PREFIX.html

`On UNIX one can use the DESTDIR mechanism in order to relocate the whole installation. DESTDIR means DESTination DIRectory. It is commonly used by makefile users in order to install software at non-default location. It is usually invoked like this: make DESTDIR=/home/john install`

It might help if you try installing to a folder were you have write permissions.

SoyYoRafa:
Happy to submit a PR, just following contribution guidelines that request that I post here first. 
The destination is hard coded in compiler/CMakeLists.txt. Presumably because the author wanted it to be picked up automatically by bash. As a result, using CMAKE_INSTALL_PREFIX isn't going to give equivalent behavior. I can try checking write permissions but there isn't a native way in cmake. I can try using execute_process, I would have to play around with it.

`set (completion_dir /etc/bash_completion.d)`

```
set (completion_dir /etc/bash_completion.d)
...
if(EXISTS "${completion_dir}" AND IS_DIRECTORY "${completion_dir}")
    install (FILES ${completion}
        RENAME gbc
        DESTINATION ${completion_dir})
endif()
```

lalo:
Closed with f78a418

PR #933, opened 2018-06-19
Issue #932. Fixes Linux non-root install failure

SoyYoRafa:
Removes install(FILES) for bash completion file in favor of a install(CODE). It runs bash to detech if directory /etc/bash_completion.d is writable by the user installing. If it is, it will cp and chmod the gbc completion file.

It assumes that the user has bash but that is highly likely to be true if the machine contains directory /etc/bash_completion.d.

### DIFF
--- a/compiler/CMakeLists.txt
+++ b/compiler/CMakeLists.txt
@@ -104,7 +104,7 @@ set (test_sources
     tests/TestMain.hs
     ${tests})
 
-set (completion_dir /etc/bash_completion.d)
+set (completion_dir etc/bash_completion.d)
 set (completion ${CMAKE_CURRENT_BINARY_DIR}/gbc.comp)
 set (output ${CMAKE_CURRENT_BINARY_DIR}/build/gbc/gbc${CMAKE_EXECUTABLE_SUFFIX})
 set (GBC_EXECUTABLE ${output} PARENT_SCOPE)
@@ -128,9 +128,6 @@ install (FILES ${output}
     PERMISSIONS OWNER_EXECUTE GROUP_EXECUTE WORLD_EXECUTE
     DESTINATION bin)
 
-if(EXISTS "${completion_dir}" AND IS_DIRECTORY "${completion_dir}")
-    install(CODE "execute_process(COMMAND bash -c \"[[ ! -w ${completion_dir} ]] ||
-        (cp -p ${completion} ${completion_dir}/gbc && 
-         chmod 644 ${completion_dir}/gbc && 
-         echo '-- Installing: ${completion_dir}/gbc')\")")
-endif()
+install (FILES ${completion}
+         RENAME gbc
+         DESTINATION ${completion_dir})
--- a/compiler/CMakeLists.txt
+++ b/compiler/CMakeLists.txt
@@ -129,7 +129,8 @@ install (FILES ${output}
     DESTINATION bin)
 
 if(EXISTS "${completion_dir}" AND IS_DIRECTORY "${completion_dir}")
-    install (FILES ${completion}
-        RENAME gbc
-        DESTINATION ${completion_dir})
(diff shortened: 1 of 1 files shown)


compiler/CMakeLists.txt
```diff
@@ -129,7 +129,8 @@ install (FILES ${output}
     DESTINATION bin)
 
 if(EXISTS "${completion_dir}" AND IS_DIRECTORY "${completion_dir}")
-    install (FILES ${completion}
-        RENAME gbc
-        DESTINATION ${completion_dir})
+    install(CODE "execute_process(COMMAND bash -c \"[[ ! -w ${completion_dir} ]] ||
```
chwarr: What happens if the install destination is "etc/bash_completion.d/" instead of "/etc/bash_completion.d/"? Is CMake smart enough to append the prefix in that case? (I expect the IS_DIRECTORY check will need to go away.) If so, that solution seems more inline with CMake's way of thinking and doesn't rely on bash being installed.

SoyYoRafa: The destination is hardcoded in this CMakeLists.txt to "/etc/bash_completion.d" (line 107). There isn't any cmake magic in changing the path to etc. This is why a non-root install fails.

`set (completion_dir /etc/bash_completion.d)`

chwarr: What happens if you change that line to the following?

```
set (completion_dir etc/bash_completion.d)
```

If the initial slash is removed, CMake should start installing relative to the prefix.

From the CMake documentation for [install](https://cmake.org/cmake/help/v3.0/command/install.html):

> Specify the directory on disk to which a file will be installed. If a full path (with a leading slash or drive letter) is given it is used directly. If a relative path is given it is interpreted relative to the value of the CMAKE_INSTALL_PREFIX variable. The prefix can be relocated at install time using the DESTDIR mechanism explained in the CMAKE_INSTALL_PREFIX variable documentation.

SoyYoRafa: Yes, I agree it will install relative to CMAKE_INSTALL_PREFIX. I limited myself to fix the very narrow issue of non-root installs to the absolute path of "/etc/bash_completion.d". Changing the path to be relative will change the behavior of the install and that was not my goal. When no CMAKE_INSTALL_PREFIX is set, it would install to "/usr/local/etc/bash_completion.d". If you and your team is fine with that, I can send a PR for that.

SoyYoRafa: Just to be thorough, I confirmed that it installs to the relative path of CMAKE_INSTALL_PREFIX if you remove the first slash. Please let me know your preference.

chwarr:
My preference would be to install it under the prefix. If the administrator of the system wants it under /etc, they can symlink it after installing.

compiler/CMakeLists.txt
chwarr: My preference would be to install it under the prefix. If the administrator of the system wants it under /etc, they can symlink it after installing.

lalo:
I'll do a local install on Windows for sanity check.

compiler/CMakeLists.txt
lalo: This won't work on Windows since it is a hard dependency on Bash. You would have to refactor for it to only apply when on Linux.

lalo:
@chwarr CLA bot did not trigger, should I look into it?

chwarr:
Tickle PR for CLA bot testing.

chwarr:
Closing & re-opening to try to trigger automation.

chwarr:
Re-opening to try to trigger automation.

chwarr:
OK. The CLA process has figured it out. When the builds pass, we should be good to merge.

SoyYoRafa:
None that I think of.

lalo:
@SoyYoRafa thank you for your first contribution to Bond!

Issue #934, opened 2018-06-21
Question: is there a reason bond::blob doesn't have a move assignment operator defined?


volka:
Hi! I was wondering why there is no move assignment operator defined for the bond::blob type. The implementation could be quite simple:

`
blob& operator=(blob&& that)
{
    swap(that);
    return *this;
}
`
Is this just an omission or is there a technical reason not to do this? (I would like to create a pull request with the implementation / a test if not)

ara-ayvazyan:
It was missed I guess.
I think the implementation can be as simple as `blob& operator=(blob&& that) = default;`, except for VC12.

lalo:
You can go ahead and submit a PR :+1:

PR #935, opened 2018-06-21
Implement move assignment for the bond::blob type

volka:
Pull request for issue #934. The tests are probably overkill, I can remove them if you want. But they showed that
`
blob& operator=(blob&& that) = default;
`
would copy the internal buffer instead of moving it, so I chose the swap implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ different versioning scheme, following the Haskell community's
 ### C++ ###
 * Fixed a crash in `bond::ext::grpc::server`'s destructor when invoked on
   a moved-from object instance.
+* Added move-assignment operator to `bond::blob`
 
 ## 8.0.0: 2018-05-30 ##
 * `gbc` & compiler library: 0.11.0.0
--- a/cpp/inc/bond/core/blob.h
+++ b/cpp/inc/bond/core/blob.h
@@ -116,6 +116,17 @@ public:
         that._length = 0;
     }
 
+    blob& operator=(blob&& that) BOND_NOEXCEPT_IF(
+        std::is_nothrow_move_assignable<boost::shared_ptr<const char[]> >::value)
+    {
+        _buffer = std::move(that._buffer);
+        _content = std::move(that._content);
+        _length = std::move(that._length);
+        that._content = 0;
+        that._length = 0;
+        return *this;
+    }
+
     blob(const blob& that) = default;
     blob& operator=(const blob& that) = default;
 
--- a/cpp/test/core/blob_tests.cpp
+++ b/cpp/test/core/blob_tests.cpp
@@ -33,6 +33,80 @@ TEST_CASE_BEGIN(LargeBlob)
 }
 TEST_CASE_END
 
+void check_blob_is_empty(const bond::blob& b)
+{
+    // workaround: old compilers won't let us print nullptr, while new ones won't
+    // compare nullptrs to ints
+
+    void* null = 0;
+    UT_AssertAreEqual(b.data(), null);
+    UT_AssertAreEqual(b.size(), 0u);
+    UT_AssertAreEqual(b.content(), null);
+    UT_AssertIsTrue(b.empty());
+}
+
+TEST_CASE_BEGIN(BlobMoveAssign)
+{
+    using bond::blob;
+
+    const uint32_t size = 32;
+
+    // move default-constructed blob
+    {
+        auto data = boost::make_shared<std::array<char,size>>();
+        blob a{data, size};
+        blob b;
+
+        UT_AssertAreEqual(data.use_count(), 2);
+
+        a = std::move(b);
+        
+        check_blob_is_empty(a);
+        check_blob_is_empty(b);
+
+        UT_AssertAreEqual(data.use_count(), 1);
+    }
+    // move from instance with shared ptr
+    {
+        // move from other instance
+        auto data = boost::make_shared<std::array<char, size>>();
+        blob a;
+        blob b{data, size};
+
+        UT_AssertAreEqual(data.use_count(), 2);
+
+        a = std::move(b);
+
+        UT_AssertAreEqual(a.data(), data.get());
+        UT_AssertAreEqual(a.size(), size);
+
+        check_blob_is_empty(b);
+
+        UT_AssertAreEqual(data.use_count(), 2);
+    }
+    // move from raw memory
+    {
+        std::array<char, size> p1;
+        std::array<char, size> p2;
+        blob a{&p1, size};
+        blob b{&p2, size};
+        blob c;
+
+        a = std::move(b);
+        UT_AssertAreEqual(a.data(), &p2);
+        UT_AssertAreEqual(a.size(), size);
+
+        check_blob_is_empty(b);
+
+        // move to empty blob
+        c = std::move(a);
+        UT_AssertAreEqual(c.data(), &p2);
+        UT_AssertAreEqual(c.size(), size);
+
+        check_blob_is_empty(a);
+    }
+}
+TEST_CASE_END
 
 template <typename Reader, typename Writer>
 TEST_CASE_BEGIN(OutputBufferBlobs)
@@ -107,6 +181,9 @@ void BlobTests(const char* name)
 
     AddTestCase<TEST_ID(N),
         OutputBufferBlobs, Reader, Writer>(suite, "OutputBuffer blobs");
+
+    AddTestCase<TEST_ID(N),
+        BlobMoveAssign>(suite, "BlobMoveAssign");
 }
 
 


ara-ayvazyan:
Can you give more details on why  you think `=default` does not work?

cpp/test/core/blob_tests.cpp
```diff
@@ -33,6 +33,124 @@ TEST_CASE_BEGIN(LargeBlob)
...
+{
+    using bond::blob;
+
+    const uint32_t size = 32;
+
+    // test with shared_ptr
+    {
+        // move from other instance
+        blob a;
+        boost::shared_ptr<const char[]> data(new char[size], [](const char* data){delete[] data;});
```
ara-ayvazyan: Use `boost::allocate_shared_noinit`?

lalo:
@volka can you also rebase this change on top of the latest master branch. Your commit should be after 587bf3865fe271fc3ca758ac31e6f6544a98f478 and right now it is after an almost year old commit 1127832811086a013ef064387fae1606c4718e39. This is why we are getting conflicting files.

cpp/test/core/blob_tests.cpp
```diff
@@ -33,6 +33,124 @@ TEST_CASE_BEGIN(LargeBlob)
...
+        UT_AssertAreEqual(b.size(), 32);
+        UT_AssertAreEqual(data.get(), b.data());
+        a = std::move(b);
+        // move guarantees b is still valid but may not have a value
+        UT_AssertAreEqual(a.data(), data.get());
+        UT_AssertAreEqual(b.data(), nullptr);
+        // data and a hold the shared_array, b no longer holds it
+        UT_AssertAreEqual(data.use_count(), 2);
+    }
+    {
```
ara-ayvazyan: Can you simplify and have only below cases:
- default constructed blob,
- blob with shared memory,
- blob with raw memory.

cpp/test/core/blob_tests.cpp
```diff
@@ -33,6 +33,124 @@ TEST_CASE_BEGIN(LargeBlob)
 }
 TEST_CASE_END
 
+TEST_CASE_BEGIN(BlobConstruct)
```
ara-ayvazyan: Add test only for the change you are making (`BlobMoveAssign`).

cpp/inc/bond/core/blob.h
```diff
@@ -105,6 +105,11 @@ class blob
         that._content = 0;
         that._length = 0;
     }
+
+    blob& operator=(blob&& that) {
```
ara-ayvazyan: Move the `{` to the new line.

ara-ayvazyan:
My guess is that `=default` is less desired, since it will leave the raw pointer unchanged and will behave different from what move constructor already does.

volka:
Yes I looked into it and its not correct that it does a copy - the default moves the _buffer but leaves the _content pointer to point to the original storage. That's not wrong but might cause some surprising bugs if someone actually tried to reuse a moved-from blob. I like the swap solution that also swaps _content better. What do you think?

I'll also try to rebase the patch, let's see if that works.

ara-ayvazyan:
Yeah, the `swap` is better and move assignment must behave same way as the move constructor.

cpp/inc/bond/core/blob.h
```diff
@@ -116,6 +116,11 @@ class blob
         that._length = 0;
     }
 
+    blob& operator=(blob&& that) {
```
ara-ayvazyan: Can you add a note in the changelog?

ara-ayvazyan: New line for the `{`.

lalo:
CI is showing an error on line 2217 of the log https://ci.appveyor.com/project/MicrosoftBond/bond/build/1.0.1653/job/jsdywfk10asoxqij#L2217

volka:
Here you go, the AppVeyor problem seems to be the old compiler not wanting to print nullptrs, I hope my workaround fixes that.

I cleaned up the code and testcases as requested, is this better?

cpp/inc/bond/core/blob.h
```diff
@@ -116,6 +116,12 @@ class blob
         that._length = 0;
     }
 
+    blob& operator=(blob&& that)
+    {
+        swap(that);
```
ara-ayvazyan: Moving the state from `this` to `that` does not seem right.

cpp/test/core/blob_tests.cpp
```diff
@@ -33,6 +33,80 @@ TEST_CASE_BEGIN(LargeBlob)
...
+        UT_AssertAreEqual(b.data(), p1.get());
+        UT_AssertAreEqual(b.size(), size);
+
+        // move to empty blob
+        c = std::move(a);
+        UT_AssertAreEqual(c.data(), p2.get());
+        UT_AssertAreEqual(c.size(), size);
+        // our implementation does clear the source, but the standard only requires it must be 
+        // in a valid state
+        UT_AssertAreEqual(a.data(), null);
```
ara-ayvazyan: Can you have a helper function `check_blob_is_empty` which checks for `data`, `size` and `empty`. Standard requires us to keep the object in a state where it can be safely destroyed (in which case the implementation could be just `=default`), but here we want to make the moved-from blob to be empty.

cpp/test/core/blob_tests.cpp
```diff
@@ -33,6 +33,80 @@ TEST_CASE_BEGIN(LargeBlob)
...
+TEST_CASE_BEGIN(BlobMoveAssign)
+{
+    using bond::blob;
+
+    // workaround: old compilers won't let us print nullptr, while new ones won't
+    // compare nullptrs to ints
+    void* null = 0;
+
+    const uint32_t size = 32;
+    struct Big
```
ara-ayvazyan: Simply use `char[size]` or `std::array<char, size>`?

volka:
Ok here is another patch, I simplified the tests and made the move assignment behave exactly like the move construction.

cpp/inc/bond/core/blob.h
```diff
@@ -116,6 +116,16 @@ class blob
         that._length = 0;
     }
 
+    blob& operator=(blob&& that)
```
ara-ayvazyan: Missing `BOND_NOEXCEPT_IF`?

cpp/test/core/blob_tests.cpp
```diff
@@ -33,6 +33,79 @@ TEST_CASE_BEGIN(LargeBlob)
...
 
+void check_blob_is_empty(const bond::blob& b)
+{
+    // workaround: old compilers won't let us print nullptr, while new ones won't
+    // compare nullptrs to ints
+
+    void* null = 0;
+    UT_AssertAreEqual(b.data(), null);
+    UT_AssertAreEqual(b.size(), 0u);
+    UT_AssertAreEqual(b.content(), null);
```
ara-ayvazyan: Add a check for `empty` as well?

cpp/test/core/blob_tests.cpp
```diff
@@ -33,6 +33,79 @@ TEST_CASE_BEGIN(LargeBlob)
...
+        UT_AssertAreEqual(a.data(), data.get());
+        UT_AssertAreEqual(a.size(), size);
+
+        check_blob_is_empty(b);
+
+        UT_AssertAreEqual(data.use_count(), 2);
+    }
+    // move from raw memory
+    {
+        std::unique_ptr<std::array<char, size>> p1(new std::array<char, size>);
```
ara-ayvazyan: Why do you need `unique_ptr` here? Can't you just keep them on the stack (they are small enough)?

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ different versioning scheme, following the Haskell community's
 ### C++ ###
 * Fixed a crash in `bond::ext::grpc::server`'s destructor when invoked on
   a moved-from object instance.
+* Added move-assignment operator to bond::blob
```
ara-ayvazyan: Minor: `bond::blob` (surround with ``).

volka:
Ok, next try, addresses review feedback

cpp/inc/bond/core/blob.h
```diff
@@ -116,6 +116,17 @@ class blob
         that._length = 0;
     }
 
+    blob& operator=(blob&& that) BOND_NOEXCEPT_IF(
+        std::is_nothrow_move_constructible<boost::shared_ptr<const char[]> >::value)
```
ara-ayvazyan: This needs to be `is_nothrow_move_assignable`, not `constructible` :).

ara-ayvazyan: This needs to be `is_nothrow_move_assignable`, not `constructible` :).

volka:
Yes that makes sense, corrected.

ara-ayvazyan:
It would be better to send new commits rather than overwriting the previous one. It makes it harder to review this way.

lalo:
@volka thank you for your first PR to Bond :+1:

volka:
Nice, thanks for merging it :smile:

Issue #434, opened 2017-04-28
[c++] Make is easier to provide a RapidJSON implementation


chwarr:
Currently, Bond C++ uses a fixed RapidJSON version in a submodule. Some consumers of Bond are already using RapidJSON elsewhere. The CMake-based build should have a way to find/provide a different RapidJSON at build time.

SoyYoRafa:
Any views on this? I am interested in getting this done and contributing the change. I was thinking we can create variable FIND_RAPIDJSON which defaults to False. If set to true, it would use find_package. RapidJSON provides a RapidJSONConfig.cmake.

chwarr:
Sounds good to me. @ara-ayvazyan or @lalo, any concerns?

lalo:
Sounds good to me. Thanks for the contributions!

ara-ayvazyan:
Same for me, thanks.

PR #938, opened 2018-06-24
Issue #434. Adds ability to specify RapidJSON library and improves dependency documentation

SoyYoRafa:
-Creates cmake cache variable `BOND_FIND_RAPIDJSON`, defaults to False.
-If `BOND_FIND_RAPIDJSON` is set to true, it will require that RapidJSON package be found. Moreover, it will not install RapidJSON.
-if `BOND_FIND_RAPIDJSON` is set to false, the current behavior is kept where it compiles against the submodule and install the RapidJSON include directory. 
-Improves the documentation of dependencies and their management in README.md.

Tested the following cases in both Linux and macOS:
1. No `BOND_FIND_RAPIDJSON` variable is set. Verified it compiles against the submodule and installs the RapidJSON include directory. Verified I could link against Bond.
2. `BOND_FIND_RAPIDJSON` variable is set to true and no RapidJSON package is found. CMake throws an error and doesn't complete configuration.
3. `BOND_FIND_RAPIDJSON` variable is set to true and RapidJSON package is found. Verified it compiles against the found library and doesn't install the RapidJSON include directory. Verified I could link against Bond.

On a final note, if an user provides a RapidJSON library and doesn't want to build against gRPC, Bond doesn't have to checked out recursively.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ For a discussion how Bond compares to similar frameworks see [Why Bond](https://
 
 ## Dependencies
 
-The Bond repository uses Git submodules and should be cloned with the
+The Bond repository uses Git submodules for its RapidJSON and gRPC dependencies. It should be cloned with the
 `--recursive` flag:
 
 ```bash
@@ -41,6 +41,8 @@ supported bv Visual C++ 2013); a C++11 compiler is required. Additionally,
 to build Bond you will need CMake (3.1+),
 [Haskell Stack](https://docs.haskellstack.org/en/stable/README/#how-to-install)
 (1.5.1+) and Boost (1.61+).
+
+If you already have RapidJSON and would like to build against it, you may add argument `-DBOND_FIND_RAPIDJSON=TRUE` to cmake. It will use find_package(RapidJSON). Otherwise, Bond will also install RapidJSON.
 
 Following are specific instructions for building on various platforms.
 
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ to build Bond you will need CMake (3.1+),
 [Haskell Stack](https://docs.haskellstack.org/en/stable/README/#how-to-install)
 (1.5.1+) and Boost (1.61+).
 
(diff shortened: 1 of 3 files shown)


cmake/Config.cmake
```diff
@@ -180,8 +180,17 @@ cxx_add_compile_options(GNU
...
     ${BOND_GENERATED}
-    ${Boost_INCLUDE_DIRS}
-    ${CMAKE_CURRENT_SOURCE_DIR}/thirdparty/rapidjson/include)
+    ${Boost_INCLUDE_DIRS})
+
+if (BOND_FIND_RAPIDJSON)
+    find_package(RapidJSON REQUIRED)
+    include_directories (
+        ${RapidJSON_INCLUDE_DIRS})
+
```
lalo: nit: remove empty line?

README.md
```diff
@@ -29,19 +29,22 @@ For a discussion how Bond compares to similar frameworks see [Why Bond](https://
...
 [Haskell Stack](https://docs.haskellstack.org/en/stable/README/#how-to-install)
 (1.5.1+) and Boost (1.61+).
 
+Additionally, Bond requires RapidJSON and optionally requires gRPC. The Bond repository primarily uses Git submodules for these two dependencies. It should be cloned with the `--recursive` flag:
+
+```bash
+git clone --recursive https://github.com/Microsoft/bond.git
+```
+
+If you already have RapidJSON and would like to build against it, add argument `-DBOND_FIND_RAPIDJSON=TRUE` to the CMake invocation. It will use find_package(RapidJSON). If you do not provide a RapidJSON library, Bond will also install RapidJSON.
```
lalo: Also add a line on CHANGELOG.md regarding this new flag under C++ header.

cmake/Config.cmake
SoyYoRafa: Agreed. Done.

README.md
SoyYoRafa: Done.

PR #941, opened 2018-07-15
[c++] Fix short-circuiting in ValidateTwoWay

rgardner:
Due to a typo in ValidateTwoWay, if Validate<Protocols>(s1, s2) returns
false, Validate<Protocols>(s2, s1) will be called unnecessarily. This
change enables short-circuiting to avoid the unnecessary right hand side
computation.

### DIFF
--- a/cpp/inc/bond/core/validate.h
+++ b/cpp/inc/bond/core/validate.h
@@ -83,7 +83,7 @@ inline bool Validate(const RuntimeSchema& src,
 template <typename Protocols = BuiltInProtocols, typename T1, typename T2>
 inline bool ValidateTwoWay(const T1& s1, const T2& s2)
 {
-    return Validate<Protocols>(s1, s2) & Validate<Protocols>(s2, s1);
+    return Validate<Protocols>(s1, s2) && Validate<Protocols>(s2, s1);
 }
 
 } // namespace bond


lalo:
Since some tests failed, this could be by design. Can you check which test cases fail and why?

rgardner:
Investigating, they're related to schema validation

rgardner:
These are the tests that are failing on both appveyor and travis:

```
/root/bond/cpp/test/core/validate_tests.cpp(146): error: in "Validation tests/0x20010000_Validate_tests": check (Check<type1, type1_upgrade_fields>(False, Error, Error)) has failed
/root/bond/cpp/test/core/validate_tests.cpp(147): error: in "Validation tests/0x20010000_Validate_tests": check (Check<type1, type1_removed_required_field>(False, Error, Error)) has failed
/root/bond/cpp/test/core/validate_tests.cpp(151): error: in "Validation tests/0x20010000_Validate_tests": check (Check<type4, type4_compatible>(False, Error, Error)) has failed
```

I think these caught a legitimate regression that would be introduced by this change. Take the first one for example: `Check<type1, type1_upgrade_fields>(False, Error, Error)`.

```cpp
struct type1
{
    1: required int16 field1;
    2: int32 field2;
};

struct type1_upgrade_fields
{
    1: required int32 field1;
    2: int64 field2;
};
```

type1_upgrade_fields can deserialize payload generated from type1, but type1 cannot deserialize payloads from type1_upgrade_fields (it would cause truncation). By short-circuiting, this function now returns false and no longer detects that type1 cannot deserialize type1_upgrade_fields. It does seem like the correct behavior is to call both Validate(s1, s2) and Validate(s2, s1).

@lalo, would you like me to document this behavior in `ValidateTwoWay`, split this out into separate lines, or just abandon the PR?

doc:

```
// It is necessary to call both Validate(s1, s2) and Validate(s2, s1), such as when promoting an
// integer field from a smaller size to a larger size. See type1 and type1_upgrade_fields
// in the validation unit tests.
```

separate lines:

```cpp
const bool forwardDeserializable = Validate<Protocols>(s1, s2);
const bool backwardDeserializable = Validate<Protocols>(s2, s1);
return forwardDeserializable && backwardDeserializable;
```

Also, when I run `make check` locally on Ubuntu (Windows Subsystem for Linux), I also don't see the validation test running, though the README suggests that that command should run all tests.

lalo:
Seems that the best course of action is to let it as is :+1: 

Thanks for your work and investigation.

rgardner:
Will do, thanks @lalo

PR #944, opened 2018-09-11
Avoid addressof(*p) when p doesn't alias an object

glenfe:
It is fine for `destroy`, but for `construct` is technically UB to use `addressof(*p)` as `p` does not reference an object that was constructed. See [P0653R2](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0653r2.html) and `std::to_address` in C++2a. This is existing practice in libc++ (`__to_raw_pointer`), libstdc++ (`__to_address`), and Boost (`boost::to_address` since Boost 1.67).

### DIFF
--- a/cpp/inc/bond/core/nullable.h
+++ b/cpp/inc/bond/core/nullable.h
@@ -44,6 +44,19 @@ use_value
         || is_wstring<T>::value
         || !std::is_class<T>::value> {};
 
+template<class T>
+BOND_CONSTEXPR inline T* to_address(T* ptr) BOND_NOEXCEPT
+{
+    return ptr;
+}
+
+template<class Ptr>
+inline typename std::pointer_traits<Ptr>::element_type*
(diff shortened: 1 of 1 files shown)


lalo:
Can you verify if the CI failures are related to the change?

glenfe:
They do not appear to be related. Those Appveyor entries that are failing appear to be failing before even compiling any of the C++ code.

lalo:
Seems like a cmake issue related to this https://cmake.org/cmake/help/git-stage/policy/CMP0074.html

Not caused by your change.

chwarr:
PR #952 is addressing the CMake build problems.

chwarr:
A re-build of this against the current master branch has passed:

* [AppVeyor](https://ci.appveyor.com/project/MicrosoftBond/bond/builds/21537049)
* [Travis](https://travis-ci.org/Microsoft/bond/builds/477609924)

Merging now.

PR #951, opened 2018-12-21
[C# Grpc] Support initializing a grpc service client with a CallInvoker instance.

### LINKED ISSUE #950: Generate Grpc Client with Constructor of setting CallInvoker
Currently, the generated client for Grpc doesn't have a way to set the CallInvoker, it is better to have this.
We are getting CallInvoker instead of Channel when applying client side interceptors to Channel.

YanlongLi:
New feature of Interceptor was added to grpc, and client side
interceptor relies on CallInvoker to work. 
Protobuf has already added the constructor for C#.

See
[L12-csharp-interceptors](https://github.com/grpc/proposal/blob/master/L12-csharp-interceptors.md)
and
[discussion](https://github.com/grpc/grpc/issues/11646#issuecomment-312799633).

Fix #950

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,13 +24,11 @@ different versioning scheme, following the Haskell community's
 ### C# ###
 
 * Updated to gRPC v1.17.1.
-
-### C# ###
-
-* Add a constructor that accepts an `CallInvoker` instance to grpc service client
-  to support client side interceptor. [Issue #950](https://github.com/Microsoft/bond/issues/950).
-  See [L12-csharp-interceptors](https://github.com/grpc/proposal/blob/master/L12-csharp-interceptors.md)
-  and [discussion](https://github.com/grpc/grpc/issues/11646#issuecomment-312799633).
+* Add a constructor that accepts a `CallInvoker` instance to the generated
+  gRPC clients to support client-side interceptors. For more details about
+  C# interceptors, see the [proposal in the gRPC
+  project](https://github.com/grpc/proposal/blob/master/L12-csharp-interceptors.md).
+  [Issue #950](https://github.com/Microsoft/bond/issues/950)
 
 ## 8.0.1: 2018-06-29 ##
 * `gbc` & compiler library: 0.11.0.3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,13 @@ different versioning scheme, following the Haskell community's
 * C++ version: TBD
 * C# NuGet version: TBD
 
+### C# ###
+
+* Add a constructor that accepts an `CallInvoker` instance to grpc service client
+  to support client side interceptor. [Issue #950](https://github.com/Microsoft/bond/issues/950).
+  See [L12-csharp-interceptors](https://github.com/grpc/proposal/blob/master/L12-csharp-interceptors.md)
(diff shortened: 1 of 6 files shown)


YanlongLi:
@ara-ayvazyan @lalo

chwarr:
Changes LGTM. Thanks!

However, the AppVeyor build is broken due to [changes introduced in CMake 3.12](https://cmake.org/cmake/help/latest/policy/CMP0074.html). These need to be addressed before this can be merged, as we cannot merge without a clean build.

I do not have the time to fix the AppVeyor build and I won't have time in the near future either. If you want to see this merged, it would help if you can--in a separate PR--fix the AppVeyor build as well. It looks like the fix is to rename a few CMake variables to not end with `_ROOT`.

YanlongLi:
> 
> 
> Changes LGTM. One minor thing: can you update the changelog to include this new C# functionality?
> 
> However, the AppVeyor build is broken due to [changes introduced in CMake 3.12](https://cmake.org/cmake/help/latest/policy/CMP0074.html). These need to be addressed before this can be merged, as we cannot merge without a clean build.
> 
> I do not have the time to fix the AppVeyor build and I won't have time in the near future either. If you want to see this merged, it would help if you can--in a separate PR--fix the AppVeyor build as well. It looks like the fix is to rename a few CMake variables to not end with `_ROOT`.

I can try to fix the AppVeyor build, though not quite familiar with the CMake stuff.

lalo:
Random question, have you checked that it works? The comment referenced on your commit says that only ```Some of the infrastructure is in place in the C# API```, what does this "some" entail?

chwarr:
> 
> 
> Random question, have you checked that it works? The comment referenced on your commit says that only `Some of the infrastructure is in place in the C# API`, what does this "some" entail?

Full interceptor support was added in v1.10.0. We've been using v1.17.0 in a project I'm currently working on, and interceptors via `CallInvoker` instances are working in all directions. Bond-over-gRPC already needs v1.12.0, so we're good here.

CHANGELOG.md
```diff
@@ -25,6 +25,13 @@ different versioning scheme, following the Haskell community's
 
 * Updated to gRPC v1.17.1.
 
+### C# ###
+
+* Add a constructor that accepts an `CallInvoker` instance to grpc service client
```
chwarr: Consider this phrasing:

```suggestion
* Add a constructor that accepts a `CallInvoker` instance to the generated
  gRPC clients to support client-side interceptors. For more details about
  C# interceptors, see the [proposal in the gRPC
  project](https://github.com/grpc/proposal/blob/master/L12-csharp-interceptors.md).
  [Issue #950](https://github.com/Microsoft/bond/issues/950)
```

CHANGELOG.md
```diff
@@ -25,6 +25,13 @@ different versioning scheme, following the Haskell community's
 
 * Updated to gRPC v1.17.1.
 
+### C# ###
```
chwarr: This C# section can be merged with the existing one and just have two bullets.

```suggestion
```

whisperingchild:
Hi, may I ask when this change will be released? Need it desperately. Thanks.

lalo:
@whisperingchild @YanlongLi a release has been cut you can get it from Nuget.org as usual. For example -> https://www.nuget.org/packages/Bond.CSharp/

whisperingchild:
@lalo Thanks very much.

PR #952, opened 2019-01-07
[AppVeyor Fix] Change environment variable _ROOT to _PATH

YanlongLi:
The AppVeyor build is broken due to
[changes of CMake 3.1.2](https://cmake.org/cmake/help/latest/policy/CMP0074.html).
and Windows Update Service is not properly restarted.
Seting CMP0074 to NEW and configuring Windows Update Service to auto start fix it.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -78,6 +78,8 @@
               BOND_BOOST: "1.63.0"
               BOND_CMAKE_FLAGS: "-DBOND_SKIP_CORE_TESTS=TRUE;-DBOND_SKIP_GBC_TESTS=TRUE"
     install:
+        - sc config wuauserv start= auto
+        - net start wuauserv
         - ps: >-
             if (-not $env:BOND_VS_VERSION)
             {
--- a/cmake/Config.cmake
+++ b/cmake/Config.cmake
@@ -1,5 +1,9 @@
 include (Compiler)
 
+if (POLICY CMP0074)
+    cmake_policy(SET CMP0074 NEW)
+endif()
+
 set (BOND_GBC_PATH_DESCRIPTION
      "Optional path to the gbc executable to use. If set, this gbc will be used when generating code from .bond files. If not set, then gbc will be built (and the Haskell toolchain will need to be present on the machine) and the gbc tests will be run.")
 
--- a/compiler/CMakeLists.txt
+++ b/compiler/CMakeLists.txt
@@ -1,5 +1,9 @@
 cmake_minimum_required (VERSION 2.8.12)
 
+if (POLICY CMP0074)
+    cmake_policy(SET CMP0074 NEW)
+endif()
+
 set (CMAKE_MODULE_PATH
     ${CMAKE_CURRENT_SOURCE_DIR}/../cmake)
 


appveyor.yml
```diff
@@ -31,7 +31,7 @@
 
     environment:
         global:
-            STACK_ROOT: "c:\\sr"
+            STACK_PATH: "c:\\sr"
```
chwarr: `STACK_ROOT` is a [special environment variable for the stack tool](https://docs.haskellstack.org/en/stable/GUIDE/#setting-stack-root-location). We've set it to something short, as we have run into `MAX_PATH` issues in the past. It looks like stack and CMake are going to conflict. My proposal would be to change this to something like `STACK_ROOT_LOCATION` and use the `BOND_STACK_OPTIONS` CMake variable to set this when we invoke CMake by [adding it to `$cmakeFlags` around line 202](https://github.com/Microsoft/bond/blob/bd4b46e78a82dd3a38b52b00233a1a22047014d2/appveyor.yml#L202):

    $cmakeFlags += "-DBOND_STACK_OPTIONS=${env:STACK_ROOT_LOCATION}"

appveyor.yml
```diff
@@ -78,6 +78,8 @@
               BOND_BOOST: "1.63.0"
               BOND_CMAKE_FLAGS: "-DBOND_SKIP_CORE_TESTS=TRUE;-DBOND_SKIP_GBC_TESTS=TRUE"
     install:
+        - sc config wuauserv start= auto
```
chwarr: Interesting... How did you figure out that the Windows Update service was causing problems?

YanlongLi: I got error message "ERROR: Update KB3035131 installation failed (exit code 1058)." in the build log,
Issue was addressed here cocur/slugify#193

chwarr: Weird... Thanks for the details.

appveyor.yml
```diff
@@ -132,6 +134,9 @@
 
                 $env:BOOST_ROOT = $boostLoc['BOOST_ROOT']
                 $env:BOOST_LIBRARYDIR = $boostLoc['BOOST_LIBRARYDIR'][$env:BOND_ARCH]
+
+                # SET(CMAKE_INCLUDE_PATH ${CMAKE_INCLUDE_PATH} $env:BOOST_ROOT)
```
chwarr: Let's remove this commented out code.

chwarr:
The current warnings are due to boringssl's use of `/Wall` and the new off-by-default [C5045](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/c5045?view=vs-2017). This was fixed in boringssl in commit 189270cd190267f5bd60cfe8f8ce7a61d07ba6f4, which was picked up in gRPC v1.14.0.

Let me take your work here and try to upgrade to the current version of gRPC, 1.17.0.

appveyor.yml
YanlongLi: Removed, seems we don't need to change _ROOT suffix, don't know why but it seems works, PointCloudLibrary/pcl/issues/2425

YanlongLi:
> 
> 
> The current warnings are due to boringssl's use of `/Wall` and the new off-by-default [C5045](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/c5045?view=vs-2017). This was fixed in boringssl in commit 189270cd190267f5bd60cfe8f8ce7a61d07ba6f4, which was picked up in gRPC v1.14.0.
> 
> Let me take your work here and try to upgrade to the current version of gRPC, 1.17.0.

Cool.

YanlongLi:
Fix will be included in #953

PR #965, opened 2019-04-03
Remove docker login from Travis build

chwarr:
The bondciimages.azurecr.io has been configured for anonymous read-only
access, so the `docker login` step is no longer needed.


lalo:
@chwarr travis is failing to pull, is this the intention and later patch it with correct settings? https://travis-ci.org/Microsoft/bond/jobs/514944554

chwarr:
Nope. The failure to pull indicate something odd is going on with the repository. I need to follow up with the ACR team.

chwarr:
@lalo: ready for squash merge.

PR #972, opened 2019-05-07
Switch to microsoft/bond casing

chwarr:
Microsoft has renamed its GitHub organization from "Microsoft" to
"microsoft". This commit fixes various paths, as not all tools are able
to follow the Microsoft -> microsoft redirects that are in place.

These replacements were made with the following script:

    for f in $(grep -l -r -e 'github.com/Microsoft' -e 'Microsoft/bond' -e 'Microsoft.github.io' *); do
        sed -i -e 's~github.com/Microsoft~github.com/microsoft~g; s~Microsoft/bond~microsoft/bond~g;s~Microsoft.github.io~microsoft.github.io~g' "$f"
    done


sapek:
LGTM

PR #975, opened 2019-05-20
Move maybe::operator== back from maybe_common to maybe implementations

volka:
The move of the maybe::operator== from the two variants to the common
base maybe_common causes an ambiguity with the allocator operator==
from libstdc++ in GCC 9.1. Moving them back to the variants fixes this
issue.

More detailed description:

When compiling Bond 8.1 code with gcc 9.1.0 / the current libstdc++, I get an ambiguous overload error for the

    friend bool operator==(const maybe_common& lhs, const maybe_common& rhs)

from maybe.h line 195. The ambiguous other operator being

    operator==(const allocator<_T1>&, const allocator<_T2>&)

from allocator.h:168 (and another similar one from new_allocator.h).

Moving the shared operator== implementation from maybe_common to both of the maybe variants fixes the issue, but I'm not sure if this is the best approach. See the attached patch. I'm also not sure what else would need to be updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ different versioning scheme, following the Haskell community's
 * `gbc` & compiler library: TBD
 * IDL core version: TBD
 * C++ version: minor bump needed
+* Fixed ambiguous call to maybe::operator== that breaks GCC 9 build
 * C# NuGet version: TBD
 
 ### C++ ###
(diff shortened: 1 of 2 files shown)


chwarr:
This looks fine to me. Can you add a note to the changelog in the unreleased C++ section?

chwarr:
I appear to have broken the CI builds (the images we use are no longer present). I've opened issue #979 about this. Until that's resolved, this cannot be merged. I don't expect @volka to fix this however, since it's an infrastructure problem unrelated to this change.

PR #976, opened 2019-05-20
Use new `az acr repository delete --image` syntax

chwarr:
The old syntax for deleting by a manifest no longer works with newer
versions of the az CLI tool. This switches to the newer `--image
repository@manifest` syntax.

This should fix the [image cleanup job failures](https://dev.azure.com/msazure/One/_build?definitionId=21361) that we've been seeing.


chwarr:
Thanks. The clean up [job](https://dev.azure.com/msazure/One/_build/results?buildId=22288112) is passing now.

PR #986, opened 2019-07-07
Minor proposal on string 114

kant:


chwarr:
This is just the README, so the build failures aren't relevant. Merging.
Thank you.

PR #987, opened 2019-07-08
Improve OutputBuffer Grow method for large buffers

ConradBrean:
Current Grow implementation works inefficiently for
buffer sizes larger than 2/3 of int.MaxValue. After that size,
each grow will result in small incremental resize to the next
required count, copying huge amounts of data each time.

### DIFF
--- a/cs/src/core/io/safe/OutputBuffer.cs
+++ b/cs/src/core/io/safe/OutputBuffer.cs
@@ -220,9 +220,14 @@ namespace Bond.IO.Safe
         // Grow the buffer so that there is enough space to write 'count' bytes
         internal virtual void Grow(int count)
         {
-            var minLength = position + count;
+            int minLength = position + count;
             length += length >> 1;
-            if (length < minLength) length = minLength;
+
+            const int ArrayIndexMaxValue = 0x7FFFFFC7;
+            if ((uint)length > ArrayIndexMaxValue)
+                length = ArrayIndexMaxValue;
+            if (length < minLength)
+                length = minLength;
 
             Array.Resize(ref buffer, length);
         }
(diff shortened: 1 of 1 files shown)


cs/src/core/io/safe/OutputBuffer.cs
```diff
@@ -220,9 +220,14 @@ public virtual void WriteString(Encoding encoding, string value, int size)
         // Grow the buffer so that there is enough space to write 'count' bytes
         internal virtual void Grow(int count)
         {
-            var minLength = position + count;
+            int minLength = position + count;
             length += length >> 1;
-            if (length < minLength) length = minLength;
+
+            const int ArrayIndexMaxValue = 0x7FFFFFC7;
+            if ((uint)length > ArrayIndexMaxValue)
```
chwarr: Minor: The Bond code base's style is to use braces for all multi-line conditions. Can you reformat this?

```suggestion
            if ((uint)length > ArrayIndexMaxValue)
            {
                length = ArrayIndexMaxValue;
            }
            else if (length < minLength)
            {
                length = minLength;
            }
```

ConradBrean:
So, when is this going to be merged?

chwarr:
The CI build is currently failing due to a problem during the set up phase. (More details in https://github.com/microsoft/bond/issues/989) That needs to be resolved before we can merge code changes into master again.

Due to other commitments I have inside of Microsoft, I will not have time to figure out a fix for the next 6 weeks at least.

When the CI build gets fixed, I'll merge this in. You don't need to do anything more here.

chwarr:
Closing & re-opening to trigger another CI build after CI fixes.

PR #993, opened 2019-09-03
[gbc] Upgrading to stack lts-14.4

kapilash:
stack lts-14.4 uses , among others, version 7 of megaparsec and version 0.8.7 of aeson-pretty.

These versions of library have introduced some breaking changes in their respective API and this commit makes the corresponding changes in bond library. 

Also, stack lts-14.4 does not have `derive` package listed.  The `derive` package was used in test cases for deriving instances of `Arbitrary` for Bond types. This commit replaces the package with `quickcheck-arbitrary-template`. 

references
1. [Megaparsec Changelog](https://github.com/mrkkrp/megaparsec/blob/master/CHANGELOG.md)
2. [Megaparsec 7 changes](https://markkarpov.com/post/megaparsec-7.html)
3. [quickcheck-arbitrary-template](https://hackage.haskell.org/package/quickcheck-arbitrary-template)


kapilash:
Looks like we need to do stack upgrade on the CI machines

lalo:
@kapilash the previous error on CI was a 404 since lts-14.4 was not available when the test ran (they actually added it 13 hours ago see -> https://github.com/commercialhaskell/lts-haskell/commit/dff1d65b19702600dfb4adbdce462c816c9b2c1f).

It's now failing due to another error, haven't had the chance to look at it.

lalo:
I wonder if the problem is reproducible on a more recent version of stack, image used currently uses 1.9.3 instead of 2.1.3

lalo:
@kapilash looks like it works on windows, we hit an intermittent issue that is being tracked by stack/ghc folks already. Doc build is failing for other reasons which in theory will be fixed with https://github.com/microsoft/bond/pull/994 . I'm guessing that if we update the stack version on linux to 2.1.3, we might get a successful build.

kapilash:
> @kapilash looks like it works on windows, 

thanks @lalo . I will look into the gcc failure coming on linux

lalo:
There might be something wrong with the compiler haskell tests, see this https://ci.appveyor.com/project/MicrosoftBond/bond/builds/27164266/job/bnvjhcgalqxubkdu#L3280

Saying gbc-tests is not a target.

kapilash:
> 
> 
> There might be something wrong with the compiler haskell tests, see this https://ci.appveyor.com/project/MicrosoftBond/bond/builds/27164266/job/bnvjhcgalqxubkdu#L3280
> 
> Saying gbc-tests is not a target.

It seems like an outcome of one decision I took so as to not disturb the code too much.
Current code has [IO.hs](https://github.com/microsoft/bond/blob/master/compiler/IO.hs) and [Options.hs](https://github.com/microsoft/bond/blob/master/compiler/Options.hs) in the root folder and these two are used both by the app as well as the tests.

I added the '.' and tests in the [package.yaml](https://github.com/kapilash/bond/blob/cf5d9794f2020fac6d7837700205663349de7ea0/compiler/package.yaml#L78)

I am not sure why it does not fail locally

kapilash:
@lalo I think I understand what is going on.
 It seems like, stack gives out this error message (not yet sure why). However, it still goes about building tests and running them.
The build system is parsing the output and finds this message from stack and considers it an error.

compiler/bond.cabal
```diff
@@ -1,153 +1,140 @@
...
-                    base >= 4.9 && < 5,
-                    bytestring >= 0.10,
-                    cmdargs >= 0.10.10,
-                    process < 1.7,
-                    directory >= 1.1,
-                    filepath >= 1.0,
-                    monad-loops >= 0.4,
-                    text >= 0.11,
-                    megaparsec >= 6.2 && < 7
+test-suite bond-test
```
lalo: Might be failing because there was a name change of the test suite from gbc-tests to bond-tests and cmake is expecting gbc-tests. See https://github.com/microsoft/bond/blob/master/compiler/CMakeLists.txt#L126

gbc-tests would be a better name

lalo:
Try to rebase/cherry-pick this commit https://github.com/microsoft/bond/pull/995

kapilash:
The linux failure is caused by a bug at the ghc end: 
 https://gitlab.haskell.org/ghc/ghc/issues/15742

kapilash:
Here's the status as of now:

**Windows**: Doc build is failing and could get fixed by #994   (https://github.com/microsoft/bond/pull/993#issuecomment-527727692)

**Linux** ghc installation fails because of https://gitlab.haskell.org/ghc/ghc/issues/15742
To workaround this I added ghc installation to the docker file and configured stack to use system-ghc where available.

tools/ci-scripts/linux/image-builder/Dockerfile
```diff
@@ -49,5 +49,10 @@ RUN add-apt-repository ppa:cwchien/gradle && \
     gradle \
     openjdk-8-jdk-headless
 
+# Components for ghc
+RUN add-apt-repository ppa:hvr\ghc -y && \
+    apt-get update && \
+    apt-get -y install ghc-8.6.5
+
```
kapilash: @lalo is this an acceptable workaround for ghc installation issue during stack build?

compiler/tests/Tests/Syntax.hs
```diff
@@ -170,8 +170,10 @@ verifySchemaDef baseName schemaName =
         (Bond _ _ declarations) <- parseBondFile [] $ "tests" </> "schema" </> baseName <.> "bond"
         let schema = fromJust $ find ((schemaName ==) . declName) declarations
         return $
-            -- some versions aeson encode angle brackets
             BL.fromStrict $
+            -- aeson-pretty-print prints anything above 1e19 in scientific notation
+            replace "1.8446744073709551615e19" "18446744073709551615" $
```
kapilash: In theory, I should be able to use [Custom Constructor](http://hackage.haskell.org/package/aeson-pretty-0.8.7/docs/Data-Aeson-Encode-Pretty.html#t:NumberFormat) but this seems simpler change for the test.
I can make the change, if the ghc installation work around is acceptable.

tools/ci-scripts/linux/image-builder/Dockerfile
lalo: Let's give it a shot to unblock the rest of this PR, but give me some time.

compiler/stack.yaml
```diff
@@ -28,18 +28,9 @@ resolver: lts-7.24
 packages:
 - .
 
-flags:
-  bond:
-    warningsAsErrors: true
```
chwarr: This was needed so that we could build bond as part of CI builds treating warnings as errors but not to force that on everyone else who may use the Hackage package. Is there a replacement to this functionality?

compiler/bond.cabal
```diff
@@ -1,153 +1,140 @@
...
+    , mtl
+    , process
+    , scientific
+    , shakespeare
+    , text
+    , unordered-containers
+  default-language: Haskell2010
 
-  if os(windows) && arch(i386)
-    ld-options:     -Wl,--dynamicbase -Wl,--nxcompat -Wl,--large-address-aware
```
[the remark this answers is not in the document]
kapilash: Oops, I forgot all about these options. My bad. I will handle this in my next iteration.

compiler/stack.yaml
kapilash: I've added it in the package.yaml as, I thought, `-Wall -Werror` are  recommended. However, I understand we want to keep it configurable. I will make it so in the next iteration.

compiler/tests/Tests/Syntax.hs
kapilash: cc: @chwarr  and @lalo 

Do you have any suggestions on the right way to go about this? I will be happy to make the change that way.


reference:  NumberFormat pretty printing in aeson-pretty http://hackage.haskell.org/package/aeson-pretty-0.8.7/docs/Data-Aeson-Encode-Pretty.html#t:NumberFormat
Standard generated file: https://github.com/microsoft/bond/blob/3132eb2b90d29d916e703e42ec6482b9b89a86ce/compiler/tests/generated/schemadef/defaults.Foo.json#L249

compiler/package.yaml
```diff
@@ -103,3 +117,13 @@ tests:
...
     - megaparsec
     - quickcheck-arbitrary-template
+    when:
+      - condition: os(windows) && arch(i386)
+        ld-options:     -Wl,--dynamicbase -Wl,--nxcompat -Wl,--large-address-aware
+      - condition: os(windows) && arch(x86_64)
+        ld-options:     -Wl,--dynamicbase -Wl,--nxcompat -Wl,--high-entropy-va
+      - condition: flag(warningsAsError)
+        then:
+          ghc-options:    -threaded -Wall -Werror -rtsopts -with-rtsopts=-N
```
chwarr: Minor: odd indentation inconsistency in the `then` and the `else` cases.

chwarr: Minor: if we put the extra bits at the end, then it's easier to keep the two lines in sync:

```
  then:
    ghc-options: -threaded -rtsopts -with-rtsopts=-N -Wall -Werror
  else:
    ghc-options: -threaded -rtsopts -with-rtsopts=-N
```

compiler/package.yaml
```diff
@@ -42,45 +42,59 @@ dependencies:
...
+  when:
+   - condition: os(windows) && arch(i386)
+     ld-options:     -Wl,--dynamicbase -Wl,--nxcompat -Wl,--large-address-aware
+   - condition: os(windows) && arch(x86_64)
+     ld-options:     -Wl,--dynamicbase -Wl,--nxcompat -Wl,--high-entropy-va
+   - condition: flag(warningsAsError)
+     then:
+         ghc-options:    -threaded -Wall -Werror
+     else:
+         ghc-options: -threaded
```
chwarr: It looks like the two rtsopts got lost during this transition.

compiler/tests/Tests/Syntax.hs
chwarr: What you have here is fine by me. Both are valid JSON representations of the same number.

.travis.yml
```diff
@@ -47,7 +47,7 @@ env:
     - FLAVOR="cpp-grpc-master" BOOST="1.66.0" COMPILER="clang" CRON_ONLY="true"
 script:
     - if [ "$CRON_ONLY" = "true" ] && [ "$TRAVIS_EVENT_TYPE" != "cron" ] ; then echo "Skipping cron-only job"; exit 0; fi
-    - CI_BUILD_IMAGE=bondciimages.azurecr.io/ubuntu-1604:build-22756092
+    - CI_BUILD_IMAGE=bondciimages.azurecr.io/ubuntu-1604:build-24749007
```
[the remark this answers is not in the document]
kapilash: I need to wait for this #997  to go in and rebase on it ?

compiler/bond.cabal
```diff
@@ -1,153 +1,166 @@
...
-                    process < 1.7,
-                    directory >= 1.1,
-                    filepath >= 1.0,
-                    monad-loops >= 0.4,
-                    text >= 0.11,
-                    megaparsec >= 6.2 && < 7
+test-suite gbc-tests
+  type: exitcode-stdio-1.0
+  main-is: TestMain.hs
+  other-modules:
```
lalo: Missing IO and Options according to the compiler:

```
Warning: The following modules should be added to exposed-modules or other-modules in C:\projects\bond\compiler\bond.cabal:
               - In test:gbc-tests:
                   IO
                   Options
               - In test:gbc-tests:
                   IO
                   Options
           
```

.travis.yml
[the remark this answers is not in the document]
chwarr: Probably easier to just merge with it. We squash PRs into a single commit the vast majority of the time, so you don't need to worry about keeping a pretty history.

compiler/bond.cabal
kapilash: Shall I move these two to be a part of the library? 
Both of these modules functions that can be very useful for plugin developers

lalo: Hmm in my perspective we should just keep this PR to be a shift in the LTS and avoiding any regression. Pointed this out since I felt they were removed by mistake when moving things around see (plus compiler warning) https://github.com/microsoft/bond/pull/993/files#diff-8865cab0d973aeaba7290a70bfd3dcefL91

kapilash: Oops! I misuderstood the comment.
Fixed the warning.

kapilash:
Is any change needed from my end?

lalo:
Use this image to test the travis-ci errors, bondciimages.azurecr.io/ubuntu-1604:build-25186859 

This image was created with this branch https://github.com/microsoft/bond/commit/9f491669d2cd85a092772e2ea8591daae4f2f40d
 
After that works, we have to figure out the order of operations to get a valid image. But that's on my side.

kapilash:
For some reason it still is trying to install ghc. From the logs:

>> 2019-09-21 02:56:16.477629: [debug] Asking for a supported GHC version

>>  2019-09-21 02:56:16.477682: [debug] Getting system compiler version

>>  2019-09-21 02:56:16.479453: [debug] Installed tools:

>>   - ghc-8.0.1

tools/ci-scripts/linux/image-builder/Dockerfile
kapilash: Looking at the logs of the [image build](https://msazure.visualstudio.com/One/_build/results?buildId=25186859), ghc seems to get installed in /opt. And it is not getting into the docker. Am not familiar enough with docker and the scripts to fix this mismatch.

lalo:
This is from the log that creates the image

```reading package lists...
Building dependency tree...
Reading state information...
The following NEW packages will be installed:
  ghc-8.6.5
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
Need to get 73.3 MB of archives.
After this operation, 793 MB of additional disk space will be used.
Get:1 http://ppa.launchpad.net/hvr/ghc/ubuntu xenial/main amd64 ghc-8.6.5 amd64 8.6.5-9~16.04 [73.3 MB]
debconf: delaying package configuration, since apt-utils is not installed‌
Fetched 73.3 MB in 4s (18.2 MB&#x2F;s)‌
Selecting previously unselected package ghc-8.6.5.
(Reading database ... 
(Reading database ... 5%
(Reading database ... 10%
(Reading database ... 15%
(Reading database ... 20%
(Reading database ... 25%
(Reading database ... 30%
(Reading database ... 35%
(Reading database ... 40%
(Reading database ... 45%
(Reading database ... 50%
(Reading database ... 55%
(Reading database ... 60%
(Reading database ... 65%
(Reading database ... 70%
(Reading database ... 75%
(Reading database ... 80%
(Reading database ... 85%
(Reading database ... 90%
(Reading database ... 95%
(Reading database ... 100%
(Reading database ... 27989 files and directories currently installed.)
Preparing to unpack .../ghc-8.6.5_8.6.5-9~16.04_amd64.deb ...
Unpacking ghc-8.6.5 (8.6.5-9~16.04) ...
Processing triggers for libc-bin (2.23-0ubuntu11) ...
Setting up ghc-8.6.5 (8.6.5-9~16.04) ...
update-alternatives: using /opt/ghc/bin/ghc-8.6.5 to provide /opt/ghc/bin/ghc (opt-ghc) in auto mode
Processing triggers for libc-bin (2.23-0ubuntu11) ...
Removing intermediate container 8d242b551d46
 ---> 214182091712```

ghc-8.6.5 should be there

**Comment by chwarr** [comment]
I see ghc 8.6.5 is getting installed to /opt inside the Docker image, at least looking at the image that @lalo produced, but it's not on the path, and neither is the alternative symlink at `/opt/ghc/bin/ghc`. Do we need 

```
$ docker run \
>     --entrypoint=/usr/bin/find \
>     bondciimages.azurecr.io/ubuntu-1604:build-25186859 \
>     / -type f -executable -name ghc
/opt/ghc/8.6.5/lib/ghc-8.6.5/bin/ghc
$ docker run \
>     --entrypoint=/bin/sh \
>     bondciimages.azurecr.io/ubuntu-1604:build-25186859 \
>     -c "which ghc || echo 'no ghc on path'"
no ghc on path
$ docker run \
>     --entrypoint=/opt/ghc/bin/ghc \
>     bondciimages.azurecr.io/ubuntu-1604:build-25186859 \
>     --version
The Glorious Glasgow Haskell Compilation System, version 8.6.5
```

What happens if we add `/opt/ghc/bin/` to `PATH` before running stack or CMake?

**Comment by kapilash** [comment]
> What happens if we add /opt/ghc/bin/ to PATH

Thanks! I tried your suggestion with just `build_java.zsh` and the [ghc issue is gone!](https://travis-ci.org/microsoft/bond/jobs/589221881)

File: tools/ci-scripts/linux/build_cs.zsh [hunk 10]
```diff
@@ -5,6 +5,8 @@ set -eux
 nuget update -self
 nuget restore $BOND_ROOT/cs/cs.sln
 
+export PATH=/opt/ghc/bin:$PATH
```
chwarr: This can go in build.zsh around the same place as the `BOOST_ROOT` variable is set. That way, we don't need to repeat it in all the subscripts.

kapilash: done

PR #994, opened 2019-09-04
Fix doc CMakeLists

lalo:


lalo:
This is ready to merge, this change will just fix the doc building step but not all of the issues on CI.

Issue #989, opened 2019-07-08
Windows CI builds failing -- happy cannot be installed


chwarr:
We're seeing Windows CI build failures [across][1] various [PRs][2]. It looks like we can no longer build the happy library using current versions of Stack on Windows. This failure is currently blocking PRs.

```
happy               > [1 of 2] Compiling Main             ( C:\Users\appveyor\AppData\Local\Temp\1\stack1640\happy-1.19.5\Setup.lhs, C:\Users\appveyor\AppData\Local\Temp\1\stack1640\happy-1.19.5\.stack-work\dist\b7fec021\setup\Main.o )
  happy               > 
  happy               > C:\Users\appveyor\AppData\Local\Temp\1\stack1640\happy-1.19.5\Setup.lhs:20:1: warning: [-Wtabs]
  happy               >     Tab character found here, and in 39 further locations.
  happy               >     Please use spaces instead.
  happy               > [2 of 2] Compiling StackSetupShim   ( C:\sr\setup-exe-src\setup-shim-Z6RU0evB.hs, C:\Users\appveyor\AppData\Local\Temp\1\stack1640\happy-1.19.5\.stack-work\dist\b7fec021\setup\StackSetupShim.o )
  happy               > Linking C:\Users\appveyor\AppData\Local\Temp\1\stack1640\happy-1.19.5\.stack-work\dist\b7fec021\setup\setup.exe ...
  happy               > Configuring happy-1.19.5...
  happy               > build
  happy               > Building happy-1.19.5...
  happy               > Preprocessing executable 'happy' for happy-1.19.5...
  happy               > setup.exe: The program 'happy' is required but it could not be found
```

[1]: https://ci.appveyor.com/project/MicrosoftBond/bond/builds/25541874/job/n1o2jjblql7evutx
[2]: https://ci.appveyor.com/project/MicrosoftBond/bond/builds/25806187/job/6uvruyl7m6w4ejus

chwarr:
I will not have time to investigate this failure myself for the next few weeks. If someone else has time to figure this out, please submit a PR.

lalo:
This might fix it https://ci.appveyor.com/project/MicrosoftBond/bond/builds/27317758

lalo:
This looks like it fixed it but broke the python build for some reason, https://ci.appveyor.com/project/MicrosoftBond/bond/builds/27320838/job/uxlb9k37dkok23y4

I'll do a PR

PR #996, opened 2019-09-10
[CI] Happy workaround

lalo:
- Install GHC
- Use cabal to install Happy

This is to fix https://github.com/microsoft/bond/issues/989

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -232,13 +232,19 @@
 
             }
 
+            if ($env:BOND_ARCH -eq 64) {
+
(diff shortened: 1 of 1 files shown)


chwarr:
It looks like this needs to be merged with the tip of master to pick up the doc build fix.
Also, something's busted with the Python unit tests it appears. @lalo, can you dig into those?

lalo:
@chwarr looks like it was picking up some mingw python version brought in by one of the new dependencies. The merge build will likely succeed.

PR #997, opened 2019-09-13
Refresh Travis CI build image

chwarr:
The Travis CI image now includes happy pre-installed, as the Haskell LTS we
use needs happy pre-installed.

This image was built from commit dd105d96b64884c1297286a0023d726db06ec1a2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ env:
     - FLAVOR="cpp-grpc-master" BOOST="1.66.0" COMPILER="clang" CRON_ONLY="true"
 script:
     - if [ "$CRON_ONLY" = "true" ] && [ "$TRAVIS_EVENT_TYPE" != "cron" ] ; then echo "Skipping cron-only job"; exit 0; fi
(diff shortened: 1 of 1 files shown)


kapilash:
do we need to have happy on the image if we are using latest lts? I tried on a couple of different machines - stack manages to install happy where needed.

lalo:
@kapilash we can remove all the extra happy installs if they are not needed using the latest lts, that would be the best since installing happy separately is more of a workaround.

kapilash:
> @kapilash we can remove all the extra happy installs if they are not needed using the latest lts, that would be the best since installing happy separately is more of a workaround.

happy is not needed in the image, as far as I can test. 
If we can update the linux image with ghc-8.6.5, we are good as far gbc and gbc-tests are concerned.

Image needs to contain ghc-8.6.5 because -
   if ghc is not present stack tries to install
   And ghc's configure script gets confused when we set `CC` to something other than gcc (in our case, cc is set to clang I believe)

lalo:
Yeap, that's included in your PR already and this was to fix a separate problem. In this case the current lts defined in master produces an error when building happy with the latest stack. You can check the logs to see the difference in the errors produced.

PR #1010, opened 2019-11-16
Prepare for 8.2.0 release

chwarr:


CHANGELOG.md
```diff
@@ -36,6 +36,15 @@ different versioning scheme, following the Haskell community's
   `MemoryStream.TryGetBuffer()` when cloning streams like is done when
   targeting .NET Standard 1.3+.
 
+### Bond compiler library ###
```
chwarr: I've intentionally moved the gbc section after the language projections: when people upgrade they care more about the changes for the projection languages and less about the Haskell library.

compiler/tests/generated/alias_key_types.h
```diff
@@ -7,7 +7,7 @@
 #error This file was generated by a newer version of the Bond compiler and is incompatible with your version of the Bond library.
 #endif
 
-#if BOND_MIN_CODEGEN_VERSION > 0x0b03
```
chwarr: There haven't been any changes to the C++ library or generated code, so we can leave BOND_VERSION at 0x0800.

lalo:
Appveyor failed, running again in case of flake

lalo:
@chwarr do you want me to go thru the commits on the release to check if the version bumps are correct or don't need to CR that? let me know.

chwarr:
> @chwarr do you want me to go thru the commits on the release to check if the version bumps are correct or don't need to CR that? let me know.

I already did that. That's how I found the missing bond library bump that we needed.

PR #1011, opened 2019-11-19
Fix Cabal packaging errors

chwarr:
* Fix misspelled flag warningsAsErrors when it is used.
* Don't pass threading options when building the bond library: these
  only make sense for executables.


chwarr:
You'll want to focus on the compiler/package.yaml file: the rest of the changes are automatically generated.

lalo:
Do we need to update the release notes for this?

chwarr:
Added a CHANGELOG note.

PR #1021, opened 2020-01-15
[WiP] Support for xmldoc comments for Struct, Enum and Constant

sofusmortensen:
[WORK IN PROGRESS]

Support for xmldoc comments:

```
namespace example.some

/// foobar
enum FooBar
{
    /// foo
    /// two lines
    Foo = 1,

    /// bar
    Bar = 2,

    // not included
    Baz = 3
}

/// This is my struct
/// 2nd line
struct SomeStruct
{
    /// some field
    /// <code>someField = 42</code>
    // this is ignored
    /// last line
    0: int32 someField = 123;
    // this is ignored

    /// not ignored
    1: int32 otherField = 234;

    2: int32 notDocumented = 345;
}
```

Gets converted to C# where the xmldoc comments are moved over.

### DIFF
--- a/compiler/src/Language/Bond/Codegen/Cs/Types_cs.hs
+++ b/compiler/src/Language/Bond/Codegen/Cs/Types_cs.hs
@@ -199,13 +199,13 @@ namespace #{csNamespace}
             def x = if fieldMapping == PublicFields then Nothing else csDefault x
 
     -- C# enum definition for schema enum
-    typeDefinition e@Enum {..} = [lt|#{CS.typeAttributes cs e}public enum #{declName}
+    typeDefinition e@Enum {..} = [lt|#{CS.xmldocComments 1 declXmlDoc}#{CS.typeAttributes cs e}public enum #{declName}
     {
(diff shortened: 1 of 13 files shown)


sofusmortensen:
I have done a lot of work to stay very backwards compatible otherwise so I didn't have to edit existing unit tests testing schema json output. 

Gaps:
- All xmldoc comments are treated as "special" and currently only supported for Struct, Enum and Constant. This means that a comment elsewhere starting with 3 slashes will cause an error. This is not backwards compatible.
- Not sure whether xmldoc comments should be present in RuntimeSchema - but that would require more work.

sofusmortensen:
Ohh - and the C++ unit test is failing precisely because it is not fully backward compatible (not accepting comments in general starting with three slashes.

sofusmortensen:
@chwarr any chance you or someone else can comment on this effort?

chwarr:
Thanks for exploring this. However, this is not a feature that we're currently interested in merging into Microsoft's Bond repository.

Issue #1027, opened 2020-03-13
Bond build failed with error C5208 in MSVC


Cheney-W:
Hello, I'm a member of Microsoft VCPKG, when I built this port in an internal version of Visual Studio, it failed with following errors:
cpp\generated\bond\core\bond_reflection.h(36): error C5208: unnamed class used in typedef name cannot declare members other than non-static data members, member enumerations, or member classes
This error is caused by our implementation of [C ++ 20] Implement P1766R1, this change introduces a new warning/error when compiling code that looks like this:

```c++
typedef struct {
  int a;
  int b = 0;
  //    ^ non-static data member initializer not allowed
  void f() { } // not allowed
  static void sf() { } // not allowed
  struct U {
      // The above apply to nested classes:
      int b = 0;
      //    ^ non-static data member initializer not allowed
      void f() { } // not allowed
      static void sf() { } // not allowed
      U() { } // not allowed
      ~U() { } // not allowed
  };
} S;
```

In order to fix patterns like this the unnamed class must be given a name explicitly:

```c++
typedef struct **S_** {
  int a;
  int b = 0;
  //    ^ non-static data member initializer not allowed
  void f() { } // not allowed
  static void sf() { } // not allowed
  struct U {
      // The above apply to nested classes:
      int b = 0;
      //    ^ non-static data member initializer not allowed
      void f() { } // not allowed
      static void sf() { } // not allowed
      U() { } // not allowed
      ~U() { } // not allowed
  };
} **S**;
```

In the bond_reflection.h file, there are 30 structures like this that will trigger error C5208:

```c++
typedef struct : ::bond::reflection::FieldTemplate<
                8189,
                ::bond::reflection::optional_field_modifier,
                SerializableExceptionBase,
                std::string,
                &SerializableExceptionBase::message,
                &s_message_metadata
            > {}  message;
```

After testing, adding the name to the structure can fix this problem:
`typedef struct **message_**: ::bond::reflection::FieldTemplate<`

Since this file is an auto-generated file by the tool: Bond Compiler 0.11.0.3, and I'm not very clear about its generation rules, so could you please fix this problem? If you need any information, please contact me directly. Thanks in advance.

chwarr:
[...]

PR #1028, opened 2020-03-25
Make reflection.h types C++20 compatible.

BertusGreeff:
Fix for issue Bond build failed with error C5208 in MSVC #1027


compiler/src/Language/Bond/Codegen/Cpp/Reflection_h.hs
```diff
@@ -101,21 +101,23 @@ reflection_h export_attribute cpp file imports declarations = ("_reflection.h",
...
                 #{fieldOrdinal},
                 #{CPP.modifierTag f},
                 #{className},
                 #{cppType fieldType},
                 &#{className}::#{fieldName},
                 &s_#{fieldName}_metadata
-            > {}  #{fieldName};
+            > {};
+
+            struct #{fieldName} #{fieldName};
```
chwarr: I don't think we need this member variable.

bhardwajs:
@BertusGreeff - Are you still working on this PR? I am working on removing dependency on MSVC 2013 and upgraded my MSVC to 16.5.4 and hit these issues. I would rather merge your changes than redo those :)

BertusGreeff:
Yes, I’m still working on this.  It’s a part time job,  Was going to look at it tomorrow again.

bhardwajs:
> Yes, I’m still working on this. It’s a part time job, Was going to look at it tomorrow again.

Great. Thank you! Looking forward to the update :)

compiler/src/Language/Bond/Codegen/Cpp/Reflection_h.hs
```diff
@@ -106,16 +106,25 @@ reflection_h export_attribute cpp file imports declarations = ("_reflection.h",
...
 
-        fieldTemplates = F.foldMap $ \ f@Field {..} -> [lt|
+        -- fieldTemplateReservedNames are static variables names used in ::bond::reflection::FieldTemplate<>
+        fieldTemplateReservedNames = ["metadata", "field", "id"]
+        
+        fieldNames = map (\f -> fieldName f) structFields
+
+        fieldTemplateStructReservedNames = fieldTemplateReservedNames ++ fieldNames
+
+        uniqueFieldTemplateStructNames = uniqueNames (map (\n -> "_" ++ n) fieldNames) fieldTemplateStructReservedNames
```
ara-ayvazyan, line 116: We should also avoid using [reserved identifiers](https://en.cppreference.com/w/cpp/language/identifiers).
> the identifiers with a double underscore anywhere are reserved;
> the identifiers that begin with an underscore followed by an uppercase letter are reserved;
> the identifiers that begin with an underscore are reserved in the global namespace.

BertusGreeff: Addressed in latest revision.

PR #1029, opened 2020-04-04
Remove std::result_of_t deprecated in C++17

### LINKED ISSUE #1007: support C++17 conformance
I added /std:c++17  option in so that I could make use of std::string_view. I got this error in Bond:

c:\dev\vcpkg\installed\x86-windows\include\bond\core\detail\visit_any.h(23): error C4996: 'std::result_of_t': warning STL4014: std::result_of and std::result_of_t are deprecated in C++17. They are superseded by std::invoke_result and std::invoke_result_t. You can define _SILENCE_CXX17_RESULT_OF_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.

bhardwajs:
`try_visit_any` uses `std::result_of_t` which is deprecated in C++17. This PR replaces `std::result_of_t` by `std::invoke_result_t` so we can compile in C++ 17 mode without disabling any warnings.

Fixes #1007

### DIFF
--- a/cmake/Config.cmake
+++ b/cmake/Config.cmake
@@ -226,7 +226,3 @@ set (BOND_STACK_OPTIONS
 if (BOND_ENABLE_GRPC AND ((CXX_STANDARD LESS 11) OR (MSVC_VERSION LESS 1800)))
     message(FATAL_ERROR "BOND_ENABLE_GRPC is TRUE but compiler specified does not support C++11 standard")
 endif()
-
-if(CMAKE_CXX_STANDARD GREATER_EQUAL 17)
-    add_definitions(-DBOND_CXX_17)
-endif()
--- a/cmake/Config.cmake
+++ b/cmake/Config.cmake
@@ -227,6 +227,6 @@ if (BOND_ENABLE_GRPC AND ((CXX_STANDARD LESS 11) OR (MSVC_VERSION LESS 1800)))
     message(FATAL_ERROR "BOND_ENABLE_GRPC is TRUE but compiler specified does not support C++11 standard")
 endif()
 
-if(("${CMAKE_CXX_STANDARD}" EQUAL 17) OR ("${CMAKE_CXX_STANDARD}" EQUAL 20))
+if(CMAKE_CXX_STANDARD GREATER_EQUAL 17)
     add_definitions(-DBOND_CXX_17)
 endif()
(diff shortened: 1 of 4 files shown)


cmake/Config.cmake
```diff
@@ -226,3 +226,7 @@ set (BOND_STACK_OPTIONS
 if (BOND_ENABLE_GRPC AND ((CXX_STANDARD LESS 11) OR (MSVC_VERSION LESS 1800)))
     message(FATAL_ERROR "BOND_ENABLE_GRPC is TRUE but compiler specified does not support C++11 standard")
 endif()
+
+if(CMAKE_CXX_STANDARD AND ("${CMAKE_CXX_STANDARD}" GREATER_EQUAL 17))
```
chwarr, line 230: For compatibility, Bond builds with with CMake 3.5.1. It doesn't have `GREATER_EQUAL`, which was added in CMake 3.6. If you want to use this, you'll need to write it as a conjunction of an `EQUAL` and a `GREATER`.

bhardwajs: Ah, that explains why build was failing. Let me go ahead and do comparison with 17 or 20.

cmake/Config.cmake
```diff
@@ -226,3 +226,7 @@ set (BOND_STACK_OPTIONS
 if (BOND_ENABLE_GRPC AND ((CXX_STANDARD LESS 11) OR (MSVC_VERSION LESS 1800)))
     message(FATAL_ERROR "BOND_ENABLE_GRPC is TRUE but compiler specified does not support C++11 standard")
 endif()
+
+if(("${CMAKE_CXX_STANDARD}" EQUAL 17) OR ("${CMAKE_CXX_STANDARD}" EQUAL 20))
```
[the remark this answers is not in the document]
bhardwajs: Pushed an update for this.

cmake/Config.cmake
```diff
@@ -226,3 +226,7 @@ set (BOND_STACK_OPTIONS
 if (BOND_ENABLE_GRPC AND ((CXX_STANDARD LESS 11) OR (MSVC_VERSION LESS 1800)))
     message(FATAL_ERROR "BOND_ENABLE_GRPC is TRUE but compiler specified does not support C++11 standard")
 endif()
+
+if(CMAKE_CXX_STANDARD GREATER_EQUAL 17)
```
chwarr, line 230: Looks like the CMake change came back again.

bhardwajs: Argh!

bhardwajs: Pushed a new update to revert `Config.cmake`.

chwarr:
Can you add a note to the CHANGELOG?

bhardwajs:
> Can you add a note to the CHANGELOG with this bug fix?

I didn't initially think this was CHANGELOG worthy. :)

PR #1030, opened 2020-04-14
Remove MSVC 2013 support

bhardwajs:
To support MSVC 2013, Bond cppcodegen generated conditional code (using

This PR cleans up cppcodegen to remove MSVC 2013 conditional code and
changes unit tests in cppcodegen correspondingly.

It also cleans up C++ and python code of MSVC 2013 conditional code.

### DIFF
--- a/compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
+++ b/compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
@@ -308,11 +308,7 @@ types_h export_attribute userHeaders enumHeader allocator alloc_ctors_enabled ty
             -- even if implicit would be okay, fall back to explicit for
             -- compilers that don't support = default for move constructors
                                     else [lt|
-#if defined(_MSC_VER) && (_MSC_VER < 1900)  // Versions of MSVC prior to 1900 do not support = default for move ctors
-        #{explicit}
-#else
-        #{implicit}
-#endif|]
+        #{implicit}|]
           where
             -- default OK when there are no meta fields
             implicit = [lt|#{declName}(#{declName}&&) = default;|]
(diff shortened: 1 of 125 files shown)


bhardwajs:
@chwarr This PR is ready for review. My take is we can have the migration from appveyor to ADO pipeline as a follow-up PR.

FYI @BertusGreeff

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
```diff
@@ -308,11 +308,7 @@ types_h export_attribute userHeaders enumHeader allocator alloc_ctors_enabled ty
             -- even if implicit would be okay, fall back to explicit for
```
ara-ayvazyan, line 308: The [`struct _bond_vc12_ctor_workaround_`](https://github.com/microsoft/bond/blob/65017969ce65322a60d8314b0bf65c5e99ed8ee9/compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs#L219) above can also be removed ([cannot comment on collapsed lines](https://github.com/isaacs/github/issues/1655) 😞).

bhardwajs: @ara-ayvazyan - Yeah, I was going to do another round of cleanup after this one. Would you prefer to have it done in this PR?

ara-ayvazyan: I thought that part seem to be related to this PR.

ara-ayvazyan:
How about separating this into several PRs?
1. Remove the MSVC 2013 from build matrix, update doc., etc.
2. Fix the code generation.
3. Fix core library code.

bhardwajs:
> How about separating this into several PRs?
>     1. Remove the MSVC 2013 from build matrix, update doc., etc.
>     2. Fix the code generation. 
>     3. Fix core library code.

Sounds good. Let me send the first PR to remove MSVC 2013 from build matrix and update doc. I will remove `_bond_vc12_ctor_workaround_` in second PR. Third one should be unchanged.

bhardwajs:
@ara-ayvazyan Just pushed an update so that the first PR only has build matrix and docs changes.

PR #1032, opened 2020-04-23
Clean up compiler of MSVC 2013 conditional code

bhardwajs:
Clean up MSVC 2013 conditional code in C++ codegen. Also change the gbc
tests as per the codegen. To generate the tests:
> cd compiler
>  stack test :gbc-tests --test-arguments --accept

Note that this PR can only be merged after #1030 is merged.

### DIFF
--- a/compiler/src/Language/Bond/Codegen/Cpp/Enum_h.hs
+++ b/compiler/src/Language/Bond/Codegen/Cpp/Enum_h.hs
@@ -20,7 +20,7 @@ enum_h :: MappingContext -> String -> [Import] -> [Declaration] -> (String, Text
 enum_h cpp _file _imports declarations = ("_enum.h", [lt|
 #pragma once
 
-#include <stdint.h>
+#include <cstdint>
 
 #{CPP.openNamespace cpp}
 namespace _bond_enumerators
--- a/compiler/src/Language/Bond/Codegen/Cpp/Enum_h.hs
+++ b/compiler/src/Language/Bond/Codegen/Cpp/Enum_h.hs
@@ -20,7 +20,7 @@ enum_h :: MappingContext -> String -> [Import] -> [Declaration] -> (String, Text
 enum_h cpp _file _imports declarations = ("_enum.h", [lt|
 #pragma once
 
-#include <cstdint>
+#include <stdint.h>
 
 #{CPP.openNamespace cpp}
 namespace _bond_enumerators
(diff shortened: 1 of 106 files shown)


chwarr:
I merged your PR that removed MSVC 2013 from the build matrix. Can you merge with the current tip of master so that we can get a clean build?

compiler/src/Language/Bond/Codegen/Cpp/Types_h.hs
```diff
@@ -308,11 +305,7 @@ types_h export_attribute userHeaders enumHeader allocator alloc_ctors_enabled ty
             -- even if implicit would be okay, fall back to explicit for
             -- compilers that don't support = default for move constructors
```
chwarr, line 306: Minor: comment looks stale

compiler/src/Language/Bond/Codegen/Cpp/Enum_h.hs
```diff
@@ -20,7 +20,7 @@ enum_h :: MappingContext -> String -> [Import] -> [Declaration] -> (String, Text
 enum_h cpp _file _imports declarations = ("_enum.h", [lt|
 #pragma once
 
-#include <stdint.h>
+#include <cstdint>
```
chwarr, line 23: While `<cstdint>` may be more C++ish, we use the unnamespaced `<stdint.h>` type names everywhere else and consuming code may be doing so as well. This doesn't seem like we get enough benefit to make that breaking change.

compiler/src/Language/Bond/Codegen/Cpp/Util.hs
```diff
@@ -159,8 +159,8 @@ enumDefinition Enum {..} = [lt|enum #{declName}
         };|]
   where
     constant Constant {..} = [lt|#{constantName}#{optional value constantValue}|]
-    value (-2147483648) = [lt| = static_cast<int32_t>(-2147483647-1)|]
-    value x = [lt| = static_cast<int32_t>(#{x})|]
+    value (-2147483648) = [lt| = static_cast<std::int32_t>(-2147483647-1)|]
```
chwarr, line 162: See other comment: I think we want to revert this change.

compiler/src/Language/Bond/Codegen/Cpp/Enum_h.hs
bhardwajs: I looked through the current versions of `<cstdint>` in MSVC, gcc, and clang and all three would be fine with this code.

However, you are write, the consuming code might not be on the most recent toolchain. I will revert.

PR #1033, opened 2020-04-23
Removing MSVC 2013 conditional code from C++ code

bhardwajs:
In PR #1030, we remove the support for MSVC 2013 from Bond CI pipeline. PR #1032 cleans up compiler of MSVC 2013 conditional code. This PR removes MSVC 2013 conditional code from C++ and simplifies the resulting C++ code.

This PR can only be merged after #1030 is merged.


cpp/test/core/maybe_tests.cpp
```diff
@@ -377,7 +357,6 @@ TEST_CASE_BEGIN(AllocatorPropagated)
 }
 TEST_CASE_END
 
-#if !defined(_MSC_VER) || _MSC_VER >= 1900
 // MSVC 2013 doesn't have implementations of these type traits, so we skip
```
chwarr, line 360: Minor: this comment looks stale

examples/cpp/core/scoped_allocator/allocator.h
```diff
@@ -12,12 +12,6 @@ namespace scoped_allocator
     {
         using value_type = T;
 
-#if defined(_MSC_VER) && _MSC_VER < 1910
-        // MSVC 2015 and older versions have bug in std::scoped_allocator_adaptor
```
chwarr, line 16: We're still targeting MSVC 2015, so I think this need to stay.

bhardwajs: Yeah, good catch. Thanks!

bhardwajs:
@chwarr - Thanks for the review. I have rebased to maser so build should succeed.
I have removed `inc/bond/core/warning.h` and associated code block in config.cmake.
Also took care of the two minor comments.

PR #1036, opened 2020-05-11
Use non-deprecated bind/bind.hpp header

vaboca:
boost/bind.hpp is deprecated and recommends using boost/bind/bind.hpp instead, see: https://github.com/boostorg/bind/blob/develop/include/boost/bind.hpp

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ different versioning scheme, following the Haskell community's
   declare members other than non-static data members, member enumerations,
   or member classes. ([Issue
   #1027](https://github.com/microsoft/bond/issues/1027))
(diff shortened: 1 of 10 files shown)


chwarr:
It looks like the AppVeyor CI build is broken due to a change in the image names that AppVeyor uses. It doesn't look like this code is the cause of the break.

However, I cannot merge this change until that is fixed.
I've opened [issue #1038](https://github.com/microsoft/bond/issues/1038) about this.

chwarr:
We got a fix for #1038. I've merged that into this PR's branch.

chwarr:
I've added the changelog line. This is ready for merging after the CI build finishes.

PR #1039, opened 2020-05-16
Update AppVeyor to know about VS 2019

bhardwajs:
Update AppVeyor to know about VS 2019.

Fixes #1038

PR #1041, opened 2020-05-16
Make _grpc.h types C++20 compatible

chwarr:
As of C++20, it is malformed to have an unnamed class used in typedef with a
base class. In the generated reflection code for a service, we were using the
following construct to create a reflectable entry for each field in a
struct.

```
struct service {
    typedef struct /* no name */ : ::bond::reflection::MethodTemplate<...> {} method_name;
}
```

Now, instead of using an unnamed class, we generate a named class and refer
to that:

```
struct service {
    typedef struct method_name_type : ::bond::reflection::MethodTemplate<...> {} field_name;
}
```

This commit is very similar in spirit to f1cb707134. See that commit for
even more details about what's going on.

Also adds a test that compiles a service that has method names that
conflict with the names in `MethodTemplate`. The gRPC++ unit test CMake
now has a single library with the compiled generated code.

Fixes https://github.com/microsoft/bond/issues/1027


chwarr:
@BertusGreeff, can you help review this? It's very similar to your PR #1028.

chwarr:
This will probably need to be merged with https://github.com/microsoft/bond/pull/1039 to fix any AppVeyor build failures in the C# code.

PR #1045, opened 2020-05-23
Fix documents' links to .md

### LINKED ISSUE #1044: Documents contain incorrect links to .html instead of .md
For example, doc/src/bond_cpp.md has `[syntax]` links to `compiler.html#idl-syntax` instead of `compiler.md#idl-syntax`.

chpong:
Fix #1044 - Replace the links in the doc/src/*.md with links to .md file.

### DIFF
--- a/doc/src/bond_cpp.md
+++ b/doc/src/bond_cpp.md
@@ -22,7 +22,7 @@ Bond is published on GitHub at
 Basic example
 =============
 
-In Bond data schemas are defined using idl-like [syntax](compiler.html#idl-syntax):
+In Bond data schemas are defined using idl-like [syntax](compiler.md#idl-syntax):
 
 ```
 namespace example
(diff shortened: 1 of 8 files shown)


chwarr:
Thanks for the PR, but I think this will break the compiled documentation output. From a [comment of mine in issue #1044](https://github.com/microsoft/bond/issues/1044#issuecomment-634198835):

> This links to .html files are expected. Otherwise, the [compiled/published documentation](https://microsoft.github.io/bond/) wouldn't work correctly, as Pandoc doesn't know to convert, say, a link to compiler.md to compiler.html during the [documentation build](https://github.com/microsoft/bond/blob/2e7aa1ef70121bc5ec221286d9f0ee3295e1d3b2/doc/CMakeLists.txt).
>
> If you know a way to have Pandoc do this conversion, we'd be happy to integrate that so that the links work both from the documentation source code (the .md files) and the published output (the .html files).

chpong:
Thanks @chwarr, closing the PR.

PR #1046, opened 2020-05-26
Upgrade rapidjson to Apr 10, 2020 commit

### LINKED ISSUE #1042: Build steps failing due to -Werror=class-memaccess
I'm trying to build this repo in a Docker container following the steps in the readme.  Here is an excerpt from my Dockerfile:

```
FROM mcr.microsoft.com/dotnet/core/sdk:3.1

ARG tools_dependencies="gcc cmake libboost-dev libboost-thread-dev unzip zlib1g-dev"
RUN DEBIAN_FRONTEND=noninteractive apt-get update && apt-get install -y --no-install-recommends \
    ${tools_dependencies}

RUN curl -sSL https://get.haskellstack.org/ | sh

RUN cd /tmp && \
    git clone --recursive https://github.com/microsoft/bond.git && \
    cd bond && \
    mkdir build && \
    cd build && \
    cmake -DBOND_ENABLE_GRPC=FALSE .. && \
    make -j $(nproc) && \
    make install && \
    cd ../.. && \
    rm -rm /tmp/bond
```
The make step fails with:

```
/tmp/bond/thirdparty/rapidjson/include/rapidjson/document.h:1939:24: error: 'void* memcpy(void*, const void*, size_t)' writing to an object of type 'class rapidjson::GenericValue<rapidjson::UTF8<> >' with no trivial copy-assignment; use copy-assignment or copy-initialization instead [-Werror=class-memaccess]
             std::memcpy(e, values, count * sizeof(GenericValue));

```

bhardwajs:
Upgrade rapidjson to commit [8f4c021 ](https://github.com/Tencent/rapidjson/commit/8f4c021fa2f1e001d2376095928fc0532adf2ae6)from Apr 10.

Fixes #1042


chwarr:
Bond's [.gitmodules file](https://github.com/microsoft/bond/blob/2e7aa1ef70121bc5ec221286d9f0ee3295e1d3b2/.gitmodules) still has the miloyip address for RapidJSON instead of the Tencent one. Perhaps that's why the commit cannot be found?

Issue #279, opened 2016-11-20
Different namespace for TypeAlias converters


yagilofir:
Hi.
I am working with C#
From reading the code of this class:
https://github.com/Microsoft/bond/blob/9b8aa2fd15d72383ea86d6017c4e887c1efb8807/cs/src/core/expressions/TypeAlias.cs

I understand that in order to create a type-alias converter, my converter namespace must be the same as the namespace of the .bond scheme.
I think that this limitation is too harsh.
I want to create my own converters library with the namespace Core/Bond/Converters
Now, my DateTime converter (for example) will be Core/Bond/Converters/DateTimeConverter.cs
The MyStruct.Bond namespace does not have any relation to the namespace "Core/Bond/Converters/DateTimeConverter" (actually, it's namespace is "MyProgram/Models", But I do want it to use the converters from my core library.
Is it possible to tell the gbc.exe to use a different namespace for the type converters (other than the MyStruct.Bond namespace)?

chwarr:
There's nothing in gbc or the C# libraries that will permit this today. Your current recourse--if you want to keep them in separate namespaces--is to have converter implementations in `MyProgam.Models` that delegate to the ones in `Core.Bond.Converters`, which can be tedious. Static classes cannot inherit from one another, so there's lots of boilerplate typing. And, you can't just move the `BondTypeAliasConverter` type into the namespace `System`, to try to get a match for `System.DateTime` as the converter must be found in the same assembly as the type.

To enable something like this, we'd need to come up with a way to annotate the C# field/property/type with the location of the converters that is used in addition/instead of the current rules. Then, we'd need to come up with a way to tell gbc to emit such an annotation. Nothing too hard, but this one should start from a design proposal, not just a code change.

chwarr:
Having heard nothing else, closing this issue.

eharshosh:
> To enable something like this, we'd need to come up with a way to annotate the C# field/property/type with the location of the converters that is used in addition/instead of the current rules. Then, we'd need to come up with a way to tell gbc to emit such an annotation. Nothing too hard, but this one should start from a design proposal, not just a code change.

I want to suggest a different approach to tackle the problem:
Allow specifying conversion method in a 'dependency injection' style - so that the writers/readers will look up for custom converters when it needs one.

PR #1054, opened 2020-06-03
[c#] Support defining AliasConveters in any assembly

eitanhs:
This will allow the user to create a `BondTypeAliasConverter`s for types like System.TimeSpan once,
and use in any generated code (independent of the generated type namespace).

In my case, I want to distribute a library with common converters, which is currently not doable in C# (I can do this in C++)

The logic is simple: try to locate converter class in the same assembly of the target type, if not found, try to locate in any assembly.

related to issue #279

### DIFF
--- a/cs/src/core/expressions/TypeAlias.cs
+++ b/cs/src/core/expressions/TypeAlias.cs
@@ -111,8 +111,11 @@ namespace Bond.Expressions
         static Type GetConverter(Type type)
         {
             var name = type.AssemblyQualifiedName;
-            var converterName = type.Namespace + ".BondTypeAliasConverter" + name.Substring(type.FullName.Length);
-            return Type.GetType(converterName);
+            var converterName = type.Namespace + ".BondTypeAliasConverter";
+            // First try to locate converter class in the same assembly of the target type,
+            // If not found, try to locate in any assembly:
+            return Type.GetType(converterName + name.Substring(type.FullName.Length)) 
+                   ?? Type.GetType(converterName);
         }
     }
 }


cs/src/core/expressions/TypeAlias.cs
```diff
@@ -111,8 +111,11 @@ static void GetSchemaTypes(Type type, HashSet<Type> schemaTypes)
         static Type GetConverter(Type type)
         {
             var name = type.AssemblyQualifiedName;
-            var converterName = type.Namespace + ".BondTypeAliasConverter" + name.Substring(type.FullName.Length);
-            return Type.GetType(converterName);
+            var converterName = type.Namespace + ".BondTypeAliasConverter";
+            // First try to locate converter class in the same assembly of the target type,
+            // If not found, try to locate in any assembly:
+            return Type.GetType(converterName + name.Substring(type.FullName.Length)) 
+                   ?? Type.GetType(converterName);
```
chwarr, line 118: At least looking at the documentation for [`Type.GetType(string)`](https://docs.microsoft.com/en-us/dotnet/api/system.type.gettype?view=netcore-3.1#System_Type_GetType_System_String_), a name without assembly qualification will only look in the currently executing assembly/mscorlib.dll/System.Private.CoreLib.dll.

Is something more subtle going on here that I'm not seeing?

cs/src/core/expressions/TypeAlias.cs
```diff
@@ -111,8 +111,11 @@ static void GetSchemaTypes(Type type, HashSet<Type> schemaTypes)
         static Type GetConverter(Type type)
         {
             var name = type.AssemblyQualifiedName;
-            var converterName = type.Namespace + ".BondTypeAliasConverter" + name.Substring(type.FullName.Length);
-            return Type.GetType(converterName);
+            var converterName = type.Namespace + ".BondTypeAliasConverter";
```
chwarr, line 114: Can you add a unit test case for this?

cs/src/core/expressions/TypeAlias.cs
chwarr, line 118: What happens if multiple types have the same name in different assemblies?

eitanhs:
Thanks for your review.
As you pointed out, it might not work.
I'm sorry I didn't test my assumptions first...
I suppose I'll figure some other mechanism that will also adress your other concern regarding conflicting/racing definitions from different assemblies.
I'm closing for now and will open another PR  (with proper UTs this time) when I have time to rethink and implement.

Thanks again!

chwarr:
Sounds good & looking forward to it.

PR #1058, opened 2020-07-15
[c#] Add new style license and icon to NuGet pkgs.

chwarr:
The nuspec elements `iconUrl` and `licenseUrl` are deprecated in favor of
the `icon` and `license` elements.

This commit replaces `licenseUrl` with `license` (The two cannot
co-exist.) and adds `icon`. (The current recommendation is to have both
for compatibility.)

The `icon` element requires that the icon file be in the package, so the
packages all now use the repository root as the base for their files and
include the icon .png.


bhardwajs:
Does this need an entry in Changelog?

chwarr:
I don't think this warrants a changelog entry. This is addressing warnings from NuGet.org during package publish. There shouldn't be any user visible changes.

PR #1060, opened 2020-07-15
[c++ grpc] Pass --init to update gRPC submodules

chwarr:
Sometimes gRPC adds new submodule dependencies, so we need to pass the
`--init` flag to `git submodule update` so that new submodules are
initialized, lest the build fail.


chwarr:
My [test build](https://travis-ci.org/github/microsoft/bond/builds/708483384) of this change passed. The CI build that runs as part of the PR won't actually build the grpc-master config.

PR #1063, opened 2020-07-31
[c#] Move Position Computation Into PushLengthCheck Args

mikemsft:
This pull request attempts to fix a regression introduced in [b0fd4a1](https://github.com/microsoft/bond/commit/b0fd4a15a7cae946dd2855122559ca59cc34dbea#diff-9534daaa1fb3d4776494b25c8bba3939L212). 

The previous code for `CompactBinary.WriteStructBegin` would not access the `output.Position` in the Release build configuration whereas the changes introduced caused it to now be accessed regardless of configuration. This can cause exceptions in scenarios where the `Stream` is not seekable.

The fix moves the computation of the length back into the arguments of the function which ensures it will not be executed outside of the Debug build configuration.

### DIFF
--- a/cs/test/core/StreamTests.cs
+++ b/cs/test/core/StreamTests.cs
@@ -220,8 +220,9 @@
             return buf;
         }
 
-        // WriteStructBegin uses position in DEBUG configuration which would fail this test
-        // Test needs to validate position is not accessed in RELEASE so conditional is set
+        // WriteStructBegin should only access Stream.Position in the DEBUG
+        // configuration, so we only execute this test in the RELEASE
+        // configuration.
         [Conditional("RELEASE")]
         private void Stream_PositionLength_NotAccessedOnWriteStructBeginImplementation()
         {
--- a/cs/test/core/StreamTests.cs
+++ b/cs/test/core/StreamTests.cs
@@ -123,7 +123,7 @@
         [Test]
         public void Stream_PositionLength_NotAccessedOnWriteStructBegin()
         {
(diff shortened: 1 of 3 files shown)


cs/test/core/StreamTests.cs
```diff
@@ -119,6 +120,12 @@ public void Stream_PositionLength_AsExpected()
             Assert.IsTrue(input.Position == pos);
         }
 
+        [Test]
+        public void Stream_PositionLength_NotAccessedOnWriteStructBegin()
+        {
+            this.Stream_PositionLength_NotAccessedOnWriteStructBeginImplementation();
```
chwarr, line 126: Minor: no `this.` for method invocation.

cs/test/core/StreamTests.cs
```diff
@@ -212,6 +219,54 @@ protected static byte[] MakeSequential(int length)
 
             return buf;
         }
+
+        [Conditional("RELEASE")]
```
chwarr, line 223: Minor: can you add a comment about why this is `Conditional`.

cs/test/core/StreamTests.cs
```diff
@@ -212,6 +219,54 @@ protected static byte[] MakeSequential(int length)
...
+        private void Stream_PositionLength_NotAccessedOnWriteStructBeginImplementation()
+        {
+            var stream = new NonSeekableStream();
+            var output = new OutputStream(stream, 11);
```
chwarr, line 227: Minor: can you used named parameters so that the 11 and 2 have names associated with them?

Issue #1062, opened 2020-07-26
brew install bond


ankane:
Hi, thanks for open sourcing this project! I'm hoping to add bond to Homebrew so Mac users can install it with:

```sh
brew install bond
```

However, I'm running into an error when trying to test it.

```txt
Undefined symbols for architecture x86_64:
  "examples::serialization::Struct::Schema::s_n_metadata", referenced from:
      boost::disable_if_c<(is_reader<examples::serialization::Struct const&>::value) && (is_fast_path_field<examples::serialization::Struct::Schema::var::n_type, bond::Serializer<bond::CompactBinaryWriter<bond::OutputMemoryStream<std::__1::allocator<char> > >, bond::BuiltInProtocols> >::value), bool>::type bond::detail::NonBasicTypeField<examples::serialization::Struct::Schema::var::n_type, bond::Serializer<bond::CompactBinaryWriter<bond::OutputMemoryStream<std::__1::allocator<char> > >, bond::BuiltInProtocols>, examples::serialization::Struct const&>(examples::serialization::Struct::Schema::var::n_type const&, bond::Serializer<bond::CompactBinaryWriter<bond::OutputMemoryStream<std::__1::allocator<char> > >, bond::BuiltInProtocols> const&, examples::serialization::Struct const&) in serialization-74f814.o
      boost::enable_if_c<(!(is_bond_type<examples::serialization::Struct::Schema::var::n_type::field_type>::value)) && (std::is_same<examples::serialization::Struct::Schema::var::n_type::field_modifier, std::__1::integral_constant<bond::_bond_enumerators::Modifier::Modifier, (bond::_bond_enumerators::Modifier::Modifier)0> >::value), void>::type bond::detail::OptionalDefault<examples::serialization::Struct>::operator()<examples::serialization::Struct::Schema::var::n_type>(examples::serialization::Struct::Schema::var::n_type const&) in serialization-74f814.o
[...]

PR #1064, opened 2020-08-01
[c++] Install gbc with 555 permissions

chwarr:
gbc was being installed with 111 permissions (--x--x--x) instead of
555 (r-xr-xr-x).

Hat tip to Andrew Kane <andrew@chartkick.com> for reporting this.


chwarr:
FYI: @ankane, I was able to repro the permissions issue you mentioned in #1062. Here's my fix.

ankane:
Thanks @chwarr!

PR #1066, opened 2020-08-06
[c#] fix Serious performance regression in Grow method of OutputBuffer 

### LINKED ISSUE #1065: Serious performance regression in Grow method of OutputBuffer
Recently, in order to mitigate CVE-2020-1469, among other changes, [line 224 of OutputBuffer class](
https://github.com/microsoft/bond/blame/d819f3f1f80c2f1975b9da51137628e83ba0b81a/cs/src/core/io/safe/OutputBuffer.cs#L224) 
changed from `length += length >> 1;` to `length = checked(length + length >> 1);`
This actually doesn't change the value in `length` variable, [because the `+` operator precedes the bit-shift operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/#operator-precedence),
So basically we multiply and divide == no change.
This causes **serious** performance regression as the underlying buffer grows by single value at a time (instead of growing by half of current size)...

I'll open a PR with a fix suggestion soon.

eitanhs:
fixes #1065

### DIFF
--- a/cs/src/core/io/safe/OutputBuffer.cs
+++ b/cs/src/core/io/safe/OutputBuffer.cs
@@ -221,7 +221,7 @@ namespace Bond.IO.Safe
         internal virtual void Grow(int count)
         {
             int minLength = checked(position + count);
-            length = checked(length + length >> 1);
+            length = checked(length + (length >> 1));
 
             const int ArrayIndexMaxValue = 0x7FFFFFC7;
             if ((uint)length > ArrayIndexMaxValue)


chwarr:
I merged this manually to update the changelog and expand the commit message at the same time. Commit https://github.com/microsoft/bond/commit/3afea822c42dd0095fedb9e7db9ebb99165e7343 has been merged into master.

eitanhs:
> I merged this manually to update the changelog and expand the commit message at the same time. Commit https://github.com/microsoft/bond/commit/3afea822c42dd0095fedb9e7db9ebb99165e7343 has been merged into master.

Thank you!

lalo:
@eitanhs do you have some sort of regression tests in place? or how did you catch this so quickly?

eitanhs:
> @eitanhs do you have some sort of regression tests in place? or how did you catch this so quickly?

Sorry to disapoint, but no..  just updated to the latest version today and noticed the performance drop after manually running a tool I wrote that serializes large objects.
The issue was easy to detect since each time I stopped the debugger I noticed that it soppted at Array.Resize call, then I used github magic to compare changes between the previous release to the current and voilà :)

lalo:
Great work though

chwarr:
Release 9.0.3 is live now and has this fix.

Issue #1072, opened 2020-09-22
AppVeyor C# CI builds failing: not building projects in the right order?


chwarr:
The C# PR builds for a C++-only change, #1071, look like they're failing because the projects aren't build built in the right order.

From [one of the builds](https://ci.appveyor.com/project/MicrosoftBond/bond/builds/35288709/job/1sjqqpli22bgn8yf), it looks like it's trying to build examples\cs\core\enumerations\enumerations.csproj before bond.csproj.

```
 Building Bond Compiler...
  -- Building for: Visual Studio 16 2019
  Attributes -> C:\projects\bond\cs\src\attributes\bin\debug\net45\Bond.Attributes.dll
  Attributes -> C:\projects\bond\cs\src\attributes\bin\debug\netstandard1.0\Bond.Attributes.dll
  Attributes -> C:\projects\bond\cs\src\attributes\bin\debug\netstandard1.6\Bond.Attributes.dll
  reflection -> C:\projects\bond\cs\src\reflection\bin\debug\net45\Bond.Reflection.dll
  reflection -> C:\projects\bond\cs\src\reflection\bin\debug\netstandard1.0\Bond.Reflection.dll
  reflection -> C:\projects\bond\cs\src\reflection\bin\debug\netstandard1.6\Bond.Reflection.dll
  -- The C compiler identification is MSVC 19.27.29111.0
  -- The CXX compiler identification is MSVC 19.27.29111.0
  -- Detecting C compiler ABI info
  -- Detecting C compiler ABI info - done
  -- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.27.29110/bin/Hostx64/x64/cl.exe - skipped
  -- Detecting C compile features
  -- Detecting C compile features - done
  -- Detecting CXX compiler ABI info
  -- Detecting CXX compiler ABI info - done
  -- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.27.29110/bin/Hostx64/x64/cl.exe - skipped
  -- Detecting CXX compile features
  -- Detecting CXX compile features - done
  -- Stack found at C:/ProgramData/chocolatey/bin/stack.exe. Version 2.3.3, Git revision cb44d51bed48b723a5deb08c3348c0b3ccfc437e x86_64 hpack-0.33.0
  -- Configuring done
  -- Generating done
[...]

PR #1073, opened 2020-09-23
Explicit dependency for enumerations.csproj

lalo:


lalo:
Took a stab at #1072 but couldn't make it work. Must be misinterpreting the error message, also I wonder how it regressed?

chwarr:
Thanks for trying. I suspect a newer version of the .NET SDK was installed that resolves our under-specified dependencies differently. I _think_ the fix is to use `<ProjectReference>` everywhere instead of `<HintPath>$(BOND_BINARY_PATH)\net45\Bond.Attributes.dll</HintPath>`.

This isn't something I'll have time to do in the next few months at least...

chwarr:
I looked at the error messages for your build. I think enumerations.csproj is fixed now. It looks like there are other projects that are missing dependencies too. I still think that `<ProjectReference>` is the robust way forward, but SLN changes to get this building again in the meantime are fine by me.

lalo:
Yea, that makes sense. This might be simple with a vim macro.

PR #1085, opened 2021-01-12
[ci] enable github actions for linux ci

lalo:
- [x] cs
- [x] cpp-core
- [x] cpp-grpc
- [x] hs
- [x] java
- [x] optional: update image to have latest git
- [ ] optional: refactor 'sed's to proper scripting
- [x] optional: refactor java 'sed' to script (easy)
- [x] optional: cache (only for stack, clears every 7 week so semi-useful)


lalo:
https://github.blog/changelog/2020-09-24-github-actions-private-registry-support-for-job-and-service-containers/

.github/workflows/linux.yml
```diff
@@ -0,0 +1,84 @@
...
+  pull_request:
+    branches: ["master"]
+
+jobs:
+  tests:
+    name: "${{ matrix.env.FLAVOR }} ${{ matrix.env.BOOST }} ${{ matrix.env.COMPILER }}"
+    runs-on: "ubuntu-latest"
+
+    container:
+      image: bondciimages.azurecr.io/ubuntu-1604:build-32869346
```
chwarr, line 16: Did you try specifying the volumes here to get the source code under /root/bond like the build scripts are expecting? This may mean that the `sed` on the paths won't be needed.

```suggestion
      image: bondciimages.azurecr.io/ubuntu-1604:build-32869346
      volumes:
        - /github/workspace:/root/bond
```

(It looks like the source code gets cloned into `/github/workspace`.)

.github/workflows/linux.yml
```diff
@@ -0,0 +1,84 @@
...
+      fail-fast: false
+      matrix:
+        env:
+          - { FLAVOR: cs }
+          - { FLAVOR: cpp-core, BOOST: 1.66.0, COMPILER: clang }
+          - { FLAVOR: cpp-grpc, BOOST: 1.66.0, COMPILER: clang }
+          - { FLAVOR: hs }
+          - { FLAVOR: java }
+
+          - { FLAVOR: cpp-core, BOOST: 1.65.1, COMPILER: clang, CRON_ONLY: true }
```
chwarr, line 28: I think the way to do the cron is to have two top-level workflows:

1. A workflow with the current `on` that has combinations without `CRON_ONLY: true`.
2. A workflow with an `on.schedule` that has the combinations with `CRON_ONLY: true`.

`CRON_ONLY` could go away from these workflows. The ZSH scripts never refer to it. It is just used by the Travis CI build script to talk to itself.

chwarr (changes requested):

chwarr:
Thanks for getting this started! I left some review comments with questions about volume mounts. I _think_, just from reading the docs, that the `sed` step can be removed.

.github/workflows/linux.yml
lalo: it didn't work, not sure why. I also tried - /home/runner/work/bond:/root since in the initialize container step it references /home/runner/work/ as the volume that later becomes the workspace

.github/workflows/linux.yml
lalo: done

.github/workflows/linux.yml
chwarr: OK. Thanks for experimenting.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,84 @@
...
+      fail-fast: false
+      matrix:
+        env:
+          - { FLAVOR: cs }
+          - { FLAVOR: cpp-core, BOOST: 1.66.0, COMPILER: clang }
+          - { FLAVOR: cpp-grpc, BOOST: 1.66.0, COMPILER: clang }
+          - { FLAVOR: hs }
+          - { FLAVOR: java }
+
+          - { FLAVOR: cpp-core, BOOST: 1.65.1, COMPILER: clang }
```
chwarr, line 28: The variants covered by the cron workflow can be removed from the CI workflow.

chwarr (changes requested):

.github/workflows/linux.yml
lalo: done

chwarr (approved):

chwarr:
Let me know when this is ready and I can merge it. Then you can revise the build.zsh script paths and we can drop the Git install step by using a newer image.

lalo:
@chwarr feel free to merge

PR #1090, opened 2021-01-14
[ci] update image for cron run

lalo:
forgot this one @chwarr

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,7 +26,7 @@
         - CHANGELOG.md
         # These files are only used by the Linux CI builds and don't affect
         # the Windows builds.
-        - .github/**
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -26,7 +26,7 @@
         - CHANGELOG.md
         # These files are only used by the Linux CI builds and don't affect
         # the Windows builds.
-        - .github/**
+        - .github/**/*
```
chwarr, line 29: Confirming that the extra `/*` at the end is what the [filtering rules](https://www.appveyor.com/docs/how-to/filtering-commits/#file-matching-rules) say should be used to match all files in all directories recursively.

chwarr (approved):

PR #1093, opened 2021-01-21
[c++] Clarify some details of the Compact Binary encoding

jviotti:
In particular:

- ID bits consists of big endian unsigned integers
- int8 uses two's complement
- Just to make it super clear, "variable encoding" of integers refers to
  LEB128
- Real numbers use IEEE 764
- Enum constants are int32

I had to run various experiments to find these things out and they might
be helpful for the next reader.

Signed-off-by: Juan Cruz Viotti <jv@jviotti.com>

### DIFF
--- a/cpp/inc/bond/protocol/compact_binary.h
+++ b/cpp/inc/bond/protocol/compact_binary.h
@@ -40,7 +40,7 @@
    field             | id+type  |  value   |
                      '----------'----------'
 
-                                            .---.---.---.---.---.---.---.---.                       i - id bits
+                                            .---.---.---.---.---.---.---.---.                       i - id bits (BE unsigned int)
    id+type           0 <= id <= 5           | i | i | i | t | t | t | t | t |                       t - type bits
                                             '---'---'---'---'---'---'---'---'                       v - value bits
                                               2       0   4               0
@@ -66,12 +66,14 @@
                                             '---'---'---'---'---'---'---'---'
                                               7                           0
 
+                                            signed uses two's complement
+
                                             .---.---.   .---.---.---.   .---.
                      uint16, uint32,        | 1 | v |...| v | 0 | v |...| v |  [...]
                      uint64                 '---'---'   '---'---'---'   '---'
                                                   6       0       13      7
 
-                                            variable encoding, high bit of every byte
+                                            LEB128 variable encoding, high bit of every byte
                                             indicates if there is another byte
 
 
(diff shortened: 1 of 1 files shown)


cpp/inc/bond/protocol/compact_binary.h
```diff
@@ -61,17 +61,21 @@
                                             '---'---'---'---'---'---'---'---'
                                                                           0
 
+                                            falsy booleans fields are not encoded
```
chwarr, line 64: This isn't the case: false boolean fields will be encoded if they are `required`, `required_optional`, or `= nothing`.

jviotti, line 64: @chwarr I have an e-mail notification about a comment on this statement, but it doesn't appear on GitHub anymore. Was the comment deleted?

cpp/inc/bond/protocol/compact_binary.h
```diff
@@ -86,8 +90,10 @@
                                             and then encoded as unsigned integer
 
 
-                     float, double          little endian
+                     float, double          little endian IEEE 764
```
chwarr, line 93: The size can be included too. How about "32- or 64-bit little endian IEEE 764"?

chwarr (changes requested):

cpp/inc/bond/protocol/compact_binary.h
chwarr: I moved it into a review. Should be visible now, under that review.

chwarr:
Thanks! A few tweaks and this will be good to merge.

cpp/inc/bond/protocol/compact_binary.h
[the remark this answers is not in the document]
jviotti: Ah, good to know! Done!

cpp/inc/bond/protocol/compact_binary.h
jviotti: Sounds good!

cpp/inc/bond/protocol/compact_binary.h
```diff
@@ -98,7 +104,8 @@
                                             units
 
                            characters       1-byte UTF-8 code units (for string) or 2-byte
-                                            UTF-16LE code units (for wstring)
+                                            UTF-16LE code units (for wstring) without a NUL
```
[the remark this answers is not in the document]
jviotti: I see. I'll remove the change! However:

> but Bond strings can contain interior and terminating NULs, if that is the string the application sends

What is the point of appending a NUL terminator in that case then?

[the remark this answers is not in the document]
jviotti: Scratch that. I got confused while looking at another format. Bond does not append a NUL delimiter and that is what I was trying to clarify. Nevermind!

jviotti:
@chwarr All requested changes were addressed!

Issue #1086, opened 2021-01-13
Doc builds are not pushing to GitHub pages


chwarr:
A recent build of commit a3002778e ([cs] Mention inlineNested in perf guidance, 2021-01-12) had documentation changes. [The "Docs" AppVeyor build](https://ci.appveyor.com/project/MicrosoftBond/bond/builds/37232906/job/r92grcunv62i76sm) was unable to commit and push these to the gh-pages branch.

It looks like something is going wrong copying the generated files into the repo cloned into the gh-pages directory.

```lang-powershell
Remove-Item * -Recurse

Copy-Item ..\html\* . -Recurse

git add --all .

git commit -m "Update documentation"

git push origin gh-pages 2>&1 | out-null
```

PR #1094, opened 2021-01-28
[doc] Fix dead link in compilers page to custom type mapping documentation

1lann:
The link on the compilers page (http://microsoft.github.io/bond/manual/compiler.html#type-aliases) to the custom type mapping documentation is dead as it doesn't point to the correct page. Here's a PR to fix it.

### DIFF
--- a/doc/src/compiler.md
+++ b/doc/src/compiler.md
@@ -211,8 +211,9 @@ including a definition of another alias:
 using times = array<time>;
 ```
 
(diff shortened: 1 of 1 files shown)


chwarr (approved):

chwarr:
Thanks. There's [a bug publishing new versions of the docs](https://github.com/microsoft/bond/issues/1086) currently, so this fix won't go live until after that is resolved.

PR #1097, opened 2021-02-14
BOND_FIND_GRPC to enable search for external GRPC installation

bhardwajs:
Added BOND_FIND_GRPC so that when BOND_ENABLE_GRPC is TRUE, an external
GRPC installation can be used. Previous behavior was to build GRPC from
thirdparty subdirectory.


chwarr:
I have other commitments that will preclude me from looking at this for a few weeks.

bhardwajs:
@chwarr - Thinking through this PR, what do you think of changing `BOND_ENABLE_GRPC` so that GRPC-integration requires GRPC as an external library? This would simplify the GRPC workflow and would have the additional advantage that cloning Bond would require cloning two small repositories as submodules.

chwarr:
Bond-over-gRPC's current design assumes that gRPC is the specific pinned version in the submodule. Bond-over-gRPC uses low-level APIs that aren't guaranteed to be stable over point releases.

There have been very few (possibly no) requests to support a range of gRPC versions, so I would hesitate to increase the support matrix that way: the distro package managers can have wildly different ages of gRPC.

chwarr (approved):

PR #1098, opened 2021-02-16
Add Doxygen path to build path of Doc step in CI pipeline

bhardwajs:
Attempt to fix doc build issue #1086 by doing a manual refresh env in Appveyor script after installing Doxygen.

In my local testing when I didn't do refreshenv, doxygen was not found and CMake didn't generate the `documentation` target.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -164,7 +164,7 @@
 
                 $env:Path += "C:\Program Files\doxygen\bin"
 
-                Get-Command pandoc, doxygen
+                Get-Command doxygen,pandoc
 
             }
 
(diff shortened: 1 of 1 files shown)


chwarr:
Is the build even getting to where the `-Force` was added? It sort of looks like the `cmake ../doc` call [on line 270](https://github.com/microsoft/bond/blob/df96fdd321bade00a33a42815575ae735eb6c898/appveyor.yml#L270) is failing, but the build isn't exiting right away.

bhardwajs:
@chwarr - I fixed the docs build issue.

Fixes #1086

chwarr (approved):

PR #1099, opened 2021-03-09
[doc] Switch to robocopy to mirror doc build output

chwarr:
We're having trouble using the Remove-Item/Copy-Item pair to synchronize
the documentation build output with the gh-pages branch.

Switch to robocopy, which has a /MIR (mirror) mode as well as
/XD (exclude directory).

Test $LASTEXITCODE instead of $?. Due to how PowerShell handles the
error stream, $? is set to false if the process writes to stderr, even
if the process exits with 0.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -286,27 +286,33 @@
 
                 cmake --build . --target documentation -- /verbosity:minimal /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"
 
-                if ($? -And $env:BOND_TOKEN -And $env:APPVEYOR_REPO_BRANCH -eq "master") {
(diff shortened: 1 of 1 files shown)


chwarr:
I performed [a test build](https://ci.appveyor.com/project/MicrosoftBond/bond/builds/38140378) without the `git push` step to validate this.

bhardwajs:
Any specific reason not to use `ErrorActionPreference` in Powershell to stop at the first error?
https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.1#erroractionpreference

I should have fixed this when I was looking at the docs build; clearly, I failed to look closer at the output.

chwarr:
We tried `$ErrorActionPreference = 'Stop'` in the past. It doesn't work very well with the native commands the build uses. If a native command were to write to stderr, even if it exits with 0, `$ErrorActionPreference = 'Stop'` will fail the whole build.

Tools like `git` and `stack` and `cmake` report progress and detailed information via stderr sometimes.

PR #1101, opened 2021-03-10
[doc] Handle robocopy exit codes

chwarr:
Robocopy uses exit codes 0-7 to convey details about what was copied.
Exit code 8 starts the range of errors.

Only fail if robocopy exits with a value greater than or equal to 8.


chwarr:
The [previous build of master](https://ci.appveyor.com/project/MicrosoftBond/bond/builds/38144588/job/6vxpdoy6wiarvjd1) with the robocopy change encountered this.

lalo (approved):

PR #1107, opened 2021-04-09
[c#] Exclude files in output dir from implicit codegen

chwarr:
Bond's implicit code generation includes any .bond files in the
project's output directories (e.g., `bin/` or
`obj/debug/netstandard1.0`). Some projects that copied .bond files into
their output directories would then encounter codegen issues when gbc
was run on these .bond files implicitly, particularly during incremental
builds.

Omit any .bond files that match `$(DefaultItemExcludes)` or
`$(DefaultExcludesInProjectFolder)` from implicit codegen.

Add a regression codegen test.

This behavior now matches what the .NET SDK does for .cs files.

### DIFF
--- a/cs/cs.sln
+++ b/cs/cs.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27428.2043
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.31025.218
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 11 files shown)


cs/cs.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27428.2043
+# Visual Studio Version 16
```
bhardwajs, line 3: Is this intentional? Not sure if we are communicating VS version change to 2019.

chwarr: This is just some [metadata that indicates which version of VS last touched the file](https://docs.microsoft.com/en-us/visualstudio/extensibility/internals/solution-dot-sln-file?view=vs-2019).

chwarr:
The person who was hitting the underlying bug has tried a preview release of the package and confirmed the fix is working. Merging now.

Issue #1115, opened 2021-07-26
error C3861: 'GenericWriteVariableUnsigned': identifier not found


jeanga:
When I naively compile using bond, I always get the following error message:

```
7>S:\vcpkg\installed\x64-windows-static\include\bond\protocol\encoding.h(53,5): error C3861: 'GenericWriteVariableUnsigned': identifier not found (compiling source file F:\Projects\Flashback\tests\FlashTest\bond_dictionary.cpp)
7>S:\vcpkg\installed\x64-windows-static\include\bond\protocol\encoding.h(53,5): message : 'GenericWriteVariableUnsigned': function was not declared in the template definition context and can be found only via argument-dependent lookup in the instantiation context (compiling source file F:\Projects\Flashback\tests\FlashTest\bond_dictionary.cpp)
7>S:\vcpkg\installed\x64-windows-static\include\bond\protocol\compact_binary.h(814): message : see reference to function template instantiation 'void bond::WriteVariableUnsigned<Flashback::Details::Bond::Stream,T>(Buffer &,T)' being compiled
7>        with
7>        [
7>            T=uint32_t,
7>            Buffer=Flashback::Details::Bond::Stream
7>        ] (compiling source file F:\Projects\Flashback\tests\FlashTest\bond_dictionary.cpp)
7>S:\vcpkg\installed\x64-windows-static\include\bond\protocol\compact_binary.h(892): message : see reference to function template instantiation 'void bond::CompactBinaryWriter<Flashback::Details::Bond::Stream>::Write<uint32_t>(const T &)' being compiled
7>        with
7>        [
7>            T=uint32_t
7>        ] (compiling source file F:\Projects\Flashback\tests\FlashTest\bond_dictionary.cpp)
7>S:\vcpkg\installed\x64-windows-static\include\bond\protocol\compact_binary.h(892): message : see reference to function template instantiation 'void bond::CompactBinaryWriter<Flashback::Details::Bond::Stream>::Write<uint32_t>(const T &)' being compiled
7>        with
7>        [
7>            T=uint32_t
7>        ] (compiling source file F:\Projects\Flashback\tests\FlashTest\bond_dictionary.cpp)
7>S:\vcpkg\installed\x64-windows-static\include\bond\protocol\compact_binary.h(717): message : see reference to function template instantiation 'void bond::CompactBinaryWriter<Flashback::Details::Bond::Stream>::LengthBegin<Flashback::Details::Bond::Stream>(T &)' being compiled
7>        with
[...]

PR #1118, opened 2021-09-03
[c++] Forward declare GenericWriteVariableUnsigned

chwarr:
Some users are reporting that they are receiving error "C3861:
'GenericWriteVariableUnsigned': identifier not found" when building.

I think this happens when custom streams that _do not_ have their own
`WriteVariableUnsigned` member function.

Add a forward declaration of `GenericWriteVariableUnsigned`.

Fixes https://github.com/microsoft/bond/issues/1115


lalo (approved):

jeanga:
This is exactly what fixed this issue in my use case.
Thanks!

PR #1122, opened 2021-09-10
[c++] Add an ability to apply transform to a schema

ara-ayvazyan:
Sometimes it is necessary to apply a custom transform to a schema, rather than an object or a serialized data. One such use case is the `bond::InitSchemaDef` which is applied to a compile-time schema in order to generate a corresponding runtime one.

One way to achieve that is to apply it to a dummy object. That is not always possible or convenient to do, like in this case,, and we choose to rely on some undefined behavior and fake objects.

There is no canonical way of doing the same for runtime schema, and instead a manual traversal of `bond::SchemaDef` structure is needed, or more tricky ways of using one of the existing protocol readers.

This change is adding two more overloads to `bond::Apply` that would make such transformation possible in a canonical and legal way.

### DIFF
--- a/cpp/inc/bond/core/reflection.h
+++ b/cpp/inc/bond/core/reflection.h
@@ -383,10 +383,29 @@ remove_bonded
 };
 
 
-template <typename T> struct
-remove_bonded<bonded<T> >
+template <typename T, typename Reader> struct
+remove_bonded<bonded<T, Reader> >
 {
     typedef typename remove_bonded<T>::type type;
+};
+
+
+template <typename T> struct
+remove_bonded_value
+{
+    typedef T type;
+};
+
+template <typename T, typename Reader> struct
+remove_bonded_value<bonded<T, Reader> >
+{
+    typedef T type;
+};
+
+template <typename T, typename Reader> struct
+remove_bonded_value<value<T, Reader> >
+{
+    typedef T type;
 };
 
 
--- a/cpp/inc/bond/core/schema.h
+++ b/cpp/inc/bond/core/schema.h
@@ -159,7 +159,18 @@ namespace detail
 
     template <typename T, typename Unused>
     once_flag SchemaCache<T, Unused>::flag;
+
+
+    struct SchemaReader
+    {
+        using Parser = StaticParser<SchemaReader&>;
+    };
 }
+
+
+template <typename Unused> struct
+uses_marshaled_bonded<detail::SchemaReader&, Unused> : std::false_type
+{};
 
 
 //
@@ -189,7 +200,7 @@ public:
     template <typename T>
     bool Base(const T& /*value*/) const
     {
-        TypeDef type = GetTypeDef<T>();
+        TypeDef type = GetTypeDef<typename remove_bonded<T>::type>();
         This().base_def.set(type);
         return false;
     }
@@ -202,7 +213,7 @@ public:
 
         field.id = id;
         field.metadata = metadata;
-        field.type = GetTypeDef<typename remove_maybe<T>::type>();
+        field.type = GetTypeDef<typename remove_bonded_value<T>::type>();
 
         This().fields.push_back(field);
         return false;
@@ -309,15 +320,7 @@ namespace detail
 template <typename T, typename Unused>
 void SchemaCache<T, Unused>::AppendStructDef(SchemaDef* s)
 {
-    // To apply InitSchemaDef transform we need a reference to an
-    // object T. However the transform never accesses the object or
-    // its fields. We can't construct an actual object since we
-    // need to support stateful allocators that can't allocate from
-    // a default-constructed instance and containers which allocate in
-    // default constructor.
-    //
-    // Thus, we make a dummy const T& from a nullptr.
-    Apply(InitSchemaDef(*s), static_cast<const T&>(*static_cast<T*>(0)));
+    Apply<T>(InitSchemaDef{ *s });
 }
 
 } // detail
--- a/cpp/inc/bond/core/schema.h
+++ b/cpp/inc/bond/core/schema.h
@@ -169,7 +169,7 @@ struct SchemaReader
 
 
 template <typename Unused> struct
-uses_marshaled_bonded<detail::SchemaReader&, Unused> : std::false_type
+uses_marshaled_bonded<SchemaReader&, Unused> : std::false_type
 {};
 
 
--- a/cpp/inc/bond/core/schema.h
+++ b/cpp/inc/bond/core/schema.h
@@ -159,13 +159,13 @@ namespace detail
 
     template <typename T, typename Unused>
     once_flag SchemaCache<T, Unused>::flag;
-
-
-    struct SchemaReader
-    {
-        using Parser = StaticParser<SchemaReader&>;
-    };
 }
+
+
+struct SchemaReader
+{
+    using Parser = StaticParser<SchemaReader&>;
+};
 
 
 template <typename Unused> struct
--- a/cpp/inc/bond/core/value.h
+++ b/cpp/inc/bond/core/value.h
@@ -151,7 +151,7 @@ BOND_NO_INLINE void Skip(Reader& input, const std::nothrow_t&)
 }
 
 template <typename T, typename Reader = detail::SchemaReader, typename boost::enable_if<std::is_same<Reader, detail::SchemaReader> >::type* = nullptr>
-void Skip(detail::SchemaReader&, const std::nothrow_t&)
+void Skip(detail::SchemaReader&, const std::nothrow_t&) BOND_NOEXCEPT
 {}
 
 template <typename Reader>
@@ -165,7 +165,7 @@ BOND_NO_INLINE void Skip(Reader& input, const RuntimeSchema& schema, const std::
     {}
 }
 
-inline void Skip(detail::SchemaReader&, const RuntimeSchema&, const std::nothrow_t&)
+inline void Skip(detail::SchemaReader&, const RuntimeSchema&, const std::nothrow_t&) BOND_NOEXCEPT
 {}
 
 template <typename Reader>
@@ -179,7 +179,7 @@ BOND_NO_INLINE void Skip(Reader& input, BondDataType type, const std::nothrow_t&
     {}
 }
 
-inline void Skip(detail::SchemaReader&, BondDataType, const std::nothrow_t&)
+inline void Skip(detail::SchemaReader&, BondDataType, const std::nothrow_t&) BOND_NOEXCEPT
 {}
 
 
--- a/cpp/inc/bond/core/value.h
+++ b/cpp/inc/bond/core/value.h
@@ -152,8 +152,8 @@ BOND_NO_INLINE void Skip(Reader& input, const std::nothrow_t&) BOND_NOEXCEPT
     {}
 }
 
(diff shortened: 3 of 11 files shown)


cpp/test/core/apply_tests.cpp
```diff
@@ -134,6 +136,197 @@ Bonded(uint16_t version = bond::v1)
...
+        }
+
+        _this << ";\n";
+        return false;
+    }
+
+    template <typename T, typename Reader>
+    void Container(const bond::value<T, Reader>& value, std::size_t size) const
+    {
+        UT_AssertAreEqual(size, 0);
```
ara-ayvazyan, line 182: Use `assert`.

cpp/test/core/apply_tests.cpp
```diff
@@ -134,6 +136,197 @@ Bonded(uint16_t version = bond::v1)
...
+            break;
+        }
+
+        _this << "<" << GetTypeName(value) << ">";
+    }
+
+    template <typename K, typename T, typename Reader>
+    void Container(const bond::value<K, Reader>& key, const bond::value<T, Reader>& value, std::size_t size) const
+    {
+        UT_AssertAreEqual(size, 0);
```
ara-ayvazyan, line 202: Use `assert`.

cpp/test/core/apply_tests.cpp
```diff
@@ -134,6 +136,197 @@ Bonded(uint16_t version = bond::v1)
...
+        return false;
+    }
+
+    void UnknownEnd() const noexcept
+    {}
+
+    void End() const
+    {
+        _this << "};\n";
+        _this.flush();
```
ara-ayvazyan, line 229: Remove `flush()` call.

cpp/test/core/apply_tests.cpp
```diff
@@ -134,6 +136,197 @@ Bonded(uint16_t version = bond::v1)
...
+        _out << _this.str();
```
ara-ayvazyan, line 230: Can merge directly with `rdbuf()`.

cpp/inc/bond/core/reflection.h
```diff
@@ -383,13 +383,32 @@ remove_bonded
...
+remove_bonded<bonded<T, Reader> >
 {
     typedef typename remove_bonded<T>::type type;
 };
 
 
+template <typename T> struct
+remove_bonded_value
+{
+    typedef T type;
```
Lyr1k, line 396: why not using?

cpp/inc/bond/core/value.h
```diff
@@ -605,6 +633,24 @@ class value<void, Reader>
 };
 
 
+template <typename Protocols, typename Transform>
+void DeserializeContainer(const Transform& transform, const value<void, detail::SchemaReader&>& element, detail::SchemaReader& input)
+{
+    switch (element.GetTypeId())
+    {
+    case bond::BT_SET:
```
Lyr1k, line 641: what do you think by putting these cases into function to reduce code duplication?

cpp/test/core/apply_tests.cpp
```diff
@@ -134,6 +140,219 @@ Bonded(uint16_t version = bond::v1)
...
+    }
+
+    template <typename T>
+    bool Base(const T& value) const
+    {
+        if (!_structs)
+            return true;
+
+        _this.seekp(-3, std::ios_base::end);
+        _this << " : ";
```
Lyr1k, line 169: why not on 1 line like in `Begin` method?

cpp/test/core/apply_tests.cpp
```diff
@@ -134,6 +140,219 @@ Bonded(uint16_t version = bond::v1)
...
+
+    template <typename T>
+    void FormatValue(const T& value) const
+    {
+        _this << value;
+    }
+
+    void FormatValue(bool value) const
+    {
+        _this << std::boolalpha << value;
```
Lyr1k, line 290: this can be called once?

cpp/test/core/apply_tests.cpp
```diff
@@ -134,6 +140,219 @@ Bonded(uint16_t version = bond::v1)
...
+    }
+
+    void FormatValue(const std::wstring& value) const
+    {
+        std::string str(value.size(), ' ');
```
Lyr1k, line 295: just `std::for_each` without `std::string`?

cpp/test/core/apply_tests.cpp
```diff
@@ -190,6 +413,14 @@ void ApplyTests(const char* name)
     AddTestCase<TEST_ID(N), AllTests, Reader, Writer>(suite, "Use generated *_apply.cpp");
 }
 
+template <uint16_t N>
```
Lyr1k, line 416: minor: `std`?

cpp/inc/bond/core/reflection.h
ara-ayvazyan: To be in-sync with existing old style in this header.

cpp/inc/bond/core/value.h
ara-ayvazyan: This way it reads identical as with other similar checks here.

cpp/inc/bond/core/schema.h
```diff
@@ -159,9 +159,20 @@ namespace detail
 
     template <typename T, typename Unused>
     once_flag SchemaCache<T, Unused>::flag;
+
+
+    struct SchemaReader
+    {
+        using Parser = StaticParser<SchemaReader&>;
```
ara-ayvazyan, line 166: Use `typedef` for now.

cpp/test/core/apply_tests.cpp
ara-ayvazyan: This one can be.

cpp/test/core/apply_tests.cpp
ara-ayvazyan: Yep.

cpp/test/core/apply_tests.cpp
ara-ayvazyan: Same, trying to align with the existing style in the file at least.

cpp/inc/bond/core/schema.h
```diff
@@ -189,7 +200,7 @@ class InitSchemaDef
     template <typename T>
     bool Base(const T& /*value*/) const
     {
-        TypeDef type = GetTypeDef<T>();
+        TypeDef type = GetTypeDef<typename remove_bonded<T>::type>();
```
chwarr, line 203: When is `Base` invoked with a `bonded<>`?

In IDL, `struct Foo : bonded<Bar>` is illegal.

cpp/inc/bond/core/schema.h
```diff
@@ -202,7 +213,7 @@ class InitSchemaDef
 
         field.id = id;
         field.metadata = metadata;
-        field.type = GetTypeDef<typename remove_maybe<T>::type>();
```
chwarr, line 205: Is the `remove_maybe` not needed because `FieldTemplate` already does this?

cpp/inc/bond/core/schema.h
```diff
@@ -202,7 +213,7 @@ class InitSchemaDef
...
+        field.type = GetTypeDef<typename remove_bonded_value<T>::type>();
```
chwarr, line 216: It looks to me like the `remove_bonded_value` here means that `type.bonded_type` won't be set properly anymore. Am I missing something?

cpp/test/core/apply_tests.cpp
```diff
@@ -134,6 +141,210 @@ Bonded(uint16_t version = bond::v1)
...
+            _structs = {};
+    }
+
+    template <typename T>
+    bool Base(const T& value) const
+    {
+        if (!_structs)
+            return true;
+
+        _this.seekp(-3, std::ios_base::end);
```
chwarr, line 169: Minor: a comment for the -3 would be nice. 'Removing "\\n{\\n" to add base'?

cpp/test/core/apply_tests.cpp
```diff
@@ -134,6 +141,210 @@ Bonded(uint16_t version = bond::v1)
...
+    void Container(std::size_t size, const T&... value) const
+    {
+        BOOST_VERIFY(size == 0);
+        assert(bond::BT_LIST <= _container && _container <= bond::BT_MAP);
+        assert((sizeof...(T) == 2) == (_container == bond::BT_MAP));
+
+        const char* labels[] = { "list", "set", "map" };
+        _this << labels[_container - bond::BT_LIST] << '<';
+        (void)std::initializer_list<int>{ (TypeName(value), _this << ", ", 0)... };
+        _this.seekp(-2, std::ios_base::end);
```
chwarr, line 259: Minor: like above, a comment for the -2 would be nice. 'Removing \", \" from after last type'?

cpp/test/core/apply_tests.cpp
```diff
@@ -134,6 +141,210 @@ Bonded(uint16_t version = bond::v1)
...
+
+    template <typename T>
+    void FormatValue(const T& value) const
+    {
+        _this << value;
+    }
+
+    void FormatValue(const std::wstring& value) const
+    {
+        std::transform(value.begin(), value.end(), std::ostreambuf_iterator<char>(_this), [](wchar_t wc) { return char(wc); });
```
chwarr, line 289: I guess none of the default string values in the test cases use non-ASCII values? ;-)

chwarr (changes requested):
LGTM, overall. I have some clarifying questions and some minor changes to improve readability.

cpp/inc/bond/core/schema.h
ara-ayvazyan: Originally, `T` used to come as the base struct itself, since the transform was performed over an object. While now structs are coming as `bonded<T, Reader>` (see the `StaticParser`). That is the reason why `remove_bonded` applies recursively (at most twice), because in cases when the field is a `bonded<T>`, then the transform's `Field` will receive `bonded<bonded<T>>`. In case of a `Base`, that won't happen.

cpp/inc/bond/core/schema.h
ara-ayvazyan: Same reason. The `T` used to be the actual value type (i.e. `FieldT::value_type`) which could be a `maybe<FieldT::field_type>`. While now we are receiving a `value` or ` bonded` of `FieldT::field_type`. Here we do not really care. Otherwise, `FastPathType` would call into an overload that receives the full `FieldT`.

cpp/inc/bond/core/schema.h
ara-ayvazyan: It should work properly I believe (tests are covering these and I looked manually as well). Note, that unlike `remove_bonded`, the new `remove_bonded_value` removes the `bonded` only once (in case of a `bonded` field we receive `bonded<bonded<T>>`). That is the reason why I added this combined trait, instead of adding another `remove_value` and using it in combination with existing `remove_bonded` .

cpp/test/core/apply_tests.cpp
ara-ayvazyan: Yep, will update.

cpp/test/core/apply_tests.cpp
ara-ayvazyan: Yeah, none :) We could replace all these to use wide strings to support both.

chwarr (approved):

PR #1123, opened 2021-09-14
[c++] Remove unnecessary blob::_content member

ara-ayvazyan:


CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ different versioning scheme, following the Haskell community's
 * Fixed `bond::DynamicParser` that may not emit transform's `OmittedField`
   for compile-time schema and an omitted optional field in the payload.
   ([Issue \#1120](https://github.com/microsoft/bond/issues/1120))
+* Removed `bond::blob`'s unnecessary `_content` member, making the type
+  smaller (32 -> 24 bytes).
```
Lyr1k, line 32: remove platform specific size?

Lyr1k (approved):

cpp/inc/bond/core/blob.h
```diff
@@ -87,42 +81,36 @@ class blob
...
     /// Not recommended for performance reasons. Use boost::shared_ptr whenever possible.
     template <typename T, template <typename U> class SmartPtr>
     blob(const SmartPtr<T>& buffer, uint32_t offset, uint32_t length)
-        : _buffer(wrap_in_shared_ptr(buffer)),
-          _content(bond::detail::checked_add(_buffer.get(), offset)),
+        : _buffer(wrap_in_shared_ptr(buffer, offset)),
           _length(length)
     {
-        bond::detail::checked_add(_content, length);
+        bond::detail::checked_add(_buffer.get(), length);
```
biocomp, line 97: Use `content()` or `data()` instead of `_buffer.get()` in most places?

cpp/inc/bond/core/blob.h
```diff
@@ -416,7 +391,7 @@ inline T blob_cast(const blob& from)
 template <typename A>
 inline blob blob_prolong(blob src, const A& allocator)
 {
-    if (src._buffer)
+    if (src._buffer.use_count() != 0)
```
biocomp, line 394: Is using `use_count()` safe?

biocomp (approved):

cpp/inc/bond/core/blob.h
ara-ayvazyan: Outside of the scope of this PR. The whole class needs a clean and modern rewrite.

cpp/inc/bond/core/blob.h
ara-ayvazyan: Here we are checking for the ownership presence (i.e. not the actual reference count). That should be safe even for concurrent access.

chwarr (changes requested):
LGTM with a minor request for a comment to be added.

cpp/inc/bond/core/blob.h
```diff
@@ -38,47 +37,42 @@ class blob
     ///
     /// Not recommended because of buffer lifetime management.
     blob(const void* content, uint32_t length)
-        : _buffer(),
-          _content(static_cast<const char*>(content)),
+        : _buffer(boost::shared_ptr<const void>(), static_cast<const char*>(content)),
```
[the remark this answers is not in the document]
ara-ayvazyan: This looked obvious to me, but will add.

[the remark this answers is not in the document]
ara-ayvazyan: Let me know if the wording of the comment looks ok.

PR #1125, opened 2021-09-23
[c++] Specify blob's non-throwing functions as noexcept

ara-ayvazyan:


cpp/inc/bond/core/blob.h
```diff
@@ -173,66 +173,66 @@ class blob
...
+    void swap(blob& src) BOND_NOEXCEPT
     {
         std::swap(_length, src._length);
         _buffer.swap(src._buffer);
     }
 
     /// @brief Clear reference to the underlying memory buffer and reset the
     /// blob to empty
-    void clear()
+    void clear() BOND_NOEXCEPT
```
biocomp, line 184: Not sure we can assume `shared_ptr` destruction doesn't throw. With aliasing constructor you could put all kinds of nasty stuff in there :)

biocomp (changes requested):

cpp/inc/bond/core/blob.h
ara-ayvazyan: The `~shared_ptr` is not marked as `noexcept(false)`, which means it is implicitly `noexcept` as any other destructor. Also the `shared_ptr<T>::reset` is explicitly marked as `noexcept`.

biocomp: Ah, I now see that `shared_ptr`'s destructor (both for boost and std) is `noexcept`, so nevermind :)

biocomp (approved):

chwarr (approved):

PR #1130, opened 2021-10-16
Fixes multiple symbol definition for Win32Exception in grpc/win_thread_pool.h

### LINKED ISSUE #1129: multiply symbols errors from: Win32Exception(const char* message)
Getting multiply symbol definition errors that appear to be a result of multiple includes from file:
vcpkg\installed\x64-windows\include\bond\ext\grpc\win_thread_pool.h

`    namespace detail
    {
        BOND_NORETURN void Win32Exception(const char* message)
        {
            throw std::system_error{
                std::error_code{ static_cast<int>(::GetLastError()), std::system_category() },
                message };
        }

    } // namespace detail
`
I was able to correct by adding inline specifier to the header file code definition i.e.:
inline BOND_NORETURN void Win32Exception(const char* message)

I'm i missing something fundamental here?

bhardwajs:
`bond::ext::grpc::detail::Win32Exception` is not marked inline which could lead to multiple symbol definition. This PR fixes that by adding `inline` to that function.

Fixes #1129


cpp/inc/bond/ext/grpc/win_thread_pool.h
```diff
@@ -25,7 +25,7 @@ namespace bond { namespace ext { namespace grpc
 {
     namespace detail
     {
-        [[noreturn]] void Win32Exception(const char* message)
+        inline [[noreturn]] void Win32Exception(const char* message)
```
chwarr, line 28: Minor: the rest of the Bond code uses `[[noreturn]] inline` in that order:

```suggestion
        [[noreturn]] inline void Win32Exception(const char* message)
```

bhardwajs: Changed order and force pushed an update.

PR #1137, opened 2022-02-25
[linux] Bump CI image to one built 2022-02-25

chwarr:


chwarr:
Looks like we may need to pin to Gradle 6. Testing that locally.

chwarr:
I have a proposed fix for the Gradle 6 error. Please see [PR #1139, \[java\] Pin to Gradle 6](https://github.com/microsoft/bond/pull/1139)

PR #1140, opened 2022-03-05
Update Install-Boost.ps1

iamhardikat11:
I have tried implementing the functionality that if the version of boost is >=1.66 then the .dll and .lib files should be placed together in the *-x32-* or *-x64-* folders respectively, but the other functionality i am confused how will the .dll and .lib will get segregated.


tools/ci-scripts/windows/Install-Boost.ps1
```diff
@@ -122,14 +122,18 @@ function Install-BoostComponent([string]$Component)
             -PackageId $packageId `
             -InstallDir $workDir `
             -PackageVersion $Version
-
+        if($Version < 1.66){
```
chwarr, line 125: Comparing versions as either numbers or strings will not work correctly. A numeric sort won't handle versions like 1.2.3, and string sorting fails on versions like 10.1 compared to 2.8.

Try casting the version to `System.Version` and comparing those:

```powershell
if ([System.Version]$Version -lt [System.Version]"1.66") {
```

tools/ci-scripts/windows/Install-Boost.ps1
```diff
@@ -122,14 +122,18 @@ function Install-BoostComponent([string]$Component)
...
         Move-Item `
-            -Path ([System.IO.Path]::Combine($workDir, $packageId, 'lib', 'native', 'address-model-32', 'lib', '*')) `
+            -Path ([System.IO.Path]::Combine($workDir, $packageId, 'lib', 'native', '*-x32-*', 'lib', '*')) `
             -Destination $lib32Dir
 
         Move-Item `
-            -Path ([System.IO.Path]::Combine($workDir, $packageId, 'lib', 'native', 'address-model-64', 'lib', '*')) `
+            -Path ([System.IO.Path]::Combine($workDir, $packageId, 'lib', 'native', '*-x64-*', 'lib', '*')) `
```
chwarr, line 131: Under the 1.66+ model, I believe that all the files are under `lib\native` with names like `foo-x32-stuff.dll`. This copy command is trying to copy a directory that doesn't exist right now, it looks.

If I were to do this, I'd install a copy of some of the packages into a temp directory and see how they're laid out.

chwarr (changes requested):
Thanks for starting this PR!

You asked:

> i am confused how will the .dll and .lib will get segregated.

I'd first try copying all the `*x32*`/`*x64*` files into the respective 32- and 64-bit output directories that the script is already creating. Then, try using that as a BOOST_LIBRARYDIR when you run `cmake` to generate build files for Bond.

If that doesn't work, you'll want to read the FindBoost CMake package and figure out what it's expecting and then implement that. I don't know what it's looking for myself...

iamhardikat11:
Sure I will try all those requested changes.

tools/ci-scripts/windows/Install-Boost.ps1
iamhardikat11: I could not understand the meaning of the above suggestion, can you elaborate a bit more, on the file structure, and about .dll and .ldb files in the directory.

chwarr: I suggest that you get the Bond CMake C++ build working with Boost 1.65. Then, download the Boost 1.66 packages, extract them, and try to get the build working with those packages. You'll need to delete your `build/` directory each time you try something new so that CMake regenerates from scratch. Things I'd try:

* Pointing `$env:BOOST_LIBRARYDIR` at the `lib\native` directory without renaming any thing.
* Create an `x32` directory and move all the `*x32*` files into that directory.
* Create an `x32` directory and move all the `*x32*` files into that directory and strip the `x32` component of the file name.

chwarr:
It has been over a year since there has been activity on this PR. I am going to close it for now. If you are still interested in working on this, let me know, and I can re-open this. Thanks.

PR #1148, opened 2022-04-29
[ci] update ci to use ubuntu 18

lalo:


bhardwajs:
@lalo - This PR looks good to merge. We should also have the commits squashed in one. It seems you need to sign the CLA as well. Thanks!

Issue #1143, opened 2022-03-30
Deprecating Bond support for EOL platforms and toolchains


bhardwajs:
Bond is currently supporting Ubuntu 14.04 with Clang 3.8 and g++ 5.4 and VS 2017 on Windows. Ubuntu 14.04 is past support as is VS 2017. Should we update the dependencies to?
- Ubuntu 20.04 and the versions of Clang & GCC it ships with
- VS 2019 atop Server 2019
My read is that the versions of gcc/clang/MSVC would support C++ 17. Should we bump up to C++ 17? 

We can also Boost dependency to require version 1.73 or later which was released on Apr 28, 2020 almost two years ago (https://www.boost.org/users/history/).

For .Net, we can set support surface to
- .Net Framework 4.8+
- .Net Core 3.1
- .Net 5+

chwarr:
All the minimums proposed here LGTM.

PR #1155, opened 2022-06-26
[c#] add support for init-only setters on generated classes

### LINKED ISSUE #1154: [C#] Support for init properties on generated classes
C# 9 introduced [init properties](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/init) that allow using object initializers to set readonly properties. This is quite useful for enforcing the copy-on-write pattern for immutable objects:

```cs
class Person
{
    public string FirstName { get; init; }
    public string LastName { get; init; }

    public Person(Person copy)
    {
        this.FirstName = copy.FirstName;
        this.LasName = copy.LastName;
    }    
}

var johnDoe = new Person { FirstName = "John", LastName = "Doe" };
var janeDoe = new Person(johnDoe) { FirstName = "Jane" };
```

I see the Bond compiler CLI already has a `--readonly-properties` switch for making `private set` properties on generated classes, but that is too restrictive for this use case. Is it possible to add a new switch such as `--init-properties` to allow the above code sample on a Bond-generated class?

Related: https://github.com/microsoft/bond/issues/439

enioluwas:
Closes #1154 by introducing an `--init-only-properties` switch for the Bond C# compiler.

Mostly just mimicked the implementation for the `--readonly-properties` switch. This was tricky to add unit tests for, because the `init` keyword was introduced in C# 9 / .NET 5. Please review and see if it's okay as-is. I considered adding a new unit test project for .NET 5+, but it wouldn't be compatible with VS 2017.

### DIFF
--- a/compiler/Main.hs
+++ b/compiler/Main.hs
@@ -112,11 +112,11 @@ csCodegen options@Cs {..} = do
     typeMapping = if collection_interfaces
             then csCollectionInterfacesTypeMapping
             else csTypeMapping
-    fieldMapping = if readonly_properties
-            then ReadOnlyProperties
-            else if fields
-                 then PublicFields
-                 else Properties
+    fieldMapping
+      | readonly_properties = ReadOnlyProperties
+      | init_only_properties = InitOnlyProperties
+      | fields = PublicFields
+      | otherwise = Properties
     constructorOptions = if constructor_parameters
             then ConstructorParameters
             else DefaultWithProtectedBase
--- a/compiler/Options.hs
+++ b/compiler/Options.hs
@@ -53,6 +53,7 @@ data Options
         , namespace :: [String]
         , collection_interfaces :: Bool
         , readonly_properties :: Bool
+        , init_only_properties :: Bool
         , fields :: Bool
         , jobs :: Maybe Int
         , no_banner :: Bool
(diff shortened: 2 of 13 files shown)


chwarr (changes requested):
Overall, LGTM. Can you add a changelog entry? This is a minor version bump for C#.

I forgot that `init` properties were a C# 9+ and .NET 5+ feature. Sigh. As you've noted, the CI build only uses the .NET Core 3.1 SDK today.

Bond still needs to support .NET Framework, so we can't require .NET 5 or newer. In [issue #1143, Deprecating Bond support for EOL platforms and toolchains](https://github.com/microsoft/bond/issues/1143), @bhardwajs proposed a new minimum set of supported platforms, including updating the Windows CI build to Visual Studio 2019/`dotnet`. That needs to land before this can be merged. I don't see a way to test this new code without using a newer toolchain to build...

I'm not comfortable merging in code that doesn't have CI test coverage.

enioluwas:
Alright, will hold off on this, and consider helping with #1143

chwarr:
It has been more than six months since there have activity on this PR. I am going to close it for now. If you are still interested in working on this, let me know, and I can re-open this. The .NET build changes will still need to come first. Thanks.

PR #1161, opened 2022-08-12
[c#] support for system.collections.immutable collections

### LINKED ISSUE #1159: [C#] Support System.Collections.Immutable collections as custom containers
C# [immutable collections](https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable?view=net-6.0#remarks) are useful data types for building truly immutable Bond structs. 

Bond doesn't properly support using them as custom type aliases, for the following reasons:
1. Generated classes always use the `new ImmutableX()` as the default constructor, but those do not exist for immutable collections; they are instead constructed with the static field `ImmutableX<T>.Empty`.
2. Serialization works, but deserialization fails since the compiler assumes that all BT_LIST/BT_SET aliases properly implement `ICollection<T>.Add()`, but all mutation methods from `ICollection<T>` [throw a NotSupportedException](https://source.dot.net/#System.Collections.Immutable/System/Collections/Immutable/ImmutableList_1.cs,935) for immutable collections. Same issue with BT_MAP, but with the setter `IDictionary<K, V>.Item[]`. 

The desired mappings would be:

BT_LIST -> [ImmutableArray&lt;T&gt;](https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutablearray-1?view=net-6.0), [ImmutableList&lt;T&gt;](https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutablelist-1?view=net-6.0)

BT_SET -> [ImmutableHashSet&lt;T&gt;](https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutablehashset-1?view=net-6.0), [ImmutableSortedSet&lt;T&gt;](https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutablesortedset-1?view=net-6.0)

BT_MAP -> [ImmutableDictionary&lt;T&gt;](https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutabledictionary-2?view=net-6.0), [ImmutableSortedDictionary&lt;T&gt;](https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutablesorteddictionary-2?view=net-6.0)

Based on my initial investigation, the following would need to be done:
1. Modify the C# compiler to detect immutable collections and use the static field to construct them, rather than a parameterless constructor. 
2. Modify the C# deserializer to specially handle reconstructing immutable collections, either using their [builders](https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutablelist-1.builder?view=net-6.0) as an intermediary, using the specific `Add` methods for the BT_LIST/BT_SET objects, and the `SetItem` method for the BT_MAP objects.

enioluwas:
Closes #1159

Adds proper support for immutable collections as laid out in the issue description. Implemented this in a way such that the Bond C# package doesn't actually depend on System.Collections.Immutable, roughly following how it was done [in Newtonsoft.Json](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Utilities/ImmutableCollectionsUtils.cs). 

The collections are constructed using the builders, which performs a lot better doing bulk operations.

The change to the codegen in Haskell is small, but since I don't have much experience, any suggestions to make it more idiomatic are welcome. 

Added unit tests both for the compiler and for [de]serialization in C#.

### DIFF
--- a/cs/src/core/expressions/DeserializerTransform.cs
+++ b/cs/src/core/expressions/DeserializerTransform.cs
@@ -83,8 +83,9 @@ namespace Bond.Expressions
         static readonly MethodInfo bufferBlockCopy =
             Reflection.MethodInfoOf((byte[] a) => Buffer.BlockCopy(a, default(int), a, default(int), default(int)));
 
-        // Immutable collection types are represented/identified as strings to avoid depending on the
-        // System.Collections.Immutable nuget package
+        // Immutable collection types are represented/identified as strings
+        // to avoid depending on the System.Collections.Immutable assembly
+        // or NuGet package
         static readonly HashSet<string> immutableListSetTypeNames = new HashSet<string>
         {
             "System.Collections.Immutable.ImmutableArray`1",
--- a/cs/src/core/expressions/DeserializerTransform.cs
+++ b/cs/src/core/expressions/DeserializerTransform.cs
@@ -83,6 +83,24 @@ namespace Bond.Expressions
         static readonly MethodInfo bufferBlockCopy =
             Reflection.MethodInfoOf((byte[] a) => Buffer.BlockCopy(a, default(int), a, default(int), default(int)));
 
+        // Immutable collection types are represented/identified as strings to avoid depending on the
+        // System.Collections.Immutable nuget package
+        static readonly HashSet<string> immutableListSetTypeNames = new HashSet<string>
+        {
+            "System.Collections.Immutable.ImmutableArray`1",
+            "System.Collections.Immutable.ImmutableList`1",
+            "System.Collections.Immutable.ImmutableHashSet`1",
+            "System.Collections.Immutable.ImmutableSortedSet`1",
+        };
+
+        static readonly HashSet<string> immutableMapTypeNames = new HashSet<string>
+        {
+            "System.Collections.Immutable.ImmutableDictionary`2",
+            "System.Collections.Immutable.ImmutableSortedDictionary`2",
+        };
+
+        static readonly HashSet<string> immutableCollectionTypeNames = new HashSet<string>(immutableListSetTypeNames.Concat(immutableMapTypeNames));
+
         public DeserializerTransform(
             Expression<Func<R, int, object>> deferredDeserialize,
             Factory factory,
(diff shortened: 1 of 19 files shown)


chwarr:
I have other commitments that I need to attend to at the moment, so I won't be able to review this PR for a while. It's on my to do list, however. I cannot give an estimate on when I will be able to review this, unfortunately.

enioluwas:
Hey @chwarr any chance you could prioritize reviewing this? My team already depends on Bond for schema definitions, and partially for serialization. This feature is the last blocker in moving all our serialization to Bond, which would be a performance boost for us over JSON. Thanks!

cs/src/core/expressions/DeserializerTransform.cs
```diff
@@ -83,6 +83,22 @@ internal class DeserializerTransform<R>
         static readonly MethodInfo bufferBlockCopy =
             Reflection.MethodInfoOf((byte[] a) => Buffer.BlockCopy(a, default(int), a, default(int), default(int)));
 
+        static readonly HashSet<string> immutableListSetTypeNames = new HashSet<string>
```
chwarr, line 86: Minor: a comment explaining why these are represented as strings (to avoid an assembly dependency) would be helpful here.

cs/src/core/expressions/DeserializerTransform.cs
```diff
@@ -408,6 +424,33 @@ Expression Container(IParser parser, Expression container, Type schemaType, bool
...
+
+                            return Expression.Block(
+                                new[] { item, builder },
+                                constructBuilder,
+                                beforeLoop,
+                                ControlExpression.While(next, addItem),
+                                afterLoop,
+                                reconstructImmutable);
+                        }
+
```
chwarr, line 453: Minor: could you put the `ICollection.Add` code path in an `else` branch of `if (immutableListSetTypeNames.Contains(containerGenericTypeDef.FullName))`?

chwarr (changes requested):
Overall, LGTM. Thank you for the PR and for your patience.

Can you add:

* an example project,
* a note in the documentation (under [the Custom Containers section](https://github.com/microsoft/bond/blob/master/doc/src/bond_cs.md#custom-containers)?) about the special handing for the immutable types, and
* an entry in the changelog under Unreleased.

Thanks.

cs/src/core/expressions/DeserializerTransform.cs
enioluwas: Added explanatory comment

cs/src/core/expressions/DeserializerTransform.cs
enioluwas: Done, also conformed to the pattern of just defining the necessary variables so there's a single return at the end

enioluwas:
> Overall, LGTM. Thank you for the PR and for your patience.
> 
> Can you add:
> 
> * an example project,
> * a note in the documentation (under [the Custom Containers section](https://github.com/microsoft/bond/blob/master/doc/src/bond_cs.md#custom-containers)?) about the special handing for the immutable types, and
> * an entry in the changelog under Unreleased.
> 
> Thanks.

Done! Added the `immutable_collections` example project. Let me know if the wording on any of the doc changes needs to be modified

chwarr (approved):
I'll merge when the CI build for my minor changes passes.

Issue #1164, opened 2022-08-30
[ci] Image cleanup script needs to handle 18.04 images


chwarr:
The [Docker image cleanup script](https://github.com/microsoft/bond/tree/master/tools/ci-scripts/linux/image-cleanup) doesn't properly handle that we now have both Ubuntu 16.04 and Ubuntu 18.04 images. It's only trying to clean up the 16.04 ones.

PR #1166, opened 2022-08-30
[ci] collect image for other images

lalo:
solves #1164


lalo:
@chwarr not sure which python format tool you used before, tried black but it's messing to many things

chwarr:
I don't think I used a Python formatting tool. I used Mypy for static typing. I probably used Emacs' python-mode for formatting. The indentation style where arguments are indented to the opening parenthesis looks like what python-mode would do:

```python
    parser.add_argument('root_filters',
                        nargs='+',
                        type=semi_list,
                        help="""semi-colon seperated list of arguments for one invocation of `git
                        rev-parse`. The commits in the union of these
                        rev-parse invocations are used to find container images
                        that are referenced by source code. Ex:
                        '--remotes=origin;--since=2~weeks~ago' or '--tags;-n;1'""")
```

Issue #1156, opened 2022-07-12
Potential breaking change: Upgrade to exclusively use Newtonsoft.Json >= 13.0.1


ankitC:
In the wake of a new vulnerability discovered in the Newtonsoft.Json library, it is recommended that consumers use version greater than or equal to 13.0.1. (https://github.com/JamesNK/Newtonsoft.Json/issues/2457)

In order to comply with the recommendation, I propose the community to consider upgrading the dependency on Netwonsoft.Json to the recommended version. It will help all the consumers of the Bond library to make their applications and services inherently more secure.

The Bond library states it's dependency as >= 9.0.1, and consumers are free to use any version that suits their needs. While this recommendation is not incorrect, it's not in the best interests of the consumers as we have the knowledge of the Newtonsoft issue. As a best practice, the latest version of the library should be using the most secure dependencies.

If there are enough community members that agree to this proposal, I'd be happy to work with the community and contribute to the proposed change.

chwarr:
I don't see a compelling reason to take this breaking change in Bond, so I'm going to leave Bond at the minimum version of Newtonsoft.Json that is requires.

PR #1169, opened 2022-09-01
Improper Handling of Exceptional Conditions

imhunterand:
## Describe the bugs: 🐛
Newtonsoft.Json prior to version 13.0.1 is vulnerable to Insecure Defaults due to improper handling of expressions with high nesting level that lead to StackOverFlow exception or high CPU and RAM usage. Exploiting this vulnerability results in Denial Of Service (DoS).

The serialization and deserialization path have different properties regarding the issue.
Deserializing methods (like `JsonConvert.DeserializeObject`) will process the input that results in burning the CPU, allocating memory, and consuming a thread of execution. Quite high nesting level (>10kk, or 9.5MB of `{a:{a:{...` input) is needed to achieve the latency over 10 seconds, depending on the hardware. 

Serializing methods (like `JsonConvert.Serialize` or `JObject.ToString`) will throw StackOverFlow exception with the nesting level of around 20k.

To mitigate the issue one either need to update Newtonsoft.Json to 13.0.1 or set `MaxDepth` parameter in the `JsonSerializerSettings`. This can be done globally with the following statement. After that the parsing of the nested input will fail fast with `Newtonsoft.Json.JsonReaderException`:
```
JsonConvert.DefaultSettings = () => new JsonSerializerSettings { MaxDepth = 128 };
```
Repro code:
```
//Create a string representation of an highly nested object (JSON serialized)
int nRep = 25000;
string json = string.Concat(Enumerable.Repeat("{a:", nRep)) + "1" +
 string.Concat(Enumerable.Repeat("}", nRep));

//Parse this object (leads to high CPU/RAM consumption)
var parsedJson = JsonConvert.DeserializeObject(json);

// Methods below all throw stack overflow with nRep around 20k and higher
// string a = parsedJson.ToString();
// string b = JsonConvert.SerializeObject(parsedJson);
```
Note the original statement about the problem only affecting IIS applications is misleading. Any application is affected, however the IIS has a behavior that stops restarting the instance after some time resulting in a harder-to-fix DoS.

GHSA-5crp-9r3c-p9vr
`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H`
Severity High
`7.5/ 10`

### DIFF
--- a/examples/cs/core/simple_json/simple_json.csproj
+++ b/examples/cs/core/simple_json/simple_json.csproj
@@ -17,7 +17,7 @@
     <ProjectReference Include="..\..\..\..\cs\src\core\Bond.csproj" />
     <ProjectReference Include="..\..\..\..\cs\src\io\IO.csproj" />
     <ProjectReference Include="..\..\..\..\cs\src\json\JSON.csproj" />
-    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
   </ItemGroup>
   <Import Project="$(BOND_PATH)\build\nuget\Bond.CSharp.targets" />
 </Project>


chwarr:
Thanks for the PR. This is not a change we're going to make to Bond right now.

The minimum version of Newtonsoft.JSON that Bond currently needs is 9.0.1, so this is the version referenced in the example.

See also: [#1156, Potential breaking change: Upgrade to exclusively use Newtonsoft.Json >= 13.0.1](https://github.com/microsoft/bond/issues/1156) and [my comment](https://github.com/microsoft/bond/issues/1156#issuecomment-1182453809) explaining why I don't think this makes sense for Bond. It also outlines a handful of approaches you can take to use a newer version of Newtosoft.JSON in your app should you need one.

PR #1170, opened 2022-09-13
Fix missing cmake call in build instruction for Linux and MacOS

ThomsonTan:
CMake should be invoked to generate build script before running make, or there is nothing to make int he create build directory.


chwarr:
Whoops! Thanks for catching this and fixing it.

PR #1174, opened 2022-11-29
[ci] Revert happy install workaround

chwarr:
CI builds are failing with linker errors trying to install happy. This workaround no longer appears to be needed.

Remove the manual install step.


chwarr:
This is the same as #1173, but it targets the correct branch now. I accidentally targeted the wrong branch in that one...

lalo (approved):

PR #1177, opened 2022-12-21
[C#] Prevent null references in CompactBinary serialization when a required string is null

danslocombe:


danslocombe:
> @microsoft-github-policy-service agree [company="{your company}"]

@microsoft-github-policy-service agree [company="Microsoft"]

danslocombe:
@microsoft-github-policy-service agree company="Microsoft"

chwarr:
Thanks for the PR. This change takes Bond in a direction that we don't want it to go in. In Bond, a zero-length string is different from a null string. There is no sensible serialized representation of a Bond `string` field for C#'s `null` value. `nullable<string>` is the Bond way of saying that a string field allows null.

Serialization does not employ explicit null checks, as they are slower in our microbenchmarks for serialization. Bond relies on the the .NET runtime to catch these nulls.

I'm going to close this PR for now. Let me know if you have any other questions.

Issue #1180, opened 2023-02-20
cmake/Config.cmake looks for the EOLed Python-2.7


yurivict:

chwarr:
A PR to repoint it a Python 3 would be welcome.

PR #1192, opened 2023-08-09
enhanced to python3 #1180

VenkateshSoni:
Changed the line `set(GBC_EXECUTABLE ${BOND_GBC_PATH})` to `set(GBC_EXECUTABLE ${PYTHON3_EXECUTABLE} ${CMAKE_CURRENT_LIST_DIR}/gbc.py)`. This change is necessary because the gbc executable is now written in Python, so we need to tell CMake to use the Python interpreter to run it.
Added the line `find_package(Python3)` to the beginning of the file. This line tells CMake to find the Python interpreter, so that it can be used to run the gbc executable.

### DIFF
--- a/cmake/Config.cmake
+++ b/cmake/Config.cmake
@@ -18,7 +18,7 @@ find_program (BOND_GBC_PATH "gbc"
     NO_DEFAULT_PATH)
 
 if (BOND_GBC_PATH)
-    set (GBC_EXECUTABLE ${BOND_GBC_PATH})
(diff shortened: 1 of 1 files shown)


chwarr (changes requested):
gbc is written in Haskell, not Python. Can you expand on what you're trying to do in this PR?

If you are trying to change Bond to build against Python 3 instead of Python 2, I would expect that changes are needed to the Linux and Windows CI build pipelines as well as some CMake changes.

I would not expect needing to run gbc via Python.

chwarr:
I'm going to close this PR. If you want to re-open it in the future, let us know.

Issue #1199, opened 2024-03-06
Upgrade to Stack lts-22.12 for aarch64 support


Freddicus:
Building `bond` on Apple Silicon is impossible today. The `haskell-stack` installed during setup provides the system with ghc-9.8.2, which is rejected by the Stack resolver `lts-14.4` (`ghc-8.6.5`) defined in the `compiler/stack.yaml` file.

You can see it rejects newer versions of ghc.

```console
   2024-03-06 09:12:13.992344: [debug] Not using compiler at
  "/opt/homebrew/Cellar/ghc/9.8.2/bin/ghc-9.8.2": PrettyException
  UnwantedCompilerVersion
```

Changing the resolver to `lts-22.12` allows the build to pass the initial checks, but compilation ultimately fails on the bond source code. I imagine there are breaking changes migrating from ghc 8.x to 9.x, but Apple Silicon support wasn't added until 9.2.x and later (and is still experimental, IIRC).

Example 1:

```console
/Users/<me>/git-projects/bond/compiler/src/Language/Bond/Syntax/JSON.hs:84:19: error: [GHC-39999]
    • No instance for ‘FromJSON TypeParam’ arising from a use of ‘.:’
    • In the second argument of ‘(<$>)’, namely ‘o .: "value"’
      In the expression: BT_TypeParam <$> o .: "value"
      In a case alternative:
          String "parameter" -> BT_TypeParam <$> o .: "value"
   |
84 |                 o .: "value"
   |                   ^^
```

Example 2:

```console
/Users/<me>/git-projects/bond/compiler/src/Language/Bond/Syntax/Util.hs:64:32: error: [-Wunused-record-wildcards, Werror=unused-record-wildcards]
    No variables bound in the record wildcard match are used
      Possible fix: omit the ‘..’
   |
64 | isScalar (BT_UserDefined Enum {..} _) = True
   |   
```

chwarr:
Bond running on native Apple Silicon isn't something that we need at Microsoft ourselves. We have no plan to invest in this at present. The architectures we target with full support are only x86 and x64 only.

PR #1200, opened 2024-03-07
minimal changes for ghc-8.10.7

Freddicus:
Possible steppingstone for #1199. Not sure about the accuracy of the changes, especially compiler/src/Language/Bond/Parser.hs#476

I was able to get this to compile on Apple Silicon when using [GHCup](https://www.haskell.org/ghcup/) and installing/setting the ghc version to 8.10.7 via `ghcup tui`.

Needed to `brew remove haskell-stack` and `cabal` and `ghc` first. Also, you will need `brew install llvm@13`.

### DIFF
--- a/compiler/Main.hs
+++ b/compiler/Main.hs
@@ -35,10 +35,10 @@ main = do
     options <- (if null args then withArgs ["--help=all"] else id) getOptions
     setJobs $ jobs options
     case options of
-        Cpp {..}    -> cppCodegen options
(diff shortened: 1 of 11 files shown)


chwarr:
Thanks for the PR. I mentioned this in #1199 as well, but we don't have any plans to support native Apple Silicon at present. Since we don't need this ourselves and we can't commit to maintaining and testing it going forward, I'd rather not merge these changes. Thanks for understanding.

PR #1210, opened 2024-09-30
Fixes for DoS vulnerabilities in C++

jandupej:
The following changes are included:
- Recursion depth is now also limited when deserializing JSON payloads in C++. Maximum depth can be set with `bond::SetDeserializeMaxDepth`.
- When deserializing containers in C++, arrays are now preallocated only after it is verified that there is enough data in the buffer to fill the array. If the expected serialized size cannot be calculated, a growing array is used to prevent excessive memory allocations using malicious payloads.
- `bond::CoreException` is now thrown when deseralizing an object whose declared size exceeds the remaining size of the payload in C++.
- `InputBuffer` now throws a `bond::StreamException` when trying to skip beyond the end of the stream in C++.
- Custom buffers (in C++) must now implement a `CanRead` method to check the remaining buffer size for reading.
- Custom protocols (in C++) must now implement a `CanReadArray` method to check the remaining buffer size for reading.
- Custom containers (in C++) must now supply `reset_list` and `list_insert` implementations to enable incrementally growing and filling the containers.
- Documentation was updated to reflect the new API.


vladimir-aubrecht (approved):

chwarr (changes requested):
I've sent you some comments over IM about the Linux build that I, alas, cannot share here.

jandupej:
> I've sent you some comments over IM about the Linux build that I, alas, cannot share here.

I fixed the gcc/clang build issues.

